PR #192, opened 2021-12-13
Add staging URL as a json key pair

brentru:
This pull request:
* Adds an optional `io_url` field to the filesystem's `secrets.json` file. Used for connecting to IO development servers. 
* Disables CI from building staging-specific firmware, all .uf2 files should be able to connect to a different server with a small modification to `secrets.json`.
*  Changes `mqttRC = WS._mqtt->connect(WS._username, WS._key);` to `mqttRC = WS._mqtt->connect();` as the `Adafruit_MQTT_Client` object was configured within a previous call to `setupMQTTClient`


brentru:
Tested with PyPortal and Adafruit Feather ESP32-S2

PR #219, opened 2022-02-22
Remove unused I2CDriver properties and memory fixes

brentru:
- Removes unused `DriverType_t` enum
- Makes `sensorAddress` a getter/setter within I2CDriver base class: `setI2CAddress` and `getI2CAddress`
- Properly [allocates](https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/219/files#diff-e578580947dd066c8681127545d0e027b1b2f5de808a3e02f2f34ca67dac65f6R43) and [deallocates](https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/219/files#diff-e578580947dd066c8681127545d0e027b1b2f5de808a3e02f2f34ca67dac65f6R52) memory for each sensor driver.
  - Destructors do not unnecessarily reset variables or set pointers within drivers to NULL.
- Sensor drivers pointers are properly [deleted and removed](https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/219/files#diff-3dd55e689e75bcb8859a65746742c85d8b634ee427ba635ece3bd5d91711eee6R319) from the vector, instead of simply being erased.


ladyada:
im a little paranoid about having public variables because people could set them by accident, unless you mean that?

brentru:
@ladyada I wanted to simplify the `src/drivers/WipperSnapper_I2C_Driver.h` interface, but I can also refactor it to utilize getters/setters to wrap these variables (`sensorAddress`, `isInitialized`) as private members.

ladyada:
i think its best to do so, in case something else comes up! its too easy for folks to accidentally set em especially as we get contributors!

brentru:
@ladyada  I've made a good # of modifications since the original PR comments and am requesting a re-review.

ladyada (approved):

PR #223, opened 2022-03-02
Add Adafruit ESP32 Feather V2

brentru:
Adds platform support for Adafruit ESP32 Feather V2
* Because a new release of https://github.com/espressif/arduino-esp32 was not cut yet, we're installing the arduino-esp32 bsp locally on the github runner, instead of from the BSP URL (https://github.com/brentru/ci-arduino). 
  * This step only executes on the ESP32 Feather V2, all other ESP32 platforms use the current Arduino ESP32 release
 
image


ladyada:
can we tweak the adafruit ci to support manual installation as an option?

brentru:
@ladyada  I could submit a separate PR over there! This repo uses the brentru adafruit ci fork though b/c of tinyusb build options.

ladyada:
i'd really prefer we just keep to one arduino-ci script, or it will cause us confusion later :D

brentru:
@ladyada  ok, no problem, lemmie work on combining them.

ladyada (approved):

PR #249, opened 2022-04-11
Fix IO_URL not set for LittleFS

brentru:
Fixes `io_url` never getting set for platforms using LittleFS.

Addresses
https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/236


ladyada (approved):

brentru:
Noting that this requires (https://github.com/adafruit/ci-arduino/pull/126) to pass the `feather_esp32_v2` check

Issue #150, opened 2021-09-13
Device disconnection and reboot after initial connection


brentru:
I loaded the beta 9 on my funhouse. I put it really close to our wifi router just to make sure I had the best signal possible. Windows 10 64 bit. It loaded fine, went through the lights, cyan to blue to amber came up in the IO , showed that it had been upgraded to the 9 beta, but disconnected immediately although the amber lights stayed on. Still shows offline and I notice that it seems like it reboots every 5 minutes or so. It's not an immediate reboot loop like the other, but it seems like it takes it a couple minutes and then you see it cycle through the lights again.

https://forums.adafruit.com/viewtopic.php?f=66&t=183113

Issue #143, opened 2021-08-30
Registration endlessly publish/encoded


brentru:
1st Report: MagTag
```
Registering Board...
registerBoard()
Encoding registration message...Encoded!
Publishing registration message...Published!
Publishing registration message...Published!
Publishing registration message...Published!
Publishing registration message...Published!
Publishing registration message...Published!
Publishing registration message...Published!
Publishing registration message...Published!
Publishing registration message...Published!
Publishing registration message...Published!
Publishing registration message...Published!
Publishing registration message...Published!
IO Throttle Error: "USER data rate limit reached, 1 seconds until throttle released"
Device is throttled for 1000ms and blocking command execution.
Device is un-throttled, resumed command execution
Publishing registration message...Published!
Publishing registration message...Published!
Publishing registration message...Published!
Publishing registration message...Published!
Publishing registration message...Published!
IO Throttle Error: "USER data rate limit reached, 1 seconds until throttle released"
Device is throttled for 1000ms and blocking command execution.



I've also removed all components from the device.```

PR #268, opened 2022-06-08
Fix registration UID bug

brentru:
This PR fixes a long-standing edge-case bug within the registration workflow. The bug causes a device to poll for a registration message response from the broker indefinitely. I previously thought this was a network reliability issue and was unable to reproduce user issues.

What happens:
The last three octets of the MAC address sent by the device, used as its unique identifier in IO, are converted into an integer to fit inside `int32_t _wippersnapper_description_v1_CreateDescriptionRequest::mac_addr` and pushed to the Broker.

The application code uses the original string of the last 3 octets of the MAC address as an identifier to build the topic strings required for WipperSnapper to communicate with Adafruit IO.

If a device's MAC NIC string starts with a zero (`0417489`), this string will be stripped of its zero within Wippersnapper::encodePubRegistrationReq()'s packing code, as the broker and protobuf both expect an integer (ie.: a 
formatted macAddr string (sUID) is char:0417489 and the atoi(sUID) is int:417489)

Since the broker is publishing a different topic:
`brentrubell/wprsnpr/io-wipper-adafruit-feather-esp32s3-4mbflash-2mbpsram417489/info/status/broker`

And the device is listening on a different topic than the broker:
`brentrubell/wprsnpr/io-wipper-feather-esp32s3-4mbflash-2mbpsram0417489/info/status/broker`

The device will never receive the registration message's transmission from the broker.

**This PR** modifies the device unique identifier topic to exactly match what is sent to the broker, ensuring the topics are the same.

**Tested on: ** ESP32-S3 with "bad MAC address" and ESP32-S3 with "good MAC address"

---

Resolves the following issues and requires forum follow-up on the following topics:
https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/150
https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/143
https://forums.adafruit.com/viewtopic.php?f=66&t=182565&start=30
https://forums.adafruit.com/viewtopic.php?f=66&t=190247

### DIFF
--- a/library.properties
+++ b/library.properties
@@ -1,5 +1,5 @@
 name=Adafruit WipperSnapper Beta
-version=1.0.0-beta.35
+version=1.0.0-beta.36
 author=Adafruit
 maintainer=Adafruit <adafruitio@adafruit.com>
 sentence=Arduino client for Adafruit.io WipperSnapper
--- a/src/Wippersnapper.cpp
+++ b/src/Wippersnapper.cpp
@@ -1074,15 +1074,22 @@ bool Wippersnapper::buildWSTopics() {
   }
   snprintf(WS.sUID, sizeof(WS.sUID), "%02d%02d%02d", WS._macAddr[0],
            WS._macAddr[1], WS._macAddr[2]);
-  // Set device UID
+  
+
+  // Set machine_name
   WS._boardId = BOARD_ID;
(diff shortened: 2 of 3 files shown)


lorennorman (approved):
Nasty problem, easy fix. Great work! 💯

Issue #269, opened 2022-06-08
Improve LED signaling during connection


lorennorman:
**Is your feature request related to a problem? Please describe.**
During initial connection currently, LEDs blink while working, then blink fast if there is an error. I believe this is a poor user experience because a user looking at a blinking LED will have to wonder "Is this blinking faster/slower than another hypothetical blinking LED?" They can only tell there is a problem if they know the behavior already and are watching the LEDs closely the entire time.

**Describe the solution you'd like**
Better, I believe, would be if the LEDs blink while there is activity, and go solid upon an error/failure. This way there is a very clear difference between the working/error states, and blinking has an implication of activity while solid might imply "hung" (or similar).

**Additional context**
This will go hand-in-hand with the upcoming Adafruit IO WipperSnapper firmware installer, get excited!

lorennorman:
This seems to do the opposite of what was requested:
- the lights were solid all the way through, never blinking
- it worked fine, progressed through the states (yellow, blue, purple, all solid) and pulsed green upon connection

PR #272, opened 2022-06-08
Improves LED signaling during connection 

brentru:
Addresses and resolves https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/269

Tested on: Adafruit Feather ESP32-S2

cc @lorennorman  - once built, you should be able to drag the UF2 asset onto a board like the FunHouse to ensure it works as you described.


lorennorman:
oops commented on the issue instead of here:

This seems to do the opposite of what was requested:
- the lights were solid all the way through, never blinking
- it worked fine, progressed through the states (yellow, blue, purple, all solid) and pulsed green upon connection

PR #277, opened 2022-06-17
Remove "beta" from library.properties "name" field

brentru:
Will open issue within https://github.com/arduino/library-registry/issues to update the name of the repository/library once a new release is cut.


brentru:
Thanks for the idea, will look into how to do this within the actions release workflow.

PR #281, opened 2022-07-08
Move C3 build to separate action and create combined file

makermelissa:

PR #282, opened 2022-07-11
Fix WipperSnapper GCC bugs for ESP32-2.0.4

brentru:
* Fixes bugs raised by GCC in ESP32 BSP 2.0.4
* Fixes regression for QSPI/QIO board(s) caused by ESP32 BSP 2.0.4 (https://github.com/espressif/arduino-esp32/pull/6962) 
* Removes `--no_warn` flag on ci-arduino runs

Will release once https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/281 is also tested and in-place


src/components/i2c/WipperSnapper_I2C.cpp
```diff
@@ -365,7 +365,8 @@ void WipperSnapper_Component_I2C::deinitI2CDevice(
   for (iter = drivers.begin(), end = drivers.end(); iter != end; ++iter) {
     if ((*iter)->getI2CAddress() == deviceAddr) {
       // Delete the object that iter points to
-      delete *iter;
+      // delete *iter;
+      *iter = nullptr;
```
ladyada, line 369: this does not delete? which is OK, as long as its not allocated by you

ladyada (approved):

src/components/i2c/WipperSnapper_I2C.cpp
brentru: 👍

brentru:
@makermelissa  Yes. https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/blob/main/.github/workflows/build-clang-doxy.yml#L38 was added in this PR as it temporarily fixes https://github.com/espressif/arduino-esp32/pull/6962 (I've integrated this patch only for the feather in the brentru repo).

The brentru/mqtt dep is also supposed to be there, but it's a much older dependency. I need to check over what changes I've made that warrant it being brentru/ instead of adafruit/.

makermelissa:
Thanks, just checking. :)

PR #283, opened 2022-07-12
Fix and merge binaries for ESP32 Web Tool

brentru:
ESP Web Tools do not work while uploading binaries at offsets after the Arduino-ESP32 BSP v2.0.4 release. @makermelissa  and I believe its due to web tools not patching the binary with flash size/freq, while esptool.py does.

This PR provides a patch that:
* Creates combined binaries for ESP32-4M, ESP32-8M with Esptool's `merge_bin` cmd to be uploaded at 0x0. WipperSnapper_Boards definitions for these boards will be updated with the new `structure` and offset.

Subsequent PR(s) will reduce the number of actions steps, and map the flash frequency and flash size to the appropriate esp32 board.

### DIFF
--- a/.github/workflows/build-clang-doxy.yml
+++ b/.github/workflows/build-clang-doxy.yml
@@ -396,7 +396,7 @@ jobs:
 
   clang_and_doxy:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


brentru:
Pull request tested on hardware: QT Py ESP32 Pico, Feather ESP32v2, Feather ESP32 Huzzah

PR #286, opened 2022-07-12
Add driver for SCD40 CO2, Temperature, and Humidity Sensor

marnibrewster:
This pull request adds support for using the SCD40 co2, temperature, and humidity sensor (https://www.adafruit.com/product/5187).
There is not an Adafruit library for this sensor, only a [Sensirion Arduino library](https://github.com/Sensirion/arduino-i2c-scd4x). This is already included in `library.properties`, but it does not implement `Arduino_Sensor`, so this PR likely needs some help. Also, I have never written C++ before, so I am likely botching things.
The output from doxygen that referenced my changes complained about defining variables, but I was unsure how to handle the way the Sensirion library makes one call to read (vs [the Adafruit python library that creates `scd.temperature`](https://github.com/adafruit/Adafruit_CircuitPython_SCD4X/blob/e4dba358a4e8a11c7e191c6ec8fdb143102ca91f/adafruit_scd4x.py#L122), for example).

With all of that being said, I was able to get this running locally:
Screen Shot 2022-07-12 at 3 57 46 PM
Screen Shot 2022-07-12 at 3 57 38 PM
Screen Shot 2022-07-11 at 8 01 09 PM

### DIFF
--- a/src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD40.h
+++ b/src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD40.h
@@ -8,7 +8,7 @@
  * please support Adafruit and open-source hardware by purchasing
  * products from Adafruit!
  *
- * Copyright (c) Brent Rubell 2021 for Adafruit Industries.
+ * Copyright (c) Marni Brewster 2022 for Adafruit Industries.
  *
  * MIT license, all text here must be included in any redistribution.
  *
--- a/src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD40.h
+++ b/src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD40.h
@@ -81,14 +81,9 @@ public:
     float temperature;
     float humidity;
     uint16_t error;
-    bool isDataReady = false;
-    error = _scd->getDataReadyFlag(isDataReady);
-    if (_tempSensorPeriod != 0 && error || !isDataReady) {
-      return false;
-    }
 
     error = _scd->readMeasurement(co2, temperature, humidity);
-    if (error || co2 == 0) {
+    if (_tempSensorPeriod != 0 && error || co2 == 0) {
       return false;
     }
 
@@ -111,14 +106,9 @@ public:
     float temperature;
     float humidity;
     uint16_t error;
-    bool isDataReady = false;
-    error = _scd->getDataReadyFlag(isDataReady);
-    if (_humidSensorPeriod != 0 && error || !isDataReady) {
-      return false;
-    }
 
     error = _scd->readMeasurement(co2, temperature, humidity);
-    if (error || co2 == 0) {
+    if (_humidSensorPeriod != 0 && error || co2 == 0) {
       return false;
     }
 
@@ -141,15 +131,12 @@ public:
     float temperature;
     float humidity;
     uint16_t error;
-    bool isDataReady = false;
-    error = _scd->getDataReadyFlag(isDataReady);
-    if (_CO2SensorPeriod != 0 && error || !isDataReady) {
-      return false;
-    }
+
     error = _scd->readMeasurement(co2, temperature, humidity);
-    if (error || co2 == 0) {
+    if (_CO2SensorPeriod != 0 && error || co2 == 0) {
       return false;
     }
+
     // TODO: This is a TEMPORARY HACK, we need to add CO2 type to
     // adafruit_sensor
     co2Event->data[0] = co2;
--- a/src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD40.h
+++ b/src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD40.h
@@ -83,7 +83,7 @@ public:
     uint16_t error;
 
     error = _scd->readMeasurement(co2, temperature, humidity);
-    if (_tempSensorPeriod != 0 && error || co2 == 0) {
+    if ((_tempSensorPeriod != 0 && error) || co2 == 0) {
       return false;
     }
 
@@ -108,7 +108,7 @@ public:
     uint16_t error;
 
     error = _scd->readMeasurement(co2, temperature, humidity);
-    if (_humidSensorPeriod != 0 && error || co2 == 0) {
+    if ((_humidSensorPeriod != 0 && error) || co2 == 0) {
       return false;
     }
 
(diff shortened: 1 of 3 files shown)


marnibrewster:
@brentru here is my (somewhat pathetic) attempt at the driver

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD40.h
```diff
@@ -0,0 +1,136 @@
...
+    }
+    return true;
+  }
+
+protected:
+  SensirionI2CScd4x *_scd; ///< SCD40 driver object
+  uint16_t _co2;
+  float _temperature;
+  float _humidity;
+  uint16_t _error;
```
marnibrewster, line 133: this is what doxygen was complaining about, but I was trying to avoid NACK from calling `readMeasurement` in each method.

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD40.h
```diff
@@ -0,0 +1,136 @@
+/*!
+ * @file WipperSnapper_I2C_Driver_SCD40.h
+ *
+ * Device driver for the SCD40 CO2, Temperature, and Humidity sensor.
+ * TEMPORARY HACK
```
brentru, line 5: Please add your name here :)

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD40.h
```diff
@@ -0,0 +1,136 @@
...
+      @brief    Gets the SCD40's current temperature.
+      @param    tempEvent
+                Pointer to an Adafruit_Sensor event.
+      @returns  True if the temperature was obtained successfully, False
+                otherwise.
+  */
+  /*******************************************************************************/
+  bool getEventAmbientTemperature(sensors_event_t *tempEvent) {
+    // check if sensor is enabled and data is available      
+    _error = _scd->readMeasurement(_co2, _temperature, _humidity);
```
brentru, line 74: Declare `uint16 error` and the rest of the measurements in each funcn instead of as a private member below.

Also, perform the error check and handling from the example, prior to `readMeasurement()`, in each `get...()` function:
https://github.com/Sensirion/arduino-i2c-scd4x/blob/master/examples/exampleUsage/exampleUsage.ino#L109

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD40.h
```diff
@@ -0,0 +1,136 @@
...
+  /*!
+      @brief    Gets the SCD40's current relative humidity reading.
+      @param    humidEvent
+                Pointer to an Adafruit_Sensor event.
+      @returns  True if the humidity was obtained successfully, False
+                otherwise.
+  */
+  /*******************************************************************************/
+  bool getEventRelativeHumidity(sensors_event_t *humidEvent) {
+    // check if sensor is enabled and data is available
```
brentru, line 92: Check if data is ready here: https://github.com/Sensirion/arduino-i2c-scd4x/blob/master/examples/exampleUsage/exampleUsage.ino#L109

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD40.h
```diff
@@ -0,0 +1,136 @@
...
+                Pointer to an Adafruit_Sensor event.
+      @returns  True if the humidity was obtained successfully, False
+                otherwise.
+  */
+  /*******************************************************************************/
+  bool getEventRelativeHumidity(sensors_event_t *humidEvent) {
+    // check if sensor is enabled and data is available   
+    if (_humidSensorPeriod != 0 && _error)
+      return false;
+
```
brentru, line 95: Then, perform `readMeasurement` to update the `humidity` value: https://github.com/Sensirion/arduino-i2c-scd4x/blob/master/examples/exampleUsage/exampleUsage.ino#L119

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD40.h
brentru: This may be resolved after you implement the changes above.

Clang is failing on our CI, please resolve using: https://learn.adafruit.com/the-well-automated-arduino-library/formatting-with-clang-format

brentru (changes requested):
Thank you so much for the PR, it looks great.

I have some feedback to tighten up the implementation and fix the CI errors you mentioned

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD40.h
marnibrewster: Thank you! This isn't in the exampleUsage.ino file in the arduino code 🤔

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD40.h
marnibrewster: Thanks for the feedback @brentru! I've pushed the requested changes.
The way you suggested is how I initially approached this, but changed it because of the errors I'm now seeing since making your change requests. Are these errors an issue? I'm assuming it's happening because [we are trying to `readMeasurement` again before the signal update interval has restarted](https://github.com/Sensirion/arduino-i2c-scd4x/blob/6b9835ea64435409ce169a29cde67cfbc2143711/src/SensirionI2CScd4x.h#L9)
This is happening before every successful reading / publish of co2, humidity, and temp:
Screen Shot 2022-07-15 at 6 20 14 PM
2)?

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD40.h
marnibrewster: Looks like I had 3.0 installed, and 3.1 wasn't available through arduino. Installing via zip worked!

brentru: Sensirion should update it, it's an external dependency and I do not have an ETA for this vendor. I'll file an issue on their repo and update this class when it's released properly.

Could you revert back to checking error like the example in Arduino library code, sorry, https://github.com/Sensirion/arduino-i2c-scd4x/blob/5c387a1417b1742f18eab26d6579e9ffb23d4334/examples/exampleUsage/exampleUsage.ino#L68

There's also one more compiler issue popping up in CI, please fix that and we should be able to PR this in.

marnibrewster: Thanks, @brentru - reverted the error logic and silenced the compiler whining 😄. It's working on my HUZZAH.

brentru (approved):
lgtm :shipit:

brentru:
@marnibrewster  Merging in and releasing, thanks for the work here!

Issue #284, opened 2022-07-12
Multiple boards push button showing opposite status 


kattni:
I set up a push button with the Feather ESP32-S3 to finish up filling in the Usage template. Once I finally had it wired up, it's showing the opposite pushed vs released results on the WipperSnapper dashboard, i.e. when I press the button, WS says "unpressed", and when I release it, WS says "pressed". 

I know we deal with this issue in CircuitPython by including `if not button.value:` in the code, but I'm not sure how it would be handled here.

I was using beta.38(1). I just now updated to beta.40 to make sure it hadn't been resolved - it has not.

Button is set up following the usage page instructions (on change, pull up), and wired to pins D5 and GND. (If you wish to verify I did it right, please see the screenshots here)

Please let me know if you need further information or testing.

kattni:
FunHouse: the built-in buttons are supported, but they don’t work at all with a pullup, and with a pulldown, the results are reversed. If I wire up an external button (which I did before realising the built-in buttons were available), and set it up with pullup on one of the STEMMA connector pins, results are reversed.

MagTag: Built-in buttons return backwards status. External button on A1 kept causing MagTag to disconnect (unrelated, but worth noting). Set up on D10, and the external button shows reverse status results.

PyPortal: External button shows reverse status.

brentru:
@kattni  We've pushed a fix for this to the website. Could you check the board(s) you were having trouble with?

kattni:
@brentru Sure thing. Will check now.

kattni:
Whee! Ok. @brentru Here's the situation.

Feather ESP32-S3 D5 external button - fixed.
MagTag built-in button - fixed.
PyPortal D3 external button - fixed.
FunHouse built-in button - no response from any of them. WS says "unpressed" and shows no change when button is pressed.

brentru:
![image](https://user-images.githubusercontent.com/322428/178817981-8f161893-7c42-498e-b8fc-7d8c8114faef.png)
@kattni  - I just tested my FunHouse and it works as expected.

Configuration:
![image](https://user-images.githubusercontent.com/322428/178818083-5d75168a-6553-49b9-a1c4-5d2857ec77c5.png)

brentru:
Perfect! thank you for filing this bug report

PR #287, opened 2022-07-13
Fix push buttons with internal pull-up resistor 

brentru:
Previously, this was handled on the application (invert the value whenever its shown). This functionality was not optimal and was removed in a recent revision of how the device interface buttons work. This PR allows the device to track if a digital GPIO pin has a pull-up resistor enabled, and invert the value prior to sending it to Adafruit IO.

Resolves https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/284


brentru:
@kattni  Could you test this on one of the boards you were having an issue with? UF2 files should be available from the GitHub runner (https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/actions/runs/2665282961) under "build files".

PR #316, opened 2022-09-08
Add PWM Output

brentru:
This pull request adds PWM output functionality to WipperSnapper.

**Feature**
* Adds a new component, `ws_pwm`, to WipperSnapper for PWM **output**
  * Used to drive new components on Adafruit IO such as the Pizeo Buzzer and Servo 
* Adds new helper functions `getChannel()` and `getInactiveChannel()` to `ws_ledc` component. 
  * Adds `resolution` to `attachPin` call.

**Stability and refactoring**
* Refactors `buildWSTopics()` to check for `malloc` failures immediately after allocating topic strings.
* Adds ESP32-x reset reason to serial output for enhanced debug output
* Adds new function, `generateDeviceUID()`, which generates device UID. This functionality was previously within `buildWSTopics()`.
* Moves the ledc servo driver to new `ledc/drivers` folder

Testing notes:
*  PWM should be tested w/ ESP32, ESP32-S2, ESP8266, SAMD
* Since there were changes to the ESP32-specific LEDC driver, both Servo and PWM should both be tested on ESP32


brentru:
Tested on ESP32, ESP8266

PR #319, opened 2022-09-12
Add si7021

tyeth:
This adds the Silicon Integrated Si7021 Temperature and Humidity driver.
https://www.adafruit.com/product/3251

Using adafruit-si7021 arduino library.

Device file is basically copy-pasta from SCD40 with changed names/methods and an added destructor like the guide suggests.

Matching component https://github.com/adafruit/Wippersnapper_Components/pull/73


brentru:
@tyeth  Please run doxygen across `src/components/i2c/drivers/WipperSnapper_I2C_Driver_SI7021.h` (CI Log: https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/actions/runs/3039463375/jobs/4895070376#step:6:8)

brentru:
@tyeth  You should be able to test this PR with Adafruit IO at this point.

The artifacts for this build, for all of the boards we test against, are on https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/actions/runs/3040113913 under "build-files".

tyeth:
Thanks, had to add one more commit. It was strange, I saw no error in console (using debug mode on in demo sketch), because I was just returning true after a failed init I guess, but it would stall at the ping/application runs and then watchdog would hard reset.
Fixed to pass in the i2c, which wasn't necessary by default in arduino, but is with wippersnapper (maybe when we pick the second i2c bus ;) )
Tested at 30second intervals fine. Survived a reboot etc.

brentru:
@tyeth Could you bump the version number here https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/blob/main/src/Wippersnapper.h#L70 and here https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/blob/main/library.properties#L2 to beta 49?

brentru (approved):

tyeth:
Blessed, done, thank you. Wondered where that was set :)

brentru:
@tyeth  Were you following https://learn.adafruit.com/how-to-add-a-new-component-to-adafruit-io-wippersnapper at all? Does it not contain info about where the version number is set? Any feedback to add?

tyeth:
I was following the guide, I'll admit skimming back and forth on the i2c driver page, but I dont remember version number, and having just rechecked it's not there. I assumed CI would do it, but I also noticed it was asking in the Devices page to update firmware (when testing) so probably add to that section of the guide: https://learn.adafruit.com/how-to-add-a-new-component-to-adafruit-io-wippersnapper/adding-an-i2c-component-driver#testing-your-new-sensor-3111513

I found it relatively easy to follow, just had an issue with clang and doxygen, but to be honest I had issues there because of windows and not using the commandline / taking shortcuts (using doxywizard probably was a bad idea) and it turned up loads of false positives so I had it misconfigured. I should have avoided windows and gone to linux/wsl but was on the slow laptop 🤷 so used vscode and :ahem: blagged it / manually linted

brentru:
Thanks for the feedback!

> but I dont remember version number, and having just rechecked it's not there. I 
I'll add this, thanks

> but I also noticed it was asking in the Devices page to update firmware (when testing) so probably add to that section of the guide: https://learn.adafruit.com/how-to-add-a-new-component-to-adafruit-io-wippersnapper/adding-an-i2c-component-driver#testing-your-new-sensor-3111513
and this too!

Issue #336, opened 2022-09-27
Improve PlatformIO Build Instructions in README.md


brentru:
The [PlatformIO build instructions in the README.md](https://github.com/adafruit/Adafruit_Wippersnapper_Arduino#platform-io-build-instructions) could use some improvement:

1. Test the PlatformIO build instructions on your local machine, comment below with feedback on how it can be improved. We'll discuss more below!
2. Update the build instructions in the `README.md` to reflect feedback from discussion
3. Create a pull request with the updated `README.md` file

tyeth:
Hiya, I've tried running and the feather32v2 compiles but wont link.
![image](https://user-images.githubusercontent.com/6692083/196054681-372ac9c3-507c-406c-ab81-4afdaaff421c.png)
The feather-s2 runs into a tinyUSB error maybe due to missing tiny_usb rename script (ref https://github.com/platformio/platform-espressif32/issues/809 ). 
![image](https://user-images.githubusercontent.com/6692083/196054976-2789b0e3-91b2-46b6-9831-271fb5feb9eb.png)
The feather8266 builds and wont link
![image](https://user-images.githubusercontent.com/6692083/196055024-4f1b40b7-68d0-48bd-94ef-8ac4ecadec42.png)
Running on windows 11 in a powershell / platformio terminal

brentru:
@tyeth  Here's what my current `.ini` looks like.

```
; PlatformIO Project Configuration File for Adafruit.IO WipperSnapper
;
; Please visit documentation for the other options and examples
; https://docs.platformio.org/page/projectconf.html

[platformio]
description = "Adafruit.IO WipperSnapper"
default_envs = featheresp32v2, featheresp32s2

; Common WipperSnapper Build Environment
; https://docs.platformio.org/en/latest/projectconf/section_env.html
[env]
framework = arduino
monitor_speed = 115200
lib_compat_mode = strict
lib_extra_dirs = ~/Documents/Arduino/libraries


; WipperSnapper local build environment for Adafruit Feather ESP32 v2
[env:featheresp32v2]
platform = https://github.com/platformio/platform-espressif32.git
;platform = https://github.com/Jason2866/platform-espressif32.git
monitor_filters = time, esp32_exception_decoder, colorize
board = adafruit_feather_esp32_v2
build_flags = -D ARDUINO_ADAFRUIT_FEATHER_ESP32_V2
upload_port = /dev/cu.wchusbserial54790143761
[...]

PR #357, opened 2022-10-31
Platformio readme and dependency changes

tyeth:
relates to #336


brentru:
@tyeth I'm hesitant to include this as it uses the DallasTemp library and we install that from scratch in the CI.

tyeth:
Yeah don't include it yet, I wanted to ask why it was done that way around,
i.e. in CI via a git clone and not listed in library.properties?
I'll remove the commit happily too, just needed it to test building in
arduino and platformIO in the Sandbox.

tyeth:
This seems no longer needed thanks to #446

PR #386, opened 2022-12-09
Add Sensirion SEN55 (maybe SEN54/SEN50 for free)

tyeth:
This adds the implementation for particulates <1 <2.5 <4 <10 in ug/m3, although <4 is yet to be included in wippersnapper, it is scaffolded out and there will be a corresponding patch when the respective sensor period implemented. 

The NOx and VOCs are supplied in the raw value, bytes 0 and 1 respectively. They correspond to an index, not a unit. 

There are also the usual humidity and temperature metrics.

It would be nice to test this with a SEN50 as there is a separate method in the driver to avoid returning metrics other than particle size. I'm hoping the user would deselect the metrics that their model doesn't have, and if not they'll just gracefully not update. Makes sense to test with a SEN54 if possible too. https://www.sparkfun.com/products/19325 is probably your best bet, I went with https://www.anglia-live.com/product/2080075001/SEN55-SDN-T/SENSIRION-AG (UK)

### DIFF
--- a/library.properties
+++ b/library.properties
@@ -7,4 +7,4 @@ paragraph=Arduino client for Adafruit.io WipperSnapper
 category=Communication
 url=https://github.com/adafruit/Adafruit_IO_Arduino
 architectures=*
-depends=Adafruit NeoPixel, Adafruit SPIFlash, ArduinoJson, Adafruit DotStar, Adafruit SleepyDog Library, Adafruit TinyUSB Library, Adafruit AHTX0, Adafruit BME280 Library, Adafruit DPS310, Adafruit SCD30, Sensirion I2C SCD4x, arduino-sht, Adafruit Si7021 Library, Adafruit MQTT Library, Adafruit MCP9808 Library, Adafruit MCP9600 Library, Adafruit TSL2591 Library, Adafruit_VL53L0X, Adafruit PM25 AQI Sensor, Adafruit VEML7700 Library, Adafruit LC709203F, Adafruit seesaw Library, Adafruit BME680 Library
+depends=Adafruit NeoPixel, Adafruit SPIFlash, ArduinoJson, Adafruit DotStar, Adafruit SleepyDog Library, Adafruit TinyUSB Library, Adafruit AHTX0, Adafruit BME280 Library, Adafruit DPS310, Adafruit SCD30, Sensirion I2C SCD4x, Sensirion I2C SEN5X, arduino-sht, Adafruit Si7021 Library, Adafruit MQTT Library, Adafruit MCP9808 Library, Adafruit MCP9600 Library, Adafruit TSL2591 Library, Adafruit_VL53L0X, Adafruit PM25 AQI Sensor, Adafruit VEML7700 Library, Adafruit LC709203F, Adafruit seesaw Library, Adafruit BME680 Library
--- a/src/components/i2c/WipperSnapper_I2C.cpp
+++ b/src/components/i2c/WipperSnapper_I2C.cpp
@@ -945,6 +945,49 @@ void WipperSnapper_Component_I2C::update() {
       (*iter)->setSensorGasResistancePeriodPrv(curTime);
     }
 
+    // NOx-index sensor
+    curTime = millis();
+    if ((*iter)->getSensorNOxIndexPeriod() != 0L &&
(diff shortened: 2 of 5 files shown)


tyeth:
There seems to be an intermittent need for a 1sec delay before first metric is read, should I (A) add a delay to begin after startMeasurement, (B) accept the read event failure and wait until next metric poll event (potentially 15minutes before first value is obtained by default for all metrics)

brentru:
> (A) add a delay to begin after startMeasurement
 
☝️ Might be worth trying, otherwise you'll get a bunch of 'ERROR: ...'s in the serial monitor.

src/wippersnapper/i2c/v1/i2c.pb.h
```diff
@@ -54,7 +54,9 @@ typedef enum _wippersnapper_i2c_v1_SensorType {
     wippersnapper_i2c_v1_SensorType_SENSOR_TYPE_ECO2 = 29,
     wippersnapper_i2c_v1_SensorType_SENSOR_TYPE_UNITLESS_PERCENT = 30,
     wippersnapper_i2c_v1_SensorType_SENSOR_TYPE_AMBIENT_TEMPERATURE_FAHRENHEIT = 31,
-    wippersnapper_i2c_v1_SensorType_SENSOR_TYPE_OBJECT_TEMPERATURE_FAHRENHEIT = 32
+    wippersnapper_i2c_v1_SensorType_SENSOR_TYPE_OBJECT_TEMPERATURE_FAHRENHEIT = 32,
```
brentru, line 57: @tyeth  Please drop this file from the commit! 

Once we merge https://github.com/adafruit/Wippersnapper_Protobuf/pull/124, a new PR will be generated here containing the updated protobuf messages. I'll merge that into `adafruit/main` and you can then merge it into this PR + use it.

tyeth: @brentru done I think, nothing scarier than rewriting history 😨

tyeth:
@brentru  Removed the protobuf change as requested

brentru:
@tyeth The new protobufs were added in https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/395. You will want to merge `main` into this branch and compile. Then, fix up the errors.

brentru:
@tyeth  This should be good to go! Check that you see the SEN5X component on the picker when you enable dev. mode first.

tyeth:
Thanks @brentru, appears to be just clang now after rebasing. I've stupidly left the potato at home, and am away for new year, but will be cleaned up then tested early next week
Potato viewing as your reward: 
[https://youtube.com/shorts/TOqu0H7t5EM](https://youtube.com/shorts/TOqu0H7t5EM?feature=share)
and under the cardboard: https://www.youtube.com/watch?v=bVxNqmdw-04
[Meant for my brother so zero video forethought]

tyeth:
It's running, but nox is still just 1.0, so removed and re-added entire component without farenheit, will leave it running for a while.

tyeth:
Turns out that the index baseline for NOx is 1 not 100, so all is well with the world apart from when I use too many metrics.
I'll update the comments in driver to reflex that, and in wippersnapper to reflect expected baseline/default for nox, but is there anything backend to update regarding the NOx index default value of 1 instead of 100?
@brentru do I need to prod loren?

brentru:
@tyeth  As this all looks correct and passing, doing a review over it.

brentru (approved):
This looks good and is passing, will release along with beta 58

brentru:
@tyeth  I'm getting one of the SEN55 modules in, may give this an update when I test.

tyeth:
Interesting! Curious what you think you might change or add? 
The NOx reading really has been useless for me, but I did get it to at least raise when trapped under a saucepan and a lighter used to create a yellow flame like the youtube video from Sensirion (inefficient burning is shown by the yellow of a flame indicating soot and non complete burning). I think it would show great utility in the correct circumstance, but the SEN54 is most useful for people (Ikea have chosen that one with an LCD for their not-yet-matter-compatible new device landing in April)
 Also if you do anything additional like a couple of outputs and the neopixel you'll probably find wippersnapper needs a larger metrics count, and we could help people with a note in the docs / UI to mention the current limit

PR #451, opened 2023-07-17
Add SGP30 and tvoc/eco2

tyeth:
Implements the glue code for Sensirion SGP30 - adds a requirement on Adafruit SGP30 library.
Implements the events polling for the two additional data types, eCO2 and TVOC.

This closes https://github.com/adafruit/Wippersnapper_Components/issues/102

@brentru I've tested this on the feather S3 4mb/2mb. Data flows fine.
 I was getting a i2c packet decode error on reboots with an SI7021 connected and added to Device page in addition to the SGP30, but when I removed that (Si7021 component) it stopped the decode error. Similar to last year, the decode error prevents the initialising from working and I had to remove components and re-add (initially in testing I was removing then re-adding the SGP30), and in the final test I just removed the Si7021 and left the SGP30, and it rebooted after the deinit message and then data for the SGP30 started flowing (no serial after the reconnect) without the need to remove and re-add it.
I thought last year it was related to the number of metrics, but maybe also when we added new data types. Might need a quick look on Wednesday when you get back.


brentru:
@tyeth  Could you provide a screenshot of the Adafruit.io device page with the SGP30 (only) connected to the Feather ESP32-S3?

tyeth:
> @tyeth Could you provide a screenshot of the Adafruit.io device page with the SGP30 (only) connected to the Feather ESP32-S3?

I'm in the greenhouse at the sec, but I have a reverse tft S2 if that's good enough? I'll go upstairs within the hour and get the S3 non-tft feather.
So I used the bin file from the build-files from the action on this page, and applied to the s2-rev-tft, it works, no display but normal functionality.
Mostly zeros except when I breathed on it continuously for about 30seconds. The one upstairs in the house has been more excitable value-wise. 
image

brentru:
> Mostly zeros except when I breathed on it continuously for about 30seconds. The one upstairs in the house has been more excitable value-wise.

Sounds good, thank you for testing. You can also use a soldering iron and burn solder next to the sensor for next time to avoid blowing on the sensor.

PR #453, opened 2023-07-21
Add I2C component HTS221 humidity sensor driver

whatisbyandby:
This PR adds the I2C component driver for the HTS221 humidity sensor.

Product URL: [https://www.adafruit.com/product/4535](https://www.adafruit.com/product/4535)

I decided to use the 1hz data rate, but would be open to feedback if there is a better data rate to use.

Relates To:
Issue: [https://github.com/adafruit/Wippersnapper_Components/issues/130](https://github.com/adafruit/Wippersnapper_Components/issues/130
Component definition: [https://github.com/adafruit/Wippersnapper_Components/pull/131](https://github.com/adafruit/Wippersnapper_Components/pull/131)


whatisbyandby:
I tried to understand what was going on with the two checks that failed, but I'm not seeing anything I can make sense of. 

For the Feather ESP32 I don't see any output in the logs at all as to what failed. Maybe I'm not looking in the right place.

The Feather ESP8266 build is complaining about something on line 22 of the src/components/i2c/WipperSnapper_I2C.h which is the include statement for the WipperSnapper_I2C_Driver.h. I don't see the actual error that is coming from that line, but again maybe I'm not looking in the right place. 

Any insight as to what I'm missing would be greatly appreciated. 

Cheers!

brentru:
LGTM!

brentru (approved):

PR #455, opened 2023-07-21
Add QT Py ESP32-S3 with 2MB PSRAM

brentru:
Related PR: https://github.com/adafruit/Wippersnapper_Boards/pull/112


brentru:
Tested OK against production

image

PR #464, opened 2023-08-11
Add LPS35HW code, also supporting LPS33HW

tyeth:
Add support for Water Resistant Pressure sensor family LPS3xHW

This specifically adds the LPS35HW, but the code is identical so I will add the detection for the LPS33HW next week when I have one to test.


tyeth:
Two LPS35HW tested on same bus with QTPY ESP32S2.
image
image

brentru (approved):

brentru:
@tyeth  Looks good to merge, I like how the driver is a generic LPSx object. I'll bring this in.

PR #468, opened 2023-08-16
Add MPL115A2

tyeth:
Adds https://www.adafruit.com/product/992 with Adafruit MPL115A2 library. Tested with one on Huzzah32v2, also on same bus as TC74A0

Later retested from CI image, with BMP388 on same bus, QTPYS2. 
image


brentru (approved):
lgtm!

PR #474, opened 2023-08-16
Fix ESP8266 Compiler Warnings

brentru:
This PR seeks to reduce ESP8266 builds failing by reducing the amount of compiler warnings produced such as:

- Unused parameter
- Virtual destructor
- Comparison of integer expressions of different signedness

### DIFF
--- a/src/components/i2c/drivers/WipperSnapper_I2C_Driver.h
+++ b/src/components/i2c/drivers/WipperSnapper_I2C_Driver.h
@@ -46,7 +46,7 @@ public:
       @brief    Destructor for an I2C sensor.
   */
   /*******************************************************************************/
-  ~WipperSnapper_I2C_Driver() { _sensorAddress = 0; }
+  virtual ~WipperSnapper_I2C_Driver() {}
 
   /*******************************************************************************/
   /*!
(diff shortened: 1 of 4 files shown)


brentru:
@tyeth Could you please review this PR? If it is OK, you can merge it in. I did touch some core API things (mostly marking unused parameters to avoid warnings) so it may require a functional test.

For any PRs failing on ESP8266: Merge this PR after and see if they work

src/components/i2c/drivers/WipperSnapper_I2C_Driver.h
```diff
@@ -206,7 +206,11 @@ class WipperSnapper_I2C_Driver {
                 otherwise.
   */
   /*******************************************************************************/
-  virtual bool getEventCO2(sensors_event_t *co2Event) { return false; }
+  virtual bool getEventCO2(sensors_event_t *co2Event) {
+    (void)
+        co2Event; // Parameter is intentionally unused in this virtual function.
```
tyeth, line 211: Just curious why this is split over two lines and the next (L:256) isn't. Presumably related to comment splitting / length

src/components/ds18x20/ws_ds18x20.cpp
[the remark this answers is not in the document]
tyeth: Also a long vs uint32 warning, about sensor period.

tyeth (changes requested):
@brentru Looks good, makes sense removing the i2c bus number too for now.
Suggested changes (can't seem to suggest or comment in non-touched files + areas)

https://github.com/tyeth/Adafruit_Wippersnapper_Arduino/commit/9532d0b8eb25ebd3d37f7e0cf4c0272566c52bf1#

src/components/i2c/drivers/WipperSnapper_I2C_Driver.h
brentru: I believe it's an artifact from running `clang-format`

PR #473, opened 2023-08-16
Add DHT20 as alias of AHTx0

tyeth:
Adds the Pin version of the AHT20, aka the DHT20.
https://www.adafruit.com/product/5183

This depends on https://github.com/adafruit/Wippersnapper_Components/pull/145

PR #475, opened 2023-08-25
Add AM2301B and AM2315C as AHT20

tyeth:
Adds both as aliases of AHT20, using the Adafruit AHTx0 driver
https://www.adafruit.com/product/5181
https://www.adafruit.com/product/5182


tyeth:
Tested using buildfiles uf2 on qtpy s2
image

I'll rebase this after #473

PR #476, opened 2023-08-28
Add Sonoff TH10/16

tyeth:
This adds the initial provisional ALPHA quality release for the Sonoff TH10/16 Relays, using the ESP8266.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -68,6 +68,7 @@ monitor_filters = esp32_exception_decoder, time
 [common:esp8266]
 platform = espressif8266
 lib_ignore = WiFiNINA, Adafruit TinyUSB Library
+board_build.filesystem = littlefs
 
 ; ESP32-x Boards ;
 
@@ -183,7 +184,7 @@ build_flags = -DARDUINO_ESP8266_ADAFRUIT_HUZZAH
 
 ; Sonoff TH16 ESP8266
 ; https://wiki.iteadstudio.com/images/3/39/Sonoff_TH10A%2816A%29_schmatic.pdf
-[env:sonoff_th16]
+[env:sonoff_th]
 extends=common:esp8266
 board = sonoff_th
 build_flags = -DESP8266_SONOFF_TH
--- a/platformio.ini
+++ b/platformio.ini
@@ -182,6 +182,7 @@ board = huzzah
 build_flags = -DARDUINO_ESP8266_ADAFRUIT_HUZZAH
 
 ; Sonoff TH16 ESP8266
+; https://wiki.iteadstudio.com/images/3/39/Sonoff_TH10A%2816A%29_schmatic.pdf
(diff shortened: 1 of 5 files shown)


tyeth:
Will require the board adding to the CI system, both in this repo (doxy yaml), and ci-arduino 
https://github.com/adafruit/ci-arduino/blob/master/build_platform.py#L54-L179
(Currently the ci-arduino repo used is not the adafruit copy, but the brentru fork)

platformio.ini
```diff
@@ -68,6 +68,7 @@ monitor_filters = esp32_exception_decoder, time
 [common:esp8266]
 platform = espressif8266
 lib_ignore = WiFiNINA, Adafruit TinyUSB Library
+board_build.filesystem = littlefs
```
brentru, line 71: I am curious - why is this flag required for the sonoff_th but not for the huzzah ESP8266?

src/Wippersnapper_Boards.h
```diff
@@ -132,6 +132,11 @@
 #define USE_LITTLEFS
 #define USE_STATUS_LED
 #define STATUS_LED_PIN 0
+#elif defined(ARDUINO_ESP8266_SONOFF_TH)
+#define BOARD_ID "sonoff-th"
+#define USE_LITTLEFS
+#define USE_STATUS_LED
+#define STATUS_LED_PIN 13 // pull down
```
brentru, line 139: Is this pin pulled down by default?

src/components/digitalIO/Wippersnapper_DigitalGPIO.cpp
```diff
@@ -82,7 +82,7 @@ void Wippersnapper_DigitalGPIO::initDigitalPin(
 #endif
 
 // Initialize LOW
-#if defined(ARDUINO_ESP8266_ADAFRUIT_HUZZAH)
+#if defined(ARDUINO_ESP8266_ADAFRUIT_HUZZAH) || defined(ESP8266_SONOFF_TH)
```
brentru, line 85: This applies to every ifdef like this, would we want to detect the ESP8266 arch instead of individual boards? Both boards we're using here (and elsewhere) seem to take the same pinout/pd resistor

brentru (changes requested):
Looks OK, some questions and changes. Have you gotten this running yet?

platformio.ini
tyeth: I think maybe it is required for both. I couldn't push the filesystem image for the ESP8266 huzzah without this, I kept forgetting as it was part of my stash with secrets data file. Maybe I've got something configured wrong, but I think it was defaulting to spiffs rather than littlefs. I did the initial sonoff work just before the huzzah arrived later in the day and I was then able to get the secrets across and see debug with both sonoff + huzzah.
 I'll try to check on a fresh machine at some point soon.

PR #479, opened 2023-08-31
Fix driver for BME680 + BME688 using alt address

tyeth:
During the mega-testing after resolving decode issues, I've been struggling to get a BME688 online. It was using the alt address, and about a dozen things on the bus.
I moved it to a solo qtpyS2 and still an issue. Spotted the driver isn't passing the address.
I missed this when adding the BME688 as an alias of the BME680. I think I must have failed to connect both at the same time as in my original testing method. Mental note to still connect+screenshot both even if screenshots need one on default address.
image


tyeth:
@brentru minor issue discovered while sensoring up the banana plant. Default address is 0x77

PR #480, opened 2023-09-01
Add BMP390 as alias of BMP388

tyeth:
Adds BMP390
https://www.adafruit.com/product/4816


tyeth:
Tested on default address, also tested appears on alt address, and code uses address for begin.
image

PR #482, opened 2023-09-01
Add LPS22HB + LPS25HB Pressure sensors

tyeth:
Adds 
https://www.adafruit.com/product/4633
https://www.adafruit.com/product/4530


tyeth:
Tested with both on bus (one must use alt address) 
image

@brentru ready for review

PR #499, opened 2023-10-03
Add LTR390 sensor with switching mode example

tyeth:
@brentru was thinking something like this for the switching light modes between Ambient Light (ALS) and UV (UVS) modes, inside  each event type.


src/components/i2c/drivers/WipperSnapper_I2C_Driver_LTR390.h
```diff
@@ -0,0 +1,120 @@
...
+    if (!_ltr390->newDataAvailable())
+      return false;
+
+    lightEvent->light = _ltr390->readALS();
+    return true;
+  }
+
+  /*******************************************************************************/
+  /*!
+      @brief    Reads the LTR390's UV value into an event.
```
brentru, line 96: Do we have a unit for UV-value or is it unitless?

src/components/i2c/drivers/WipperSnapper_I2C_Driver_LTR390.h
```diff
@@ -0,0 +1,120 @@
...
+  bool begin() {
+    _ltr390 = new Adafruit_LTR390();
+    // Attempt to initialize LTR390
+    if (!_ltr390->begin(_i2c))
+      return false;
+
+    // Configure LTR390 sensor
+    // Note: This driver uses the default configuration from
+    // https://github.com/adafruit/Adafruit_LTR390/blob/master/examples/ltr390_test/ltr390_test.ino
+    _ltr390->setMode(LTR390_MODE_ALS);
```
brentru, line 65: Which mode, UV or ALS, would a person use more often? We should make that the "default" mode here.

src/components/i2c/drivers/WipperSnapper_I2C_Driver_LTR390.h
```diff
@@ -0,0 +1,120 @@
...
+      @param    lightEvent
+                Light sensor reading, in lux.
+      @returns  True if the sensor event was obtained successfully, False
+                otherwise.
+  */
+  /*******************************************************************************/
+  bool getEventLight(sensors_event_t *lightEvent) {
+    if (_ltr390->getMode() != LTR390_MODE_ALS) {
+      _ltr390->setMode(LTR390_MODE_ALS);
+      delay(100);
```
brentru, line 84: Do we need this delay?

src/components/i2c/drivers/WipperSnapper_I2C_Driver_LTR390.h
```diff
@@ -0,0 +1,120 @@
...
+      @param    rawEvent
+                Pointer to an adafruit sensor event.
+      @returns  True if the sensor event was obtained successfully, False
+                otherwise.
+  */
+  /*******************************************************************************/
+  bool getEventRaw(sensors_event_t *rawEvent) {
+    if (_ltr390->getMode() != LTR390_MODE_UVS) {
+      _ltr390->setMode(LTR390_MODE_UVS);
+      delay(100);
```
brentru, line 106: Do we need this delay?

brentru (changes requested):

src/components/i2c/drivers/WipperSnapper_I2C_Driver_LTR390.h
tyeth: I think the delay needs to be >105, maybe 110 for safety? This is assuming the 18bit resolution giving 100ms integration time
image
And the switch from standby (which makes sense to implement if you agree) to reading takes 5ms before starting reading (or checking the measurement scheduler)
image
https://optoelectronics.liteon.com/upload/download/DS86-2015-0004/LTR-390UV_Final_%20DS_V1%201.pdf

src/components/i2c/drivers/WipperSnapper_I2C_Driver_LTR390.h
tyeth: Maybe it's a UV sensor by title so we pick that 🤷 no idea really, but it also claims lux output for ambient which seemed useful and relevant to more projects maybe

src/components/i2c/drivers/WipperSnapper_I2C_Driver_LTR390.h
tyeth: seems to be a unitless count, which can be converted to an index too
image

brentru: okay, lets leave it then

src/components/i2c/drivers/WipperSnapper_I2C_Driver_LTR390.h
brentru: Let's set it to `LTR390_MODE_UVS` within the init. as that's what happens first in the example sketch and may be the largest use case.

src/components/i2c/drivers/WipperSnapper_I2C_Driver_LTR390.h
brentru: I don't want to implement a blocking loop since we're technically already polling this in a loop. We could increase the delay to 110ms (5ms for the conversion + safety)

src/components/i2c/drivers/WipperSnapper_I2C_Driver_LTR390.h
brentru: See above ^

brentru: Change to 110ms

tyeth:
@brentru that should be okay now. I've still not enabled standby mode between reads. Currently it feels like power usage is less important than having readings readily sampled, but maybe once we cover deep-sleep on esp32 we can revisit such thoughts

brentru (approved):
Changes look good, thanks for working this through with me

Issue #495, opened 2023-09-27
"Dimmable LED" Component on HUZZAH ESP8266 is inverted


brentru:
**Describe the bug**
The dimmable LED component's functionality is inverted when used with the Feather Huzzah ESP8266:

![IO_-_Device__Feather_Huzzah_ESP8266_🔊](https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/assets/322428/ee006f68-29c0-4fb8-bd29-0282c7ca0a74)

This may be caused due to the fact that the Adafruit Feather Huzzah ESP8266 has inverted wiring on the RED LED.

***Arduino board***
INSERT ARDUINO BOARD NAME/TYPE HERE

**To Reproduce**
1. Create a new Dimmable LED Component
2. Drag the slider to 100%
3. Drag the slider back to 0%

**Which components are connected to your device**
Dimmable LED (D0)

brentru:
@tyeth  Wanna take a look at this issue?

tyeth:
This sounds like something for the analog output basics guide, so we show a normal one and mention / explain how that would play out if you had it wired oppositely (i.e. the LED would be on with the Low part of the duty cycle), like you find on some boards.

PR #511, opened 2023-11-24
Fix Feather HUZZAH ESP8266 Inverted LED PWM 

brentru:
Resolves https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/495 by inverting the analog write calculation for the Adafruit Feather ESP8266 HUZZAH

### DIFF
--- a/src/components/pwm/ws_pwm.cpp
+++ b/src/components/pwm/ws_pwm.cpp
@@ -88,6 +88,10 @@ void ws_pwm::detach(uint8_t pin) {
 void ws_pwm::writeDutyCycle(uint8_t pin, int dutyCycle) {
 #if defined(ARDUINO_ARCH_ESP32)
   _ledcMgr->analogWrite(pin, dutyCycle);
+#elif defined(ARDUINO_ESP8266_ADAFRUIT_HUZZAH)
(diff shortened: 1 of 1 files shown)


src/components/pwm/ws_pwm.cpp
```diff
@@ -88,6 +88,10 @@ void ws_pwm::detach(uint8_t pin) {
 void ws_pwm::writeDutyCycle(uint8_t pin, int dutyCycle) {
 #if defined(ARDUINO_ARCH_ESP32)
   _ledcMgr->analogWrite(pin, dutyCycle);
+#elif defined(ARDUINO_ESP8266_ADAFRUIT_HUZZAH)
+  // Adafruit Feather ESP8266's analogWrite() is inverted because LED pin is
+  // reverse-wired
```
tyeth, line 93: Think this is misleading, aren't all pwm outputs inverted compared to esp32?

tyeth (approved):
Looks good, could benefit from an update to comment for clarity

src/components/pwm/ws_pwm.cpp
brentru: On every board except the esp8266 huzzah, no

Issue #463, opened 2023-08-07
Feature Req: Would be nice if Sensor always sent a reading when first added


tyeth:
Currently if a sensor is added it defaults each metric to 15mins, and doesn't send a value to adafruit IO until that polling period is expired.

It would be nice if new sensor metrics/feeds always sent data upon initialisation (although this also has rate consequences)

brentru:
@tyeth  Do you want to tackle this one?

tyeth:
Just occurred to me while thinking about this one, that it's possible some users are rate limiting themselves using the period as a starting offset. This is unlikely, but possible.

PR #512, opened 2023-11-24
Initial tweaking of previous poll period

tyeth:
Will close https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/463

### DIFF
--- a/src/components/digitalIO/Wippersnapper_DigitalGPIO.cpp
+++ b/src/components/digitalIO/Wippersnapper_DigitalGPIO.cpp
@@ -116,12 +116,15 @@ void Wippersnapper_DigitalGPIO::initDigitalPin(
              pinName, periodMs);
     WS._ui_helper->add_text_to_terminal(buffer);
 #endif
+    // get current time
+    long curtime = millis() - 1;
 
     // attempt to allocate a pinName within _digital_input_pins[]
     for (int i = 0; i < _totalDigitalInputPins; i++) {
       if (_digital_input_pins[i].period == -1L) {
         _digital_input_pins[i].pinName = pinName;
         _digital_input_pins[i].period = periodMs;
+        _digital_input_pins[i].prvPeriod = curtime - periodMs;
         break;
       }
     }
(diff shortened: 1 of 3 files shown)


tyeth:
@brentru tested on S2 TFT with LTR390, period set at 5mins, immediate data sent.

src/components/digitalIO/Wippersnapper_DigitalGPIO.cpp
```diff
@@ -116,12 +116,15 @@ void Wippersnapper_DigitalGPIO::initDigitalPin(
              pinName, periodMs);
     WS._ui_helper->add_text_to_terminal(buffer);
 #endif
+    // get current time
+    long curtime = millis() - 1;
```
brentru, line 120: `curTime`

src/components/i2c/drivers/WipperSnapper_I2C_Driver.h
```diff
@@ -1314,97 +1314,123 @@ class WipperSnapper_I2C_Driver {
     setSensorProximityPeriod(period);
   }
 
+  long SENSOR_STARTING_PERIOD_PRV =
```
brentru, line 1317: 1) Where did this formula come from?
2) Caps and case - `sensorStartingPeriodPrv`

src/components/digitalIO/Wippersnapper_DigitalGPIO.cpp
brentru, line 120: Do we want this to be a `long`, millis() returns an `unsigned long`

brentru (changes requested):
This addresses a big issue, thank you! Requesting clarification and changes

src/components/i2c/drivers/WipperSnapper_I2C_Driver.h
tyeth: Thinking yes swap to a macro. The formula was 24hrs in milliseconds, to account for the unknown period (max of 24hrs in wippersnapper)

Issue #518, opened 2023-12-07
Ensure WipperSnapper is build with latest stable release of nanopb


brentru:
Currently, nanopb version 0.4.6 is statically linked into this library's [src/nanopb](https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/tree/main/src/nanopb) directory. Ideally, we'd replace these files whenever a new release is created. However, automating this is a much better bet since we are already 2 sub-versions behind.

To ensure this library is compiled with the latest stable release version of nanopb, we'll need some tooling implemented.

PR #519, opened 2023-12-07
Add nanopb installation to CI

brentru:
Resolves https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/518


tyeth (approved):
Think that all makes sense. Initially a bit curious about their releases etc, but I see now. Only question I had was why move instead of copy that one file, line length?

PR #528, opened 2024-01-08
Initial VL6180X implementation

tyeth:
This add the STM VL6180X, returning range and light level / lux at Gain level 5.

Plan is to revisit this component and others with customisable parameters once we have I2C custom sensor parameters supported in WipperSnapper components


src/Wippersnapper.h
```diff
@@ -40,6 +40,24 @@
 #include "Wippersnapper_Boards.h"
 #include "components/statusLED/Wippersnapper_StatusLED.h"
 
+#define WS_DEBUG          ///< Define to enable debugging to serial terminal
```
brentru, line 43: @tyeth  why are these being added back?

src/components/i2c/drivers/WipperSnapper_I2C_Driver_VL6180X.h
```diff
@@ -0,0 +1,133 @@
...
+      } else if (status == VL6180X_ERROR_RANGEOFLOW) {
+        WS_DEBUG_PRINTLN("VL6180X: Range reading overflow");
+      }
+      return false;
+    }
+
+    if (range <= 0 || range > 150) {
+      return false;
+    }
+    proximityEvent->data[0] = range;
```
brentru, line 105: What units does this return - do we have a type for it?

brentru (changes requested):
@tyeth  Two things..

library.properties
```diff
@@ -5,6 +5,6 @@ maintainer=Adafruit <adafruitio@adafruit.com>
 sentence=Arduino application for Adafruit.io WipperSnapper
 paragraph=Arduino application for Adafruit.io WipperSnapper
 category=Communication
-url=https://github.com/adafruit/Adafruit_IO_Arduino
```
tyeth, line 8: @brentru This needs updating so I slipped it in with this PR

src/Wippersnapper.h
tyeth: I've moved these higher, above line 64, so that WS_DEBUG_PRINT is available to the i2c drivers for better debugging.

library.properties
brentru: Yeah I caught that one, thanks for adding it.

src/Wippersnapper.h
brentru: Okay!

src/components/i2c/drivers/WipperSnapper_I2C_Driver_VL6180X.h
tyeth: I believe millimetres for all proximity sensors, but selecting the proximity sensor / subcomponent type we have a unitless display in IO, labelled Value with the title SENSORNAME: Proximity Sensor. 
Sadly the proximity values don't agree between sensors so I'm almost thinking leave it as unitless "Value"

brentru:
@tyeth  Approved, please resolve the merge conflicts then OK to merge in

brentru (approved):

PR #537, opened 2024-01-15
Always return proximity result, or NaN if we know it's bad

tyeth:
~~Running CI so I can test easily~~

This addresses the issue raised in the meeting about NaN's being returned rather than artificially restricting values returned by sensors.


tyeth:
@brentru I forgot about this one, I should probably get it into v76, would you mind a review please

brentru (approved):

PR #538, opened 2024-01-16
Add titano

tyeth:
Drafting this to check build before review

PR #539, opened 2024-01-17
Fix prvPeriod in AnalogIO - starting value 1628136801 on SAMD

tyeth:
Separating this out from #538

### DIFF
--- a/src/components/analogIO/Wippersnapper_AnalogIO.cpp
+++ b/src/components/analogIO/Wippersnapper_AnalogIO.cpp
@@ -145,6 +145,7 @@ void Wippersnapper_AnalogIO::initAnalogInputPin(
     if (_analog_input_pins[i].enabled == false) {
       _analog_input_pins[i].pinName = pin;
       _analog_input_pins[i].period = periodMs;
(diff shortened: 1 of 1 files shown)


src/components/analogIO/Wippersnapper_AnalogIO.cpp
```diff
@@ -145,6 +145,7 @@ void Wippersnapper_AnalogIO::initAnalogInputPin(
     if (_analog_input_pins[i].enabled == false) {
       _analog_input_pins[i].pinName = pin;
       _analog_input_pins[i].period = periodMs;
+      _analog_input_pins[i].prvPeriod = 0L;
```
brentru, line 148: @tyeth  Is there anywhere else in the `analogio` or `digitalio` class this bug is present?

tyeth:
Looked like that was the only initialisation route, and I spotted the
difference in usage of prvPeriod (which seemed best solution) between
analogIO and digitalIO (which looks fine to me)

brentru:
We are going to refactor these two classes during the api-v2 transition so I wouldn't worry too much about that. Thanks for taking a look, LGTM.

brentru (approved):

PR #540, opened 2024-01-17
Add PyPortal Titano plus fallback Fs reinit and repeated serial if Halted

tyeth:


src/Wippersnapper.cpp
```diff
@@ -2306,9 +2306,14 @@ void Wippersnapper::errorWriteHang(String error) {
   WS_DEBUG_PRINTLN(error);
 #ifdef USE_TINYUSB
   _fileSystem->writeToBootOut(error.c_str());
+  delay(500);
```
brentru, line 2309: Is this wait/attach/wait pattern required here?

src/provisioning/littlefs/WipperSnapper_LittleFS.cpp
```diff
@@ -28,8 +28,8 @@
 WipperSnapper_LittleFS::WipperSnapper_LittleFS() {
   // Attempt to initialize filesystem
   if (!LittleFS.begin()) {
-    WS_DEBUG_PRINTLN("ERROR: Failure initializing LittleFS!");
     setStatusLEDColor(RED);
+    fsHalt("ERROR: Failure initializing LittleFS!");
```
brentru, line 32: Remove the `while(1)...` beneath L32 as the program will never reach it after `fsHalt()` is called

src/provisioning/littlefs/WipperSnapper_LittleFS.cpp
```diff
@@ -99,11 +94,14 @@ void WipperSnapper_LittleFS::parseSecrets() {
   LittleFS.end();
 }
 
-void WipperSnapper_LittleFS::fsHalt() {
+void WipperSnapper_LittleFS::fsHalt(String msg) {
```
brentru, line 97: Could you doxygen this function, including the new `String msg` argument?

src/provisioning/littlefs/WipperSnapper_LittleFS.cpp
```diff
@@ -99,11 +94,14 @@ void WipperSnapper_LittleFS::parseSecrets() {
...
   while (1) {
+    WS_DEBUG_PRINTLN("ERROR: Halted execution!");
```
brentru, line 99: Maybe "Fatal Error: {error msg}" instead?

src/provisioning/littlefs/WipperSnapper_LittleFS.h
```diff
@@ -33,7 +33,7 @@ class WipperSnapper_LittleFS {
   WipperSnapper_LittleFS();
   ~WipperSnapper_LittleFS();
   void parseSecrets();
-  void fsHalt();
+  void fsHalt(String msg = "");
```
brentru, line 36: Instead of providing a default argument of `""` here, we should _require_ anything calling `fsHalt()` to include the message.

src/provisioning/littlefs/WipperSnapper_LittleFS.cpp
```diff
@@ -99,11 +94,14 @@ void WipperSnapper_LittleFS::parseSecrets() {
...
+    WS_DEBUG_PRINTLN(msg.c_str());
     statusLEDSolid(WS_LED_STATUS_FS_WRITE);
     delay(1000);
```
brentru, line 102: Do we need this delay?

src/provisioning/littlefs/WipperSnapper_LittleFS.cpp
```diff
@@ -99,11 +94,14 @@ void WipperSnapper_LittleFS::parseSecrets() {
   LittleFS.end();
 }
 
-void WipperSnapper_LittleFS::fsHalt() {
+void WipperSnapper_LittleFS::fsHalt(String msg) {
   while (1) {
+    WS_DEBUG_PRINTLN("ERROR: Halted execution!");
+    WS_DEBUG_PRINTLN(msg.c_str());
     statusLEDSolid(WS_LED_STATUS_FS_WRITE);
```
brentru, line 101: Does this LED function need to get set within the loop? Can we do it outside it?

src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -95,9 +95,9 @@ Wippersnapper_FS::Wippersnapper_FS() {
 
   // If a filesystem does not already exist - attempt to initialize a new
   // filesystem
-  if (!initFilesystem()) {
-    WS_DEBUG_PRINTLN("ERROR Initializing Filesystem");
+  if (!initFilesystem() && !initFilesystem(true)) {
     setStatusLEDColor(RED);
+    fsHalt("ERROR Initializing Filesystem");
     while (1)
```
brentru, line 101: Remove L101 `while(1)` as execution wont reach this point

src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -107,7 +107,7 @@ Wippersnapper_FS::Wippersnapper_FS() {
 
   // If we created a new filesystem, halt until user RESETs device.
   if (_freshFS)
-    fsHalt();
+    fsHalt("New filesystem created! Please RESET your board.");
```
brentru, line 110: * Press the reset button on your board

src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -123,13 +123,13 @@ Wippersnapper_FS::~Wippersnapper_FS() {}
     @return   True if filesystem initialized correctly, false otherwise.
 */
 /**************************************************************************/
-bool Wippersnapper_FS::initFilesystem() {
+bool Wippersnapper_FS::initFilesystem(bool force_format) {
```
brentru, line 126: Please document `force_format` in the comment

src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -396,16 +393,24 @@ void Wippersnapper_FS::writeToBootOut(PGM_P str) {
...
 */
 /**************************************************************************/
-void Wippersnapper_FS::fsHalt() {
+void Wippersnapper_FS::fsHalt(String msg) {
+  TinyUSBDevice.attach();
+  delay(1500);
   while (1) {
+    WS_DEBUG_PRINTLN("ERROR: Halted execution!");
+    WS_DEBUG_PRINTLN(msg.c_str());
     // statusLEDSolid(WS_LED_STATUS_FS_WRITE);
```
brentru, line 414: Maybe we remove this line and the delay if the LED isn't turned on? and L414 is commented out? 

Would we want the LED on to signal to a user?

brentru (changes requested):

src/provisioning/littlefs/WipperSnapper_LittleFS.cpp
tyeth: Don't want to flood the serial, my pc doesn't like that, could be less. I debated this morning changing the LED to flash, even adding a parameter for the number of flashes (each cycle flips the LED until counter reached [reset] then skip next flash [wait 3sec] then repeat). We could then have nice error translation just from the number of flashes of a red LED. [neopixel probably shows enough already]

src/provisioning/tinyusb/Wippersnapper_FS.cpp
tyeth: I've now made it consistent with the LittleFS version, LED on solid for now indicating the user should check serial / files, with serial message repeated every second.

src/provisioning/littlefs/WipperSnapper_LittleFS.cpp
brentru: Why don't we print only once then, outside the loop, but flash indefinitely

src/Wippersnapper.cpp
tyeth: No, good shout, the filesystem should already be flushed

src/provisioning/littlefs/WipperSnapper_LittleFS.cpp
tyeth: Due to missing the serial message, which is common when the user is not expecting a problem and only connects serial after the error has been printed.

brentru:
Tested on FunHouse (secrets file persisted between reboots)

@tyeth  could you bump the library version and then merge+release?

brentru (approved):

PR #548, opened 2024-02-08
Add VL53L1X

tyeth:
Untested, ending for the day possibly, and wanted a couple of UF2's for testing next


src/components/i2c/drivers/WipperSnapper_I2C_Driver_VL53L1X.h
```diff
@@ -0,0 +1,101 @@
...
+  /*!
+      @brief    Gets the VL53L1X's current proximity.
+      @param    proximityEvent
+                Pointer to an Adafruit_Sensor event.
+      @returns  True if the proximity was obtained successfully, False
+                otherwise.
+  */
+  /*******************************************************************************/
+  bool getEventProximity(sensors_event_t *proximityEvent) {
+    if (!_VL53L1X->dataReady()) {
```
tyeth, line 80: I half wanted to add a delay / retry here, like try 3 times (potentially need 500ms integration time) then return NaN, hiding that logic in here felt a bit wrong even though I've done it once before. Instead I'm following the fail=false for dataready so it will infinitely retry. Maybe adding a delay makes sense, but probably at the retry loop instead of each driver. Just adding this comment so you're aware @brentru

brentru: What does the example driver do?

tyeth:
Ignoring the build size, this is now ready for review @brentru

src/components/i2c/drivers/WipperSnapper_I2C_Driver_VL53L1X.h
tyeth: nothing, just loops again, but there is a delay for visual clarity in the loop.

brentru:
Holding until https://github.com/adafruit/arduino-esp32/issues/18 is merged and tested. This PR will increase the available application flash and resolve the funhouse build error.

src/components/i2c/drivers/WipperSnapper_I2C_Driver_VL53L1X.h
tyeth, line 80: Add retry logic

src/components/i2c/drivers/WipperSnapper_I2C_Driver_VL53L1X.h
```diff
@@ -0,0 +1,106 @@
...
+  /*!
+      @brief    Gets the VL53L1X's current proximity.
+      @param    proximityEvent
+                Pointer to an Adafruit_Sensor event.
+      @returns  True if the proximity was obtained successfully, False
+                otherwise.
+  */
+  /*******************************************************************************/
+  bool getEventProximity(sensors_event_t *proximityEvent) {
+    uint8_t attempts = 0;
```
brentru, line 80: Remove the retry loop, bail out and `return false` if no data is ready.

src/components/i2c/drivers/WipperSnapper_I2C_Driver_VL53L1X.h
```diff
@@ -0,0 +1,106 @@
...
+    while (!_VL53L1X->dataReady()) {
+      if (attempts++ > 4) {
+        return false;
+      }
+      delay(150);
+    }
+
+    int16_t proximityMM = _VL53L1X->distance();
+    if (proximityMM <= 0 || proximityMM > 4000) {
```
brentru, line 89: The [example](https://github.com/adafruit/Adafruit_VL53L1X/blob/main/examples/VL53L1X_simpletest/VL53L1X_simpletest.ino#L49) checks against a value of `-1.

Where are you getting this check from:
`if (proximityMM <= 0 || proximityMM > 4000)`?

src/components/i2c/drivers/WipperSnapper_I2C_Driver_VL53L1X.h
```diff
@@ -0,0 +1,106 @@
...
+      WS_DEBUG_PRINT("VL53L1X: Invalid proximity value:");
+      WS_DEBUG_PRINTLN(proximityMM);
+      WS_DEBUG_PRINT("VL53L1X status: ");
+      WS_DEBUG_PRINTLN(_VL53L1X->vl_status);
+      proximityEvent->data[0] = NAN;
+    } else {
+      proximityEvent->data[0] = proximityMM;
+    }
+    _VL53L1X->clearInterrupt();
```
brentru, line 98: I am not sure if we need to call `clearInterrupt()` here. [On the example](https://github.com/adafruit/Adafruit_VL53L1X/blob/main/examples/VL53L1X_simpletest/VL53L1X_simpletest.ino#L49), an invalid read value of `-1` `return`s from the `loop()` before the interrupt is cleared.

brentru (changes requested):
@tyeth  Requesting changes across this PR

brentru:
@tyeth  Why is funhouse failing? Is this PR simply using the old workflow and can't be updated until we merge?

brentru (approved):

PR #555, opened 2024-03-15
Change AnalogIO DEFAULT_HYSTERESIS to 0.02

tyeth:
Was experiencing some lower than expected resolution with the analog potentiometer, S3 TFT Feather. Double checked the code and it's good as far as resolution etc goes. Noticed the value for hysteresis was 0.2 and labelled 2%. The calculation in the update check is 
[(_analog_input_pins[i].prvPinVal * DEFAULT_HYSTERISIS);](https://github.com/tyeth/Adafruit_Wippersnapper_Arduino/blob/325dfb66c2470fe3933281e4c83176787bca6ce7/src/components/analogIO/Wippersnapper_AnalogIO.cpp#L369)

I'm running a build with 
`#define DEFAULT_HYSTERISIS 0.02 ///< Default DEFAULT_HYSTERISIS of 2%
`

### DIFF
--- a/src/components/analogIO/Wippersnapper_AnalogIO.h
+++ b/src/components/analogIO/Wippersnapper_AnalogIO.h
@@ -18,7 +18,7 @@
 
 #include "Wippersnapper.h"
 
-#define DEFAULT_HYSTERISIS 0.2 ///< Default DEFAULT_HYSTERISIS of 2%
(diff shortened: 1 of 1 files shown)


brentru (approved):
Conditionally approving this. While this is good for ESP32-S3, different MCUs have different analog read resolutions. We may want to create an issue to look at scaling this value by the read resolution.

PR #557, opened 2024-03-19
Fix: MAX17048 - Return NaN if no battery attached

tyeth:
This checks the version + ID before each read, returning NaN if not valid (no battery attached).

### DIFF
--- a/src/components/i2c/drivers/WipperSnapper_I2C_Driver_MAX17048.h
+++ b/src/components/i2c/drivers/WipperSnapper_I2C_Driver_MAX17048.h
@@ -56,7 +56,7 @@ public:
   /*******************************************************************************/
   bool begin() {
     _maxlipo = new Adafruit_MAX17048();
-    if (!_maxlipo->begin(_i2c) or !_maxlipo->isDeviceReady()) {
+    if (!_maxlipo->begin(_i2c)) {
       return false;
     }
     return true;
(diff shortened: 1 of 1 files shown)


tyeth (commented):
Tagging @brentru for review

tyeth:
Not sure whats gone on with the build there, errors suggest some bracket isn't closed or idf changed or something is fubar.
Will look when I get back later

brentru:
@tyeth  Run clang locally over `src/components/i2c/drivers/WipperSnapper_I2C_Driver_MAX17048.h`

src/components/i2c/drivers/WipperSnapper_I2C_Driver_MAX17048.h
```diff
@@ -56,11 +56,28 @@ class WipperSnapper_I2C_Driver_MAX17048 : public WipperSnapper_I2C_Driver {
...
 
+  /*******************************************************************************/
+  /*!
+      @brief Checks the response of the MAX17048 chip by verifying ID and version.
+      @returns true if the chip ID or version is invalid, false otherwise.
+   */
+  /*******************************************************************************/
+  bool deviceNotReady() {
+    uint8_t chipID = _maxlipo->getChipID();
+    if (chipID == 0x00 || chipID == 0xFF)
```
brentru, line 73: Can you #define these numbers?

src/components/i2c/drivers/WipperSnapper_I2C_Driver_MAX17048.h
```diff
@@ -56,11 +56,28 @@ class WipperSnapper_I2C_Driver_MAX17048 : public WipperSnapper_I2C_Driver {
...
+      return true;
+    uint16_t chipVersion = _maxlipo->getICversion();
+    if (chipVersion == 0x0000 || chipVersion == 0xFFFF)
```
brentru, line 76: Same here - can you check if they are present within the MAX library's header, and reference the macros or define them?

brentru (changes requested):

tyeth:
There are no longer changes required as the library has been augmented with returning NaNs, and the begin will fall if no battery attached (same check inside lib).

PR #569, opened 2024-04-16
Fix ESP32 WiFi.h missing + macAddress for arduino-esp32 v3.0.0-rc1

### LINKED ISSUE #568: Arduino-ESP32 v3.0.0-rc.1 changes network interface usage. Need to import wifi.h and use NetworkManager for MAC Address
Builds failed due to unavailability of certain defines in WiFi.h which is no longer imported by WiFiClientSecure.h which we import in the ESP32 network provisioning class. See here:
https://github.com/espressif/arduino-esp32/discussions/9496

Additionally there was an issue with the device id being generated (from MAC address) after arduino-esp32 v3.0.0-rc.1, see here: https://github.com/espressif/arduino-esp32/issues/9509
Resolved by switching to Network.macAddress (which calls through to NetworkManager which accesses the base ESP32 chips mac), the other network interfaces must be brought up (.begin() or scan etc) before they populate the individual interface mac addresses.

Build runs and passes (except C3 as still default partition size) in CI, as well as locally, for the VL53L4CD by adding changes to `src/network_interfaces/Wippersnapper_ESP32.h` in this commit:
https://github.com/tyeth/Adafruit_Wippersnapper_Arduino/commit/656f32d06b22f3f31646978101d15086223ad1d0#diff-c32ef9a1c2bfd1bc9a7cb064310e7433975912191664669be9dba5fbd743a823

Will be moved to separate PR.

tyeth:
Tagging @brentru for review.
This closes #568

### DIFF
--- a/src/network_interfaces/Wippersnapper_ESP32.h
+++ b/src/network_interfaces/Wippersnapper_ESP32.h
@@ -23,6 +23,7 @@
 #include "Adafruit_MQTT.h"
 #include "Adafruit_MQTT_Client.h"
 #include "Arduino.h"
+#include "WiFi.h"
 #include <WiFiClientSecure.h>
 extern Wippersnapper WS;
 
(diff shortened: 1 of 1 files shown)


tyeth:
CI passing, except the C3 and original FeatherESP32 (v1)

brentru (approved):
Looks good, failing on a target. Let's not release this until we fix the flash size issue for ESP32/C3

PR #573, opened 2024-04-19
move funhouse to esptool

tyeth:
This is needed to then get an alpha release through with a zip file for the funhouse_noota


tyeth:
This was tested a while ago, but it's being reintroduced ready for final alpha test, and is ready for review @brentru

brentru (approved):
Sure - let's try it

PR #574, opened 2024-04-22
Update network port/cert support for local testing

tyeth:
Adds support for local dev machine testing, by allowing esp32 to specify port, and also disable mqtt security checking if not staging/production url

### DIFF
--- a/src/network_interfaces/Wippersnapper_ESP32.h
+++ b/src/network_interfaces/Wippersnapper_ESP32.h
@@ -148,14 +148,16 @@ public:
   void setupMQTTClient(const char *clientID) {
     if (strcmp(WS._config.aio_url, "io.adafruit.com") == 0) {
       _mqtt_client->setCACert(_aio_root_ca_prod);
-    } else {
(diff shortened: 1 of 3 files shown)


tyeth:
Hey @brentru can you give this one a review please.
Will allow me to do end to end testing locally

brentru (changes requested):
Looks like it addresses the cert issue, one question

tyeth:
We default it to 8883 but it is also possible to override via Json, which
is what I need to do for the local testing on esp³², although the cert
insecure should allow local 8883 to work (if ssl setup)

brentru:
>  We default it to 8883 

Where do we do this?

tyeth:
> > We default it to 8883
> 
> Where do we do this?

https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/blob/f51d4ac2e8baa8424717ce2e3a724b7bb5b40289/src/provisioning/ConfigJson.cpp#L49

brentru (approved):

PR #577, opened 2024-04-24
Add NAU7802

tyeth:
Matching component definition at https://github.com/adafruit/Wippersnapper_Components/pull/217

Atrifacts need testing still.


tyeth:
Appreciate some initial feedback @brentru 
Not to be merged before I test on staging (I'll merge)

src/components/i2c/drivers/WipperSnapper_I2C_Driver_NAU7802.h
```diff
@@ -0,0 +1,102 @@
...
+      WS_DEBUG_PRINTLN("Failed to set LDO to 3V0");
+    }
+
+    if (!_nau7802->setGain(NAU7802_GAIN_128)) {
+      WS_DEBUG_PRINTLN("Failed to set gain to 128");
+    }
+
+    if (!_nau7802->setRate(NAU7802_RATE_10SPS)) {
+      WS_DEBUG_PRINTLN("Failed to set sample rate to 10SPS");
+    }
```
brentru, line 72: Are these are all failure states that should `return false` if chip communication fails? Would you want to refactor this out into a `bool configure_nau7802()` function that returns based on if configuration fails or succeeds?

src/components/i2c/drivers/WipperSnapper_I2C_Driver_NAU7802.h
```diff
@@ -0,0 +1,102 @@
...
+                Pointer to an Adafruit_Sensor event.
+      @returns  True if the reading was obtained successfully, False otherwise.
+  */
+  /*******************************************************************************/
+  bool getEventRaw(sensors_event_t *rawEvent) {
+    unsigned long start = millis();
+
+    // Wait for the sensor to be ready
+    while (!_nau7802->available()) {
+      if (millis() - start > 1000) {
```
brentru, line 89: Could you change this timeout from `1000ms` to `250ms`, and make it a macro?

src/components/i2c/drivers/WipperSnapper_I2C_Driver_NAU7802.h
```diff
@@ -0,0 +1,102 @@
...
+        WS_DEBUG_PRINTLN("NAU7802 not available");
```
brentru, line 90: Potentially change the warning to reflect the docstring, "Whether there is new ADC data to read", rather than saying the sensor is not available.

src/components/i2c/drivers/WipperSnapper_I2C_Driver_NAU7802.h
```diff
@@ -0,0 +1,102 @@
...
+        return false;
+      }
+    }
+    rawEvent->data[0] = (float)_nau7802->read();
```
brentru, line 94: Would we want to support a raw force value in Adafruit_Sensor?

brentru (changes requested):

src/components/i2c/drivers/WipperSnapper_I2C_Driver_NAU7802.h
tyeth: Thinking for now not to, but as Limor was talking about adding the HX711 soon (another load cell / wheatstone bridge sensor) we may wish to revisit that decision. I have added to an existing issue on the Adafruit_Sensor repo for this, with comments/thoughts, so that we can continue the discussion.

https://github.com/adafruit/Adafruit_Sensor/issues/39

tyeth:
@brentru forgot this ones also re-ready for review

src/components/i2c/drivers/WipperSnapper_I2C_Driver_NAU7802.h
```diff
@@ -0,0 +1,130 @@
...
+      return false;
+    }
+
+    if (!_nau7802->setRate(NAU7802_RATE_10SPS) &&
+        !_nau7802->setRate(NAU7802_RATE_10SPS)) {
+      WS_DEBUG_PRINTLN("Failed to set sample rate to 10SPS");
+      return false;
+    }
+
+    // Take 10 readings to flush out old readings (10 samples per second)
```
brentru, line 85: Please refactor into a `flush_nau` function or similar

brentru (approved):
approving conditionally for the one change (refactoring something out)

Issue #562, opened 2024-04-02
WipperSnapper firmware fails at network connection when loaded from MacOS but succeeds on Windows


BlitzCityDIY:
**Describe the bug**
WipperSnapper firmware fails at network connection when loaded from MacOS but succeeds on Windows.

***Arduino board***
ItsyBitsy ESP32

**To Reproduce**
1. Using the espwebtool, I erased the flash
2. Then I uploaded the Factory Reset bin
3. After the reset bin was running successfully, I went thru the WipperSnapper GUI in Chrome to load the firmware.
4. On the Connect! step, the yellow light flashes, blue flashes multiple times and then is solid blue

**Expected behavior**
When I follow these steps on Windows 10, the firmware loads properly and the device is added to my account. I'm able to add components as well.

**Which components are connected to your device**
N/A

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: MacOS Ventura 13.3 (M1)
 - Browser: Chrome
 - Version: Chrome 123.0/WipperSnapper v1.0.0-beta.78

**Additional context**
I tried taking the board with the firmware loaded via MacOS to my Windows 10 computer but it still hangs at the Connect! step (solid blue light)

Happy to test with other boards too if that's helpful.

brentru:
@tyeth  Do you have a laptop similar to Liz' that you can replicate this with? If you don't, I'll self assign and try to replicate.

tyeth:
@brentru I've an Intel Mac (not M1) running a recent beta of Sonoma (14.2 with an update to 14.5 available). Can attempt to initially recreate, but if you have an M1/2/3 or still on v13 then maybe that would be better

brentru:
@tyeth  I have a M1 on Ventura, self-assigned this to myself.

brentru:
> the only ones working are beta 72 and beta 66

videopixil:
I've been testing on windows 11

brentru:
@videopixil @BlitzCityDIY I followed your replication steps. However, I can't replicate this with a Macbook Pro running Ventura 13.6.3 and a QT PY ESP32 Pico. 

Follow-up Questions:
* Is this only with the ESP32 Pico, or other boards have the same issue?

* When you connect your ESP32 to the mac - could you open Arduino IDE's serial monitor? Then, press RESET to let the board attempt to connect. You should see serial output - could you copy that and paste it here?

BlitzCityDIY:
[...]

PR #582, opened 2024-05-08
Fix: allow key length of 40 for older users

tyeth:
This relates to #562 and #579


tyeth:
This should be fine @brentru

brentru (approved):
LGTM!

Issue #585, opened 2024-05-09
PyPortal Titano - Wippersnapper gets disconnected from WiFI and never reconnects


T-Mosher:
Reference this forum thread:
https://forums.adafruit.com/viewtopic.php?t=210269

In short, I have a PyPortal Titano that monitors the state of charge of a LiPo battery via an I2C interface. It reports the battery status every one minute. It is installed in an area that has weak/intermittent WiFi service.

At random intervals (ranging from hourly to weekly), the device's Adafruit IO status goes to Offline.

If I monitor the USB serial port connection, the code is still running and pinging, but I2C publishing is not successful.

The code never recovers from this condition, until I press the Reset button.

Logs are in the forum thread.

***Arduino board***
PyPortal Titano: SW v1.0.0-alpha.82.

**To Reproduce**
See description above and in the forum thread.

**Expected behavior**
I expect the Wippersnapper software to automatically recover if an I2C or WiFi connection failure occurs.

**Which components are connected to your device**
The issue occurs with either an Adafruit INA219, or with the Adafruit MAX17048 battery gauge via the I2C interface.

The connection failure is not related to the state of charge of the battery. The Titano is running from USB +5V power. The battery, charger, and a 60 mA load are only an experiment (so the feed has something useful to report).

After verifying that the issue occurs with either of those sensors alone, I recently added an HTU31D temperature sensor.

I also have the two discrete outputs from the Titano connnected to an Adafruit latching relay (for controlling a LiPo battery charger module).

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Safari
 - Version: current release

T-Mosher:
Follow-ups:
The NINA firmware has been updated.

Here is one of the serial port logs recorded while the Titano SW is running but isn't reconnecting the WiFi:

PING!
PING!
PING!
PING!
Sensor 0x41
Temperature: 22.86 degrees C
Sensor 0x41
Humidity: 32.68%RH
PUBLISHING -> I2C Device Sensor Event Message...ERROR: Failed to encode and publish I2CDeviceEvent!
PING!
PING!
PING!
PING!
PING!
STATUS LED BLINK KAT
Executing periodic event on A16
Publishing pinEvent...Published!
Sensor 0x40
Voltage: 3.66 v
Sensor 0x40
Current: 59.20 mA
PUBLISHING -> I2C Device Sensor Event Message...ERROR: Failed to encode and publish I2CDeviceEvent!
[...]

PR #589, opened 2024-05-31
Add RSSI logging

tyeth:
Adds print of RSSI value on connect and every MQTT ping, as part of improved diagnostic info and resolving #585


tyeth:
@brentru ready for review

src/network_interfaces/ws_networking_pico.h
```diff
@@ -137,6 +139,16 @@ class ws_networking_pico : public Wippersnapper {
     memcpy(WS._macAddr, mac, sizeof(mac));
   }
 
+  /********************************************************/
+  /*!
+  @brief  Gets the current network RSSI value
+  */
+  /********************************************************/
+  void getRSSI() {
```
brentru, line 147: Would you instead want to return the expected value of the RSSI here, instead of storing it?

src/network_interfaces/ws_networking_pico.h
```diff
@@ -108,8 +108,10 @@ class ws_networking_pico : public Wippersnapper {
 
     // Was the network within secrets.json found?
     for (int i = 0; i < n; ++i) {
-      if (strcmp(_ssid, WiFi.SSID(i)) == 0)
+      if (strcmp(_ssid, WiFi.SSID(i)) == 0) {
+        WS._RSSI = WiFi.RSSI(i);
```
brentru, line 112: @tyeth  What do you think about returning the RSSI from the `getRSSI()` function call, rather than storing it and then calling the variable directly to return it?

tyeth: I think at this point we aren't connected, but instead examining matching networks, so the rssi call from WiFi needs a network number instead of empty argument. I guess we could have an extra overload of getRSSI with and without network number.

src/network_interfaces/ws_networking_pico.h
tyeth: Not really sure, that sounds sensible, was mostly influenced by getMacAddr.
 Curious how this will play out once we want to return it in a feed. No difference really that I can see.

src/network_interfaces/ws_networking_pico.h
brentru: Let's refactor to print using `getRSSI` only so the data only comes from one code section.

src/network_interfaces/ws_networking_pico.h
brentru: Let's try it!

src/network_interfaces/ws_networking_pico.h
tyeth: @brentru done

tyeth:
This is part of pb-error-msgs

PR #594, opened 2024-06-25
Latest platformIO changes for ESP32 BSP v3.0.2

tyeth:
This is currently working for me @brentru , with the caveat that I had to remove my existing platform packages from my platformio folder inside my home folder. I actually uninstalled all espressif32 platforms, then went and removed the framework-arduinoesp32 related folders from the platform packages location.

 "I had some leftover platforms/framework-arduinoesp32's for PIO still installed, in my home folder (C:\Users\tyeth\.platformio\packages), and removing them has got me past the build error."

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -76,11 +76,12 @@ lib_deps =
     https://github.com/adafruit/WiFiNINA.git
     https://github.com/Starmbi/hp_BH1750.git
 
+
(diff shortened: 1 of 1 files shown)


brentru:
Thanks for putting this together, I can compile and upload to my Feather ESP32V2

PR #596, opened 2024-06-27
Downgrade patch for ESP32-S3 BSP 3.0.2

brentru:
arduino-esp32 board support package v3.0.2 causes an infinite reboot error when used with ESP32-S3 boards with PSRAM. 

This patch transfers the ci for these build targets to use the brentru/ci-arduino fork, which has manual instructions to build WipperSnapper using esp32 BSP 3.0.1 for the affected boards.

### DIFF
--- a/.github/workflows/build-clang-doxy.yml
+++ b/.github/workflows/build-clang-doxy.yml
@@ -22,69 +22,69 @@ jobs:
         include:
           - offset: "0x1000"
     steps:
-    - uses: actions/setup-python@v4
-      with:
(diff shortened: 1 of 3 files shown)


brentru:
Before merging:
* Verify in the CI logs that these platforms are correctly built on bsp 3.0.1
* This needs testing on the Feather ESP32-S3 Rev. TFT
* If the above is OK, use this PR to cut version for a new release

brentru:
Successfully tested with Feather w/PSRAM on BSP 3.0.0
image

PR #600, opened 2024-07-01
Adjust SGP30 init to not call IAQinit unnecessarily

tyeth:
~~Minor tweak to the SGP30 initialisation to fail if IAQ doesn't initialise successfully instead of ignoring it.
Needs retesting - was in the way on another branch~~

Updated PR to be for the removal of the IAQinit call, as it's called here https://github.com/adafruit/Adafruit_SGP30/blob/b4a631b1d75996a4dc2b1dda07e02a8529365d33/Adafruit_SGP30.cpp#L73


tyeth:
Just read this through and we don't need to call IAQinit as begin does that for us. Will adjust this PR for that instead.

tyeth:
@brentru This is ready for review

brentru (approved):
LGTM

PR #606, opened 2024-07-24
Updates for the ESP32 network interface

brentru:
This pull request updates `src/network_interfaces/Wippersnapper_ESP32.h`:
* Switches from depreciated `WiFiClientSecure` to arduino-esp32 3.x `NetworkClientSecure` object
* Fixes inability to connect to an MQTT test broker without SSL/TLS by adding an option to use `NetworkClient` instead of `NetworkClientSecure`.
* Updates the CA certificate for the io.adafruit staging server 

All of these changes have been tested on an Adafruit QtPY ESP32-S2

### DIFF
--- a/src/network_interfaces/Wippersnapper_ESP32.h
+++ b/src/network_interfaces/Wippersnapper_ESP32.h
@@ -8,7 +8,7 @@
  * please support Adafruit and open-source hardware by purchasing
  * products from Adafruit!
  *
- * Copyright (c) Brent Rubell 2020-2021 for Adafruit Industries.
+ * Copyright (c) Brent Rubell 2020-2024 for Adafruit Industries.
  *
  * MIT license, all text here must be included in any redistribution.
  *
@@ -25,7 +25,8 @@
 #include "Arduino.h"
 #include "WiFi.h"
 #include "WiFiMulti.h"
-#include <WiFiClientSecure.h>
+#include <NetworkClient.h>
+#include <NetworkClientSecure.h>
 extern Wippersnapper WS;
 
 /****************************************************************************/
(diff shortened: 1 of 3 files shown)


src/network_interfaces/Wippersnapper_ESP32.h
```diff
@@ -156,18 +158,24 @@ class Wippersnapper_ESP32 : public Wippersnapper {
...
+      _mqtt_client_secure->setCACert(
+          strcmp(WS._config.aio_url, "io.adafruit.com") == 0
+              ? _aio_root_ca_prod
+              : _aio_root_ca_staging);
+      WS._mqtt = new Adafruit_MQTT_Client(
+          _mqtt_client_secure, WS._config.aio_url, WS._config.io_port, clientID,
+          WS._config.aio_user, WS._config.aio_key);
     } else {
-      _mqtt_client->setInsecure();
+      // Insecure connections require a NetworkClient object rather than a
```
tyeth, line 172: Looks good, although not so keen that this disables SSL fully for local testing, would prefer a flag, but it also feels a bit unnecessary as we can always fudge it to test local ssl.
Probably makes sense to test the same tweak for Pico + nina

tyeth (changes requested):
Requesting the same secure/insecure client usage for Pico + NINA

brentru:
@tyeth  Not entirely sure if Nina/Airlift needs this, they are configured to be SSL-only

brentru:
@tyeth  Requesting a re-review. I've changed the Pico network interface to reflect these changes.

tyeth (approved):
This looks great. Good spot on Wifi101, I'm assuming this needs to be ignored.

brentru:
> This looks great. Good spot on Wifi101, I'm assuming this needs to be ignored.

Exactly, otherwise, the lib has a conflict between WiFi101 and WiFi.h

PR #623, opened 2024-08-23
Add HDC3021/HDC302x

tyeth:
This adds the HDC302x driver.
Upon boot the first reading for humidity was reasonable however the temp readings were -45c, so now it skips the first reading.


tyeth:
I can imagine you asking for the debug prints to be removed, which makes sense. I could leave the two status reasons as comments or comment out the debug prints

brentru (approved):
LGTM, addresses the issue you mentioned with first reading correctly

Issue #591, opened 2024-06-10
QTPY S3 will not load the WIPPER drive


Timeline8:
Last night, following guide on [https://io.adafruit.com/Timeline/device ... uf2/step-2](https://io.adafruit.com/Timeline/devices/new/qtpy-esp32s3-n4r2/uf2/step-2)
And also referencing [https://learn.adafruit.com/adafruit-qt- ... pper-setup](https://learn.adafruit.com/adafruit-qt-py-esp32-s3/wippersnapper-setup)
Booted Qt Py S3 into bootloader mode with double press of the reset button. File **INFO_UF2.TXT** reports the following:

TinyUF2 Bootloader 0.18.2 - tinyusb (0.15.0-566-gf1e006d09)
Model: Adafruit QT Py ESP32-S3 (4M Flash, 2M PSRAM)
Board-ID: ESP32S3-QTPy-N4R2-A
Date: Mar 22 2024
Flash Size: 0x00160000 bytes

Downloaded latest version of Whippersnapper for my board named **wippersnapper.qtpy_esp32s3_n4r2.1.0.0-beta.83.uf2**

Drag and dropped file onto **QTPYS3BOOT**

Board restarted but I was not offered a **WIPPER** drive nor any drive at all. Tried hitting reset and also hard power cycling the board by pulling the USB. No drive appearing and NeoPixel is showing a steady green-yellow color. Board was previously running CircuitPython however I am not presented with a **CIRCUITPY** drive either. Confirmed chip silkscreen on the board is “FHR2...” so 4MB Flash/2MB PSRAM.

Following instructions on [https://learn.adafruit.com/adafruit-qt- ... tory-reset](https://learn.adafruit.com/adafruit-qt-py-esp32-s3/factory-reset), downloaded Factory Reset Firmware UF2 and after soft reset confirmed visually by the LED "NeoPixel rainbow swirl example” was running.

[...]

PR #625, opened 2024-09-05
Fix TinyUSB Enumeration bug on ESP32-S3

brentru:
This pull request updates the TinyUSB dependency to include a patch for the ESP32-S3's USB enumeration bug first reported in #591. I've also updated WipperSnapper's USB-MSC Filesystem class to force enumeration per @hathach 's advice in https://github.com/adafruit/Adafruit_TinyUSB_Arduino/issues/436#issuecomment-2331186664

Related patch in TinyUSB: https://github.com/adafruit/Adafruit_TinyUSB_Arduino/pull/454
Resolves: https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/591


src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -209,8 +209,14 @@ void Wippersnapper_FS::initUSBMSC() {
   // init MSC
   usb_msc.begin();
 
-  // re-attach the usb device
-  TinyUSBDevice.attach();
+  // If already enumerated, additional class driver begin() e.g msc, hid, midi won't take effect until re-enumeration
+  if (TinyUSBDevice.mounted()) {
```
hathach, line 213: ah, this seems to imply it is already previously detach(), if so, this change may not be needed since we already force it to re-enumerate. 

PS: detach()/attach() previously not working with esp32s3 before latest update.

brentru: @hathach Yeah, it is previously calling `detach()` in the constructor, prior to this call:

https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/625/files#diff-2a38826e0cb37f744e63c8de81606ed9566e257d4ff9a3667b6e42fe68298304R93

Provided that, do we need to force enumeration here - or is simply calling `attach()` and waiting, to re-attach and mount the USB device OK? Should I call `attach()` and then wait for `mounted()`?

hathach: no we don't really need to force it anymore. I think we are good, just update the tinyusb which fixes the disconnect/connect with S3. Previously the code does not work well since arduino-esp32 change the USB muxing back and forth between USB OTG and JTAG.

brentru:
Tested on Adafruit Feather ESP32-S3 Reverse TFT

Issue #627, opened 2024-09-09
Data recorded in stuttering manner with SCD-30 and many boards (QT PY S2, QT PY S3, etc) - in Release 0.88


kcowin:
**Describe the bug**

SCD-30 sensor has four feeds that can be accessed. When sending the feeds
to wippersnapper every 30 seconds, they show up randomly.

![Screenshot Data Problem 2024-09-06 165114](https://github.com/user-attachments/assets/92709096-47a4-4693-87fd-07ab85777bcd)

**To Reproduce**

Attach an SCD-30 to a QT PY Board and see if you can get better results.

**Expected behavior**
You will likely see missing values, if any at all.

**Which components are connected to your device**
Which components are set up on the WipperSnapper device page?

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**
Problem appears regardless of browser or browsing platform. There are no data errors
on the wippersnapper subscription page.

tyeth:
Provisional fix in PR. @kcowin if you wish to try the build assets they are attached 
[issue632-qtpy-assets.zip](https://github.com/user-attachments/files/17063533/issue632-qtpy-assets.zip)

tyeth:
I've been running it for a couple of days, at 30second intervals so about 3500(3467) readings. 

The temp/humidity get read in one command, and then the CO2 afterwards (so more chance of a failure).

Approximately 35 CO2 readings (1%) were more that 30seconds apart, the majority at 60s apart (so a reading was skipped), a sprinkle at 90s (two readings skipped), and a few further apart (presumably connectivity issues or account throttle).

Temperature readings had 10 entries out of 3500 that didn't fall in the 30second poll period.

![image](https://github.com/user-attachments/assets/ceda8821-9beb-49d6-9c94-07ce2b80f704)

kcowin:
Yes- I noticed it was not perfect, but a substantial improvement, and now the data is usable. KenSent from my iPhoneOn Sep 23, 2024, at 7:57 AM, Tyeth Gundry ***@***.***> wrote:﻿
I've been running it for a couple of days, at 30second intervals so about 3500(3467) readings.
The temp/humidity get read in one command, and then the CO2 afterwards (so more chance of a failure).
[...]

PR #632, opened 2024-09-19
Add retry if data not ready for SCD30

tyeth:
This updates the SCD30 driver to pull the data for all metrics in one go, rather than each one in turn which caused many data_ready wait issues - appearing as no data for most metrics.

### DIFF
--- a/src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD30.h
+++ b/src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD30.h
@@ -24,7 +24,8 @@
     @brief  Class that provides a driver interface for the SCD30 sensor.
 */
 /**************************************************************************/
-class WipperSnapper_I2C_Driver_SCD30 : public WipperSnapper_I2C_Driver {
+class WipperSnapper_I2C_Driver_SCD30 : public WipperSnapper_I2C_Driver
+{
 
 public:
   /*******************************************************************************/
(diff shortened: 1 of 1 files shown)


tyeth (commented):
@brentru ready for review.
This fixes #627

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD30.h
```diff
@@ -53,6 +53,36 @@ class WipperSnapper_I2C_Driver_SCD30 : public WipperSnapper_I2C_Driver {
...
+      @returns  True if the sensor was read successfully, False otherwise.
+  */
+  /*******************************************************************************/
+  bool readSensor() {
+    // dont read sensor more than once per second
+    if (_lastRead != 0 && millis() - _lastRead < 1000) {
+      return true;
+    }
+
+    if (!_scd->dataReady()) {
```
brentru, line 68: This segment is difficult to read. Could you refactor the logic out into to a new function?

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD30.h
```diff
@@ -64,14 +94,11 @@ class WipperSnapper_I2C_Driver_SCD30 : public WipperSnapper_I2C_Driver {
...
-      return false;
-
-    // attempt to get temperature data
-    sensors_event_t humidEvent;
-    if (!_scd->getEvent(&humidEvent, tempEvent))
+    if (!readSensor()) {
       return false;
+    }
 
+    tempEvent->temperature = _temperature;
```
brentru, line 101: I think it's confusing to mix types like this (`_temperature` is a `float`, `tempEvent` is a `sensors_event_t`). 

Instead, could we have:
*  `_temperature`, `_humidity` and `_co2` declared as protected `sensors_event_t` types?
* You could also remove the need for the (`sensors_event_t tempEvent;` and `sensors_event_t humidEvent;`) decls. on L77:L78. 
* You could then remove the (now redundant) reassignments in L79:L81, too.

brentru (changes requested):
@tyeth  I agree with the overall structure, I do have some comments and questions about the types your using and how they're being packed.

brentru (approved):
This reads much better now, thanks for making adjustments.

PR #635, opened 2024-09-27
Use analogReadMilliVolts for esp32 and update hysteresis

tyeth:
~~For CI assets, testing https://github.com/tyeth/Adafruit_Wippersnapper_Arduino/blob/50f17026832ba9a53d0f919a69b8e9106302fe55/src/components/analogIO/Wippersnapper_AnalogIO.cpp#L90~~

Combined PR including millivolt change and updated hysteresis method:

This PR now switches both periodic and on_change to use the same functions, along with using analogReadMilliVolts for ESP32 based boards.

Also this modifies the on_change event to use a sliding scale of hysteresis. Originally the code awaited a 2% change in the last raw value.
Now the scale is divided into thirds, with the lower third using the default hysteresis value of 2% applied to the whole 16bit max value (65k), which results in less bounce around 0 due to wifi / power fluctuations, the middle third using double that threshold change, and the last third quadruple that default hysteresis (where a 200mV spike was generally observed around 3.1v). There is also a 500ms anti-hammer situation on top of the hysteresis requirement.

Tested with a 10k stemma Potentiometer, and separately with a resistor divider to test stability.

It occurs to me that 500ms might be too long for some people, maybe it can be configurable in the future.

### DIFF
--- a/src/components/analogIO/Wippersnapper_AnalogIO.cpp
+++ b/src/components/analogIO/Wippersnapper_AnalogIO.cpp
@@ -325,7 +325,7 @@ bool Wippersnapper_AnalogIO::encodePinEvent(
 */
 /**********************************************************/
 void calculateHysteresis(analogInputPin pin, uint16_t pinValRaw,
-                         uint16_t pinValThreshHi, uint16_t pinValThreshLow) {
+                         uint16_t *pinValThreshHi, uint16_t *pinValThreshLow) {
   // All boards ADC values scaled to 16bit, in future we may need to
   // adjust dynamically
   uint16_t maxDecimalValue = 65535;
@@ -344,8 +344,8 @@ void calculateHysteresis(analogInputPin pin, uint16_t pinValRaw,
   }
 
   // get the threshold values for previous pin value
-  pinValThreshHi = pin.prvPinVal + CURRENT_HYSTERISIS;
-  pinValThreshLow = pin.prvPinVal - CURRENT_HYSTERISIS;
+  *pinValThreshHi = pin.prvPinVal + CURRENT_HYSTERISIS;
+  *pinValThreshLow = pin.prvPinVal - CURRENT_HYSTERISIS;
 }
 
 /**********************************************************/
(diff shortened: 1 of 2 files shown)


tyeth:

src/components/analogIO/Wippersnapper_AnalogIO.cpp
```diff
@@ -310,13 +318,17 @@ bool Wippersnapper_AnalogIO::encodePinEvent(
...
+    @param periodOffset
+           Offset to add to the pin's period (used for on_change).
     @returns True if pin's period expired, False otherwise.
 */
 /**********************************************************/
-bool Wippersnapper_AnalogIO::timerExpired(long currentTime,
-                                          analogInputPin pin) {
-  if (currentTime - pin.prvPeriod > pin.period && pin.period != 0L)
+bool Wippersnapper_AnalogIO::timerExpired(long currentTime, analogInputPin pin,
+                                          long periodOffset) {
```
brentru, line 327: Do we ever pass `periodOffset` in? When this func. is called on L531, it doesn't seem to include periodOffset.

When would we include `periodOffset` and why?

src/components/analogIO/Wippersnapper_AnalogIO.cpp
```diff
@@ -359,35 +370,61 @@ void Wippersnapper_AnalogIO::update() {
...
+        if (_analog_input_pins[i].prvPeriod != 0L &&
+            !timerExpired(millis(), _analog_input_pins[i], 500)) {
+          continue;
+        }
+
         // note: on-change requires ADC DEFAULT_HYSTERISIS to check against prv
         // pin value
         uint16_t pinValRaw = getPinValue(_analog_input_pins[i].pinName);
 
+        // All boards ADC values scaled to 16bit, in future we may need to
```
brentru, line 389: Could you refactor this into a `CalculateHysteresis` function and call that function here?

brentru (changes requested):

src/components/analogIO/Wippersnapper_AnalogIO.cpp
tyeth: I reuse the function / logic check for the on change delay, passing in 500, on line 381 https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/635/files#diff-4e111e8f7d844aa38d092958262d5b77683072105409145224c36ab513c0109cR381

tyeth:
@brentru this should be ready again, retested it after the refactor

PR #599, opened 2024-06-27
Add debug build for Adafruit ESP32-S3 Rev TFT Feather

brentru:
Adding a CI build target for a dev build for ESP32-S3 Reverse TFT Feather

Related: https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/591

PR #643, opened 2024-10-11
New pio build targets

tyeth:
This adds a couple of targets to the platformIO config, specifically the funhouse, qtpys3_n4r2 and esp32s3devkitc.
Also adds a CI build target for the debug build of reverse tft S3 feather, superseding #599


tyeth:

brentru (approved):
Thanks for adding the additional platforms and the comment for delineating lvgl deps and regular deps

PR #644, opened 2024-10-18
API V2 Review - DS18x20 Component

### LINKED ISSUE #638: Refactor DS18x Component for API v2
- [x] Implement new ds18x20 component classes
  - [x] Add controller class for ds18x20
  - [x] Add hardware class for interfacing with the OneWire API and DallasTemperature API
  - [x] Add model class for interfacing with the compiled protocol buffer .pb.h file - ds18x20.proto
- [x] Add interfaces with top-level application encoder/decoder
- [x] Functionally test and verify all ds18x20.proto messages
- [x] Doxygenify the new components
- [x] Refactor Hardware class to use OneWireNG instead of DallasTemperature library (resolves https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/617)

brentru:
I've cherry-picked only the commits for the DS18x20 component and selected the base as the migration branch, the file sto review should _mostly_ reflect the DS18x20 API.

Please **do not merge** this branch, this is a review-only PR.

Resolves #638


tyeth (commented):
Was initially confused to see changes to the SensorModel class in src/components/sensor/model.cpp and .h, but they're presumably no longer used.

src/components/ds18x20/hardware.cpp
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  // Set common resolution for all sensors.
+  // Th, Tl (high/low alarm triggers) are set to 0.
+  _drv_therm.writeScratchpadAll(0, 0, _resolution);
+
+  // The configuration above is stored in volatile sensors scratchpad
+  // memory and will be lost after power unplug. Therefore store the
+  // configuration permanently in sensors EEPROM.
+  _drv_therm.copyScratchpadAll(false);
```
tyeth, line 119: not sure if we want false for copyScratchpadAll, which looks like it disables parasitic device support for that command

src/components/ds18x20/hardware.cpp
```diff
@@ -0,0 +1,195 @@
...
+  // sensor id while reissuing readScratchpadSingle() calls.
+  // Note, due to its storage class the placeholder is zero initialized.
+  static Placeholder<DSTherm::Scratchpad> scrpd;
+  ec = _drv_therm.readScratchpad(_sensorId, scrpd);
+  if (ec != OneWireNg::EC_SUCCESS)
+    return false;
+
+  // Read the temperature from the sensor
+  long temp = scrpd->getTemp2();
+  _temp_c = temp / 16.0; // Convert from 16-bit int to float
```
tyeth, line 191: not 16-bit int, maybe something like:
`// Convert from 16*int to float (avoids precision loss)`

src/components/ds18x20/model.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+/*************************************************************************/
+/*!
+    @brief  Encodes a Ds18x20Event message.
+    @param  onewire_pin
+            The OneWire bus' pin where the sensor is located.
+    @param  sensor_events_count
+            The number of sensorevent messages to encode.
+    @param  sensor_events
+            The sensorevent messages to encode.
```
tyeth, line 143: function takes no arguments

src/components/ds18x20/model.cpp
```diff
@@ -0,0 +1,192 @@
...
+  return pb_encode(&msg_stream, wippersnapper_ds18x20_Ds18x20Event_fields,
+                   &_msg_DS18x20Event);
+}
+
+/*************************************************************************/
+/*!
+    @brief  Initializes the Ds18x20Event message.
+*/
+/*************************************************************************/
+void DS18X20Model::InitDS18x20EventMsg(const char *ow_pin_name) {
```
tyeth, line 165: add to args list in docstring

src/components/ds18x20/hardware.cpp
brentru: What is not a 16-bit Int?

src/components/ds18x20/hardware.cpp
brentru: Why wouldn't we want to disable parasitic devices? We can't choose between them with the current API - only enable or disable support for them.

tyeth: When I tested v1 ages ago I thought I'd also tested parasitic successfully.
I'm not fussed either way

src/components/ds18x20/hardware.cpp
brentru: @tyeth  I could be wrong! This is why we review each other's math.

Here's what I think this does,

* `getTemp2()` is the "Temperature in Celsius degrees returned as fixed-point integer with multiplier 16"
  * e.g. 20.125 C is returned as 322.
*  We read `temp` as a long, scaled by x16
* To get the real temperature, we divide that `long` value by 16
* Then, we assign the value to a float, `_temp_c`

tyeth:
100% agree that's what I believe it does too.
The comment confused me though as it doesn't mention the division ( /16)
instead talking about a 16bit integer data type.
I only understood after visiting the getTemp2 function

brentru:
I'm going to adjust the comment to reflect this conversation

PR #652, opened 2024-10-22
Add caching for nanopb

tyeth:
Caches are based on the current ref, or a parent ref / base branch, as well as key.
When we update nanopb we'll need to update the cache key too, it's obvious as there are a few lines mentioning the version number.

This has now run once, causing a cached item to appear under the ref/pulls/this-one cache, and the second run has used that cached item for all the jobs.
Once we get it into main then one run will generate a cached item in the main cache which other things can utilise.

### DIFF
--- a/.github/workflows/build-clang-doxy.yml
+++ b/.github/workflows/build-clang-doxy.yml
@@ -59,7 +59,7 @@ jobs:
         continue-on-error: true
         run: |
           wget https://jpa.kapsi.fi/nanopb/download/nanopb-0.4.8.tar.gz
-      - if: failure() == 'true'
+      - if: ${{ failure() }}
         name: Restore cached nanopb
         id: cache-nanopb-restore
         uses: actions/cache/restore@v4
@@ -68,7 +68,7 @@ jobs:
         with:
           path: ./nanopb-0.4.8.tar.gz
           key: nanopb-0.4.8.tar.gz
-      - if: failure() =='false'
+      - if: ${{ success() }}
         name: Save nanopb to cache
         id: cache-nanopb-save
         uses: actions/cache/save@v4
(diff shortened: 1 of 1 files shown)


tyeth:

.github/workflows/build-clang-doxy.yml
```diff
@@ -55,10 +55,20 @@ jobs:
...
           git clone --depth 1 --branch development https://github.com/brentru/Adafruit_LvGL_Glue.git /home/runner/Arduino/libraries/Adafruit_LittlevGL_Glue_Library
-      - name: Download and install stable Nanopb
+      - name: Cache nanopb
+        id: cache-nanopb
+        uses: actions/cache@v4
+        env:
+          cache-name: cache-node-modules
+        with:
+          path: ./nanopb-0.4.8.tar.gz
+          key: nanopb-0.4.8.tar.gz
```
tyeth, line 65: cache key containing version also in the download and install steps beneath

.github/workflows/build-clang-doxy.yml
```diff
@@ -55,10 +55,20 @@ jobs:
...
           git clone --depth 1 --branch wippersnapper https://github.com/brentru/lvgl.git /home/runner/Arduino/libraries/lvgl
           git clone --depth 1 --branch development https://github.com/brentru/Adafruit_LvGL_Glue.git /home/runner/Arduino/libraries/Adafruit_LittlevGL_Glue_Library
-      - name: Download and install stable Nanopb
+      - name: Cache nanopb
+        id: cache-nanopb
+        uses: actions/cache@v4
+        env:
+          cache-name: cache-node-modules
+        with:
+          path: ./nanopb-0.4.8.tar.gz
```
tyeth, line 64: path can be files or folders, relative etc.

brentru:
Could we first attempt to grab the direct download, then try the cache?

tyeth:
Sure that makes sense. I was initially leaning on the side of reducing
server load for nanopb

tyeth:
@brentru should be good to go now

.github/workflows/build-clang-doxy.yml
brentru: I think the key is good as-is, it's human readable vs a hash

brentru (approved):

brentru:
@tyeth  LGTM at your convenience

PR #654, opened 2024-10-23
Reduce connection errors on QT Py ESP32-Sx due to 3D Antennas

brentru:
The PR's change sets the WiFi transmission power to a reduced dBm before performing a network scan.

We are doing this to match CircuitPython (discussion https://github.com/adafruit/circuitpython/issues/9235) and reduce connection errors - https://github.com/adafruit/circuitpython/pull/9369/files

- [x] Test a few different power increments (-5, -3 dBM from standard 18dBm)
- [x] Add inclusion guards, add this for all QT Py boards since they share the same 3D antenna.

### DIFF
--- a/.github/workflows/build-clang-doxy.yml
+++ b/.github/workflows/build-clang-doxy.yml
@@ -33,8 +33,8 @@ jobs:
           echo >>$GITHUB_ENV WS_VERSION=$(git describe --dirty --tags)
       - uses: actions/checkout@v4
         with:
-          repository: tyeth/ci-arduino
-          ref: WS_ARDUINO_598-preprocessor-script-for-printing-the-library-versions-to-boot_outtxt
+          repository: adafruit/ci-arduino
+          ref: ci-wippersnapper
           path: ci
       - name: Checkout Board Definitions
         uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


tyeth:
This is a fantastic one! I wonder how many models should have custom levels. I had a quick look at the tx issue in general and saw some modules restricted their operations to 15dbm, see https://forum.arduino.cc/t/esp32-wroom-32-wifi-power-strangeness/1167585/8 , and on some openWRT page that mostly only power amplifier based boards support 18-20dbm values.

If the C3 PCB wifi antenna chip is the issue that would be a fair few boards, but I wonder if it's more SoM/board-layout specific.
It would be lovely to have a reliable test setup for this so we don't have to unnecessarily reduce the effectiveness of all boards.

It looks like the default is the 19.5 dbm mode, which is what it returns if you haven't selected wifi mode yet (similar to ignoring requests to set if not begun/mode-set) https://github.com/espressif/arduino-esp32/blob/master/libraries/WiFi/src/WiFiGeneric.cpp#L637

brentru:
It would be all boards with the 3D Antenna, which is all Qt Pys. I'm going to expand this PR.

brentru:
We're also bumping the CI to BSP 3.0.7,
https://github.com/adafruit/ci-arduino/pull/198

brentru:
@tyeth  Requesting a review of this PR - it also updates the workflow to build off adafruit/ci-arduino's branch for wippersnapper. I've pointed ci-arduino towards the latest bsp, 3.0.7. I've also updated the PIOArduino URL for 3.0.7 latest.

tyeth (approved):
These look good, I'll cast a eye over ci-arduino now

tyeth:
@brentru one more think, I couldn't see the platformIO.ini PioArduino change

tyeth (approved):

Issue #621, opened 2024-08-22
qt py loses secrets.json when battery goes down


migs35323:
**Describe the bug**

my qt-py s3 randomly loses the secrets.json credentials. when battery power goes out/down
i had noticed this issue in firmware beta .88 and right now i was testing with alpha .89 and same behavior
I usually power it through usb but i had been testing battery endurance, so i had been letting the battery go out on purpose.

as a side note, I once had the yellow light and boot out telling me the credentials where wrong, when i opened them in notepad/vscode, the file was correct. but i had to delete the file and recreate it to accept the credentials, however i can't seem to reproduce this again.
in this last behavior the only thing that comes to my mind, is there any sensitivity between LF and CRLF in the file? maybe i edited the file in windows and mixed it in ubuntu and WipperSnapper is not liking it...

***Arduino board***
qt-py s3 psram version aka qtpy-esp32s3-n4r2

**To Reproduce**
1. setup qt py with battery and best friend
2. let it run out of battery
3. when plugging it again it shows yellow light, boot.out warns about credentials, secrets.json is back to stock again.
PS: this doesn't happen every time, but about 2 out of 3 times it's not 100% consistent

**Expected behavior**
secrets,json to not be erased, 
or maybe turn it into a toggeable feature where we can choose the secrets to be erased as security feature lol

**Which components are connected to your device**
qt py s3 (psram version) with battery best friend attached + 2 adafruit temperature sensors connected with stemma qt, 

**Desktop (please complete the following information):**
 - OS: ubuntu 22 or windows 11

**Additional context**

brentru:
@tyeth  Do you have cycles available to look at this bug?

migs35323:
> @tyeth Do you have cycles available to look at this bug?

what do you mean by cycles? battery cycles?

brentru:
@migs35323  Tyeth is an Adafruit engineer who works on WipperSnapper with me. 

To clarify - I'm using "cycles" instead of "time", so I'm asking him if he has time to look at this or if I should.

tyeth:
👋 I've been running with shorter cycles  myself this week, medical issue, so likely to have a look around the middle of next week (once BFF arrives).
 It's possible that using a mix of line endings might be considered corrupt and it self formatting. At the same time I have seen similar behaviour before. Will report back.

tyeth:
Hardware just arrived, will setup a test

migs35323:
[...]

PR #656, opened 2024-10-25
No fs format nor file recreation and boot reason changes

tyeth:
This resolves a couple of things.
Firstly we now reattach USB Mass Storage Device after recreating the filesystem, this means that the drive now shows up on first boot after an initial firmware installation (tested with erased device). Previously the first boot would not show a drive, but subsequent boots would.

Secondly, in relation to brown-out situations, we now avoid erasing files unnecessarily, and if we last reset due to brownout we do not attempt to write any files like boot log nor erase the circuitpython files. There turns out to be little reason to write to the filesystem every boot. Mostly a load of `.exists` guards to avoid unnecessary erases, and checking the boot reason for some operations.

There's a minor tweak to the print_reset_reason method as it was being used incorrectly. It was being passed CPU core number instead of passing in the result of rtc_get_reset_reason(cpuCore).
Finally the pico now has a boot reason adding in the latest Earle core, so included that too.


tyeth:
Add getResetReason for rp2040/2350
See https://github.com/earlephilhower/arduino-pico/pull/2516/files

src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -105,17 +106,33 @@ Wippersnapper_FS::Wippersnapper_FS() {
 
   // If a filesystem does not already exist - attempt to initialize a new
   // filesystem
-  if (!initFilesystem() && !initFilesystem(true)) {
-    setStatusLEDColor(RED);
-    fsHalt("ERROR Initializing Filesystem");
+  if (!initFilesystem()) {
```
brentru, line 109: I am confused about tracing the brownout detection throughout the `FS` class.

Could you help clarify the following:

1) I see `brownOutCausedReset` is set within `get_and_print_reset_reason_for_cpu()`.
2) When does this function get called within FS? I don't see any calls outwards to this function, only the boolean `brownOutCausedReset` which isn't a function

src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -105,17 +106,33 @@ Wippersnapper_FS::Wippersnapper_FS() {
...
+    if (WS.brownOutCausedReset) {
+      // try once more for good measure
+      delay(10); // let power stablise after failure
+      if (!initFilesystem()) {
+        // no lights, save power as we're probably on a low battery
+        fsHalt("Brownout detected. Couldn't initialise filesystem.");
```
brentru, line 115: Could you please add some descriptive text/instructions on what to do if we couldn't recreate the FS after a brownout?

src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -105,17 +106,33 @@ Wippersnapper_FS::Wippersnapper_FS() {
...
+      }
+    } else if (!WS.brownOutCausedReset && !initFilesystem(true)) {
```
brentru, line 117: If a brownout was detected, all the handling for this was performed inside the `if` condition above and we don't need to work with/handle it anymore.

Consider changing this to an `if` and only handle the `initFileSystem(true` case.

src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -105,17 +106,33 @@ Wippersnapper_FS::Wippersnapper_FS() {
...
   // If we created a new filesystem, halt until user RESETs device.
   if (_freshFS)
-    fsHalt("New filesystem created! Press the reset button on your board.");
+  {
+    WS_DEBUG_PRINTLN("New filesystem created! Resetting the board shortly...");
+    WS.enableWDT(500);
+    while (1)
+    {
+      delay(1000);
+    }
```
brentru, line 134: This is great, I like the idea of resetting instead of instructing the user to reset!

Shorten this by doing:
```
for (;;) delay(1000);
```

src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -159,32 +176,39 @@ bool Wippersnapper_FS::initFilesystem(bool force_format) {
...
     return false;
 
+  //TODO: Don't do this unless we need the space and createSecrets fails
   // If CircuitPython was previously installed - erase CPY FS
   eraseCPFS();
-
-  // If WipperSnapper was previously installed - remove the
-  // wippersnapper_boot_out.txt file
-  eraseBootFile();
+  // Also, should probably relabel drive to WIPPER if CIRCUITPY was there
```
brentru, line 182: Would it be possible to do this within `eraseCPFS`, as the vlume was likely named CIRCUITPY?

src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -159,32 +176,39 @@ bool Wippersnapper_FS::initFilesystem(bool force_format) {
...
-    return false;
-  writeFile.close();
-
-  writeFile = wipperFatFs.open("/.metadata_never_index", FILE_WRITE);
-  if (!writeFile)
-    return false;
-  writeFile.close();
+  if (!wipperFatFs.exists("/.fseventsd/no_log"))
+  {
+    wipperFatFs.mkdir("/.fseventsd/");
```
brentru, line 188: While we're in it, let's refactor out the macOS-specific handling to a new function, `handleMacOS()`. This gives us room for handling other OS like Windows, Linux, ChromeOs the same way.

src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -159,32 +176,39 @@ bool Wippersnapper_FS::initFilesystem(bool force_format) {
...
+  {
+    File32 writeFile = wipperFatFs.open("/.Trashes", FILE_WRITE);
+    if (!writeFile)
+      return false;
+    writeFile.close();
+  }
 
   // Create wippersnapper_boot_out.txt file
-  if (!createBootFile())
+  if (!createBootFile() && !WS.brownOutCausedReset)
```
brentru, line 211: Do we need to check if a brownout happened here?

src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -269,44 +296,66 @@ void Wippersnapper_FS::eraseBootFile() {
...
-    // Print ESP-specific info to boot file
-    #ifdef ARDUINO_ARCH_ESP32
-    // Get version of ESP-IDF
-    bootFile.print("ESP-IDF Version: ");
-    bootFile.println(ESP.getSdkVersion());
-    // Get version of this core
-    bootFile.print("ESP32 Core Version: ");
-    bootFile.println(ESP.getCoreVersion());
-    #endif
+  if (WS.brownOutCausedReset){
```
brentru, line 344: Why are we checking for a brownout here, again?

src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -349,6 +397,8 @@ void Wippersnapper_FS::createSecretsFile() {
       "Please edit it to reflect your Adafruit IO and network credentials. "
       "When you're done, press RESET on the board.");
 #endif
+  delay(500); // previously 2500
```
brentru, line 400: Why are we decreasing this delay? It was from TinyUSB Arduino

src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -349,6 +397,8 @@ void Wippersnapper_FS::createSecretsFile() {
...
+  initUSBMSC(); // re-init USB MSC to show new file to user for editing
```
brentru, line 401: 1) Where did we detach USB MSC? 
2) Is this called twice? We shouldn't be.
*  Do we call this again, outside of this function, after this is called?

src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -367,7 +417,23 @@ void Wippersnapper_FS::parseSecrets() {
   // Attempt to deserialize the file's JSON document
   JsonDocument doc;
   DeserializationError error = deserializeJson(doc, secretsFile);
-  if (error) {
+  if (error == DeserializationError::EmptyInput)
+  {
+    if (WS.brownOutCausedReset)
```
brentru, line 422: A special case to handle the brownout detection shouldn't be located here. 

We should only handle the brownout detection/errors within the class constructor/init.

brentru (changes requested):

src/provisioning/tinyusb/Wippersnapper_FS.cpp
tyeth: We still check the bootlog contents in createBootfile, and print if there is a discrepency. In the event of a failure to write bootlog, assuming the secrets are still valid, we don't want to call it a failure by returning false

src/provisioning/tinyusb/Wippersnapper_FS.cpp
tyeth: We wanted to inform that the bootlog isn't write but we can't update it. I guess I could return true instead, or comment this as to why

tyeth:
Recreation Steps:
 Firstly if you want to keep the corruption you'll need to change the code to not call initialised filesystem the second time (with forceFormat true), also disabling the check inside the initFilesystem function on failure. The same for the file.open with bootlog+secrets. I think I totally disabled the macOS file writing + eraseCPFS.

Secondly pull this repository (and read the readme) https://github.com/tyeth/python_nordic_ppk2_wipper_secrets_brownout
The hardware setup and test process is described there, but in short it powers the device and waits for wifi related serial messages (successful boot), or corruption serial messages (**test stops**), it totally ignores anything else and continues (timing out that test run at X seconds).

brentru:
Note - I will have more bandwidth to reproduce and trace this error after November, we'll re-look into it then.

tyeth:
Future Tyeth, mention closes #621 and rejig a bit after #672

flavio-fernandes:
@tyeth: Sorry for bothering you about this, but is there any update on getting this PR merged into v2? I am running into the issue where my MagTag loses secrets.json when the battery dies.
Ref: https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/831

brentru:
@flavio-fernandes  @tyeth  We should be able to merge this into the current wippersnapper firmware, not v2

tyeth:
Brought up to date with `main` (display-v2 refactor, bootlog-version disambiguation, espressif32 platform bump, new sensor drivers) and merge conflicts resolved in #929. RP2040 `raspberypi_picow` builds fully; ESP32-S2 sources compile cleanly (only the local esptool packaging step fails on the dev box — CI will verify).

tyeth:
@flavio-fernandes try [this set of CI assets](https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/actions/runs/27428411322). I've done a brief test on a feather esp32s2 reverse tft, and main release v129 can be triggered to forget it's secrets with agressive resetting at 1-2s after boot (just before drive appears in windows). The updated PR #929 avoids this (bringing this PR inline with main).
So give the build-files asset a download from this page (summary page of any CI job) https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/actions/runs/27428411322
Here's the UF2 file, and/or combined.bin (flash at 0x0 with esptool)
[wippersnapper.magtag.1.0.0-adafruit-d2fbe0dd.zip](https://github.com/user-attachments/files/28929827/wippersnapper.magtag.1.0.0-adafruit-d2fbe0dd.zip)

I'll probably release it Monday anyway.

flavio-fernandes:
Cool beans! TYSM @tyeth and @brentru. I will definitely give it a try and let you know how that goes.

flavio-fernandes:
@tyeth :  I made my magtag's (D4) battery feed public, in case you want to follow: 

[flavio's magtag battery](https://io.adafruit.com/flaviof/feeds/magtag2423701.ws-013)

Screenshot 2026-06-15 at 12 37 39 PM

flavio-fernandes:
Update: ~~looking good @tyeth ! I drained the battery and was able to have Magtag going again by simply adding power to it. Woot!~~

@tyeth @brentru -- 2026-Jun-16-Tue@07:15:39EDT: Oh no, I celebrated too soon. I'm so sorry but this morning I came to a magtag that once again has forgotten its secrets. :( My thinking is that there must be another codepath that leads to the formatting of the filesystem. I'm so sorry for the bad news. Should we revert this commit?

Screenshot 2026-06-16 at 7 18 51 AM

brentru:
@flavio-fernandes  Could you explain, in detail, how you're running your tests, the hardware you are using, the battery (and its capacity), etc?

flavio-fernandes:
> @flavio-fernandes Could you explain, in detail, how you're running your tests, the hardware you are using, the battery (and its capacity), etc?

I have a MagTag running off of a battery:
- Adafruit MagTag - 2.9" Grayscale E-Ink WiFi Display PID: 4800
- Lithium Ion Polymer Battery - 3.7v 2500mAh PID: 328

I have it running a WipperSnapper build that @tyeth asked me to try: https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/656#issuecomment-470218243

I am using it with Adafuit IO, as you would expect. It is using the actions and feeds as described in this learning page:
[no-code-magtag-every-day-goal-tracker](https://learn.adafruit.com/no-code-magtag-every-day-goal-tracker/)

What I do not have to provide you is the serial output, as I have it on my fridge. Besides that, I never used it with a USB that did not charge the battery. And since the problem only occurs when the battery gets drained, what I will try doing next is a serial connection where the VCC is not connected.

Before I do that, is there anything in particular that I could do so the serial provides us with more useful information?
@brentru  Please let me know if this is useful.

tyeth:
We will do similar testing. 

In the past I used a USB isolator which allows you to turn off the power line (obviously ignoring parasitic powering by the data lines), which means the device can speak to the PC over serial still.
 And I then powered the device with a Nordic power profile kit two (nrf PPK II/2) through the battery lines which allows a dynamically adjustable power supply voltage along with repeated resetting.

We were waiting for you to come back and see what you offered, set up and information-wise, before continuing.

  I chucked a few thoughts into the AI cooking pot as to my suspicions about the boot reason not being reliable and a more suitable method of testing before reformatting flash, and came up with a a more bulletproof idea.
#931
It's completely untested with a little more work to ideally be done and we'll get to it soon enough, but it seems reasonable as it effectively checks the flash reads correctly a couple of bytes in sensible locations for known patterns and spi/flash errors.

brentru:
Once "Lithium Ion Polymer Battery - 3.7v 2500mAh" appears back in stock on the Adafruit Shop, I'll get an email notif. and can order this for bug reproduction.

flavio-fernandes:
Hello @brentru and @tyeth . Not my proudest hack, but I managed to sacrifice a USB cable so now I can collect logs from the serial as it wipes clean the filesystem when the battery is out. I've loaded my spare magtag with the latest rev and made it's battery feed (D4) available.

[flaviofMagtagTest Battery](https://io.adafruit.com/flaviof/feeds/magtag2283801.ws-013)

Screenshot 2026-06-24 at 10 44 10 PM

Screenshot 2026-06-24 at 10 41 12 PM

Screenshot 2026-06-24 at 10 40 39 PM

To be continued... ;)

flavio-fernandes:
Logs:
[2026-06-24 23-14-38 EPD MagTag 2_9_ ESP32_S2.log](https://github.com/user-attachments/files/29337635/2026-06-24.23-14-38.EPD.MagTag.2_9_.ESP32_S2.log)

flavio-fernandes:
## Serial Log Analysis — MagTag secrets wipe on beta.130

Hi @tyeth and @brentru — I wanted to contribute something concrete to move this forward. I reproduced the bug on beta.130 with a sacrificed non-charging USB cable (TX/RX only, no VBUS) so the MagTag ran purely on battery while logging to serial. Here's what the log proves and why I think #931 is the right path.

### What the log shows

The session ran from **23:14:38 EDT (Jun 24)** through the night. Two hours of clean operation: MQTT pings, sensor reads every 30s, display updates, everything normal.

Around **01:14 EDT (Jun 25)** the battery started struggling — brief disconnect/reconnect cycles appeared. At **01:17:03** the device disconnected while stuck in an active polling loop (`Polling for registration message response` × 36+ iterations) with the WiFi radio drawing peak current.

**The next connect at 01:19:56 — the very first message was:**
```
ERROR: Please edit the secrets.json file. Then, reset your board.
```

The filesystem was wiped in that **2 min 53 sec gap**. Battery died, recovered, device rebooted — and destroyed itself.

The rest of the log (through ~08:00) is a boot loop alternating between two messages:
- *"Please edit the secrets.json file"* — a fresh format just ran this boot
- *"Invalid IO credentials"* — this boot found the default-template `secrets.json` from a previous format

The battery was right at the edge: enough to boot and even run `f_mkfs`, never enough to reach WiFi.

---

### Root cause

The culprit is [Wippersnapper_FS.cpp line 111](https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/blob/main/src/provisioning/tinyusb/Wippersnapper_FS.cpp#L111):

```cpp
if (!initFilesystem() && !initFilesystem(true)) {
```

That `&&` means: "if the first mount fails, **immediately force-format**." No question asked. `initFilesystem(true)` calls `f_mkfs()`, erasing everything.

Full failure chain:
1. Battery sags during a WiFi TX burst (~250–400 mA peak) or a `syncBlocks()` call
2. A partially-written FAT structure is left on flash
3. ESP32 resets — **not** as `RTCWDT_BROWN_OUT_RESET` but typically watchdog or `SW_CPU_RESET` (fast voltage spikes don't trigger the brownout detector)
4. Boot: `initFilesystem()` can't mount the corrupted FAT → returns `false`
5. `initFilesystem(true)` immediately runs → wipes everything → `createSecretsFile()` writes the default template → halt

---

### Why the `brownOutCausedReset` guard in #929 doesn't cover this case

PR #929's protection is gated on `WS.brownOutCausedReset`, which is only `true` when the ESP32 reports `RTCWDT_BROWN_OUT_RESET`. A **fast voltage dip from a WiFi transmission** doesn't trigger that — it shows up as a watchdog or power-on reset. So `brownOutCausedReset == false`, the guard is bypassed, and the format runs anyway.

This is why I initially reported that #929 appeared to fix it: the first few battery deaths were probably slow drains that produced a real brownout reset code. Later ones were fast WiFi-induced spikes that didn't — and the wipe recurred.

---

### Why #931 covers it

#931's `flashSafeToFormat()` doesn't trust the reset reason at all. For ESP32 (MagTag uses internal flash via `Adafruit_FlashTransport_ESP32`), it calls `esp_flash_read()` to check two areas that a FAT format never touches:

1. **Partition table at `0x8000`** — must have ESP-IDF magic `0xAA 0x50`
2. **NVS page header at `0x9000`** — state field must be a valid NVS page state

If both pass → flash is readable, power is stable → safe to format.
If either fails → flash is degraded → **halt without formatting**, USB/LED kept off to save the dying battery.

| Scenario | beta.130 | #929 | #931 |
|---|---|---|---|
| Slow drain → brownout reset | wipes ❌ | protected ✅ | protected ✅ |
| Fast WiFi sag → watchdog reset | wipes ❌ | **wipes** ❌ | protected ✅ |
| Genuine first boot | formats ✅ | formats ✅ | formats ✅ |

---

I'm happy to keep testing builds as you work toward a release. Thanks for all the work you two have put into this — it's a subtle and genuinely tricky failure mode.

flavio-fernandes:
Quick pointer for anyone landing here: this work continues in **#931**, which is built on top of this branch (`update-656-merge-main` = this PR merged with main) and contains everything here plus a refinement — *gate a reformat on flash health, not the brownout reset reason*. Git-wise, #931 is a strict superset of #656.

The consolidated on-hardware findings and how the follow-ups layer up (#931 → #937 → #942) are written up here: #931 (comment).

Posting only to save the next reader some confusion — whether to keep this open is entirely @tyeth's call.

PR #417, opened 2023-03-03
Update SCD4x driver and sensor types

brentru:
Note: This PR is a draft until we figure out why SCD40 is writing CO2 as an integer to the feed but the firmware sends a float and the device page displays a float.

This pull request:
- Refactoring around SCD40 (now SCD4X) driver's `getEvent()` calls
  - Note: In the future we will want to check if the sensor is ready to be read, currently blocking via https://github.com/Sensirion/arduino-i2c-scd4x/commit/4ec07965f30f32f4320960a04aadb7a19e6499c7
- Updates existing I2C drivers to use CO2, eCO2 sensor types (https://github.com/adafruit/Adafruit_Sensor/pull/45)
- Updates existing I2C drivers to use particle sensor types (https://github.com/adafruit/Adafruit_Sensor/pull/46)
- Updates existing I2C drivers to use sensor types for gas resistance and unitless percentage (https://github.com/adafruit/Adafruit_Sensor/pull/50)

PR #658, opened 2024-11-08
Scd4x 3x issue

tyeth:
This adds a metro s3 debug target, and changes the SCD4x to use a single read for all datapoints at once like the SCD30.
Also noticed the alreadyRead condition needed a fix.
Lastly, to get a reliable first read on the SCD4x, I pass in the sensor period to the Begin setup call, allowing the driver to set it's first read to be five seconds time (after init).


tyeth:

src/components/i2c/WipperSnapper_I2C.cpp
```diff
@@ -1328,16 +1329,17 @@ void WipperSnapper_Component_I2C::sensorEventRead(
     unsigned long curTime, wippersnapper_signal_v1_I2CResponse *msgi2cResponse,
     bool (WipperSnapper_I2C_Driver::*getEventFunc)(sensors_event_t *),
     long (WipperSnapper_I2C_Driver::*getPeriodFunc)(),
-    long (WipperSnapper_I2C_Driver::*getPeriodPrvFunc)(),
-    void (WipperSnapper_I2C_Driver::*setPeriodPrvFunc)(long),
+    ulong (WipperSnapper_I2C_Driver::*getPeriodPrvFunc)(),
+    void (WipperSnapper_I2C_Driver::*setPeriodPrvFunc)(ulong),
```
brentru, line 1333: What's the reasoning for the function header to change from `long` to `ulong`?

src/components/i2c/WipperSnapper_I2C.cpp
```diff
@@ -1328,16 +1329,17 @@ void WipperSnapper_Component_I2C::sensorEventRead(
...
     wippersnapper_i2c_v1_SensorType sensorType, const char *sensorName,
     const char *unit, sensors_event_t event,
     float sensors_event_t::*valueMember, bool &sensorsReturningFalse,
     int &retries) {
   // sensorName used for prefix + error message, units is value suffix
   curTime = millis();
   if (((*iter)->*getPeriodFunc)() != 0L &&
-      curTime - ((*iter)->*getPeriodPrvFunc)() > ((*iter)->*getPeriodFunc)()) {
+      curTime - ((*iter)->*getPeriodPrvFunc)() >
+          (ulong)((*iter)->*getPeriodFunc)()) {
```
brentru, line 1342: I have the same question here

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD30.h
```diff
@@ -156,16 +158,16 @@ class WipperSnapper_I2C_Driver_SCD30 : public WipperSnapper_I2C_Driver {
...
     return true;
   }
 
 protected:
   Adafruit_SCD30 *_scd = nullptr; ///< SCD30 driver object
   ulong _lastRead = 0;            ///< Last time the sensor was read
-  sensors_event_t _temperature;   ///< Temperature
-  sensors_event_t _humidity;      ///< Relative Humidity
-  sensors_event_t _CO2;           ///< CO2
+  float _temperature;             ///< Temperature
```
brentru, line 168: Why are we moving away from `sensors_event_t` types here?

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD4X.h
```diff
@@ -47,46 +47,85 @@ class WipperSnapper_I2C_Driver_SCD4X : public WipperSnapper_I2C_Driver {
...
-      @returns  True if the measurements were read without errors, False
-                if read errors occured or if sensor did not have data ready.
+      @brief    Checks if the sensor is ready to be read
+      @returns  True if the sensor is ready, False otherwise.
   */
-  /********************************************************************************/
-  bool readSensorMeasurements() {
-    uint16_t error;
+  /*******************************************************************************/
+  bool sensorReady() {
```
brentru, line 93: Change function to something that reflects its returning a `bool`, like `IsSensorReady()`

brentru: the rest of this function incl. the fall-thru makes logical sense, it's much better than it was before

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD4X.h
```diff
@@ -47,46 +47,85 @@ class WipperSnapper_I2C_Driver_SCD4X : public WipperSnapper_I2C_Driver {
...
+
+  /*******************************************************************************/
+  /*!
+      @brief    Reads the sensor.
+      @returns  True if the sensor was read successfully, False otherwise.
+  */
+  /*******************************************************************************/
+  bool readSensorData() {
+    // dont read sensor more than once per second
+    if (alreadyRecentlyRead()) {
```
brentru, line 115: Okay so `alreadyRecentlyRead()` returning `true` assets that the sensor is read > 1x per second? Maybe change the function's name to reflect this?

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD4X.h
```diff
@@ -47,46 +47,85 @@ class WipperSnapper_I2C_Driver_SCD4X : public WipperSnapper_I2C_Driver {
...
+      return true;
+    }
 
-    // Check if data is ready
-    error = _scd->getDataReadyFlag(isDataReady);
-    if (error || !isDataReady)
+    if (!sensorReady()) {
```
brentru, line 119: Change to something that reflects the boolean, `IsSensorReady()`

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SCD4X.h
```diff
@@ -137,18 +178,20 @@ class WipperSnapper_I2C_Driver_SCD4X : public WipperSnapper_I2C_Driver {
...
 protected:
-  SensirionI2CScd4x *_scd; ///< SCD4x driver object
-  uint16_t _co2;           ///< SCD4x co2 reading
-  float _temperature;      ///< SCD4x temperature reading
-  float _humidity;         ///< SCD4x humidity reading
+  SensirionI2CScd4x *_scd = nullptr; ///< SCD4x driver object
+  uint16_t _co2 = 0;                 ///< SCD4x co2 reading
+  float _temperature = 20.0f;        ///< SCD4x temperature reading
+  float _humidity = 50.0f;           ///< SCD4x humidity reading
+  ulong _lastRead = 0;               ///< Last time the sensor was read
```
brentru, line 194: What's the reason for the default values being defined here, why do they need to be 0, 20.0, 50.0, 0?

If they do need to be pre-defined before the function runs, let's do it in the _constructor_ instead of here.

brentru (changes requested):
The changes to the driver(s) look good.

I am confused why you changed functions within `I2c.*` and `I2c_Driver.h`. Are these necessary for this PR, and if so, can you explain why they need to be changed?

src/components/i2c/WipperSnapper_I2C.cpp
tyeth: The previous poll period (periodPrv) is stored in the backing field as a long, and initially is set to 24hours ago (negative), but the logic check was shortcutting due to the unsigned long getter and setter methods (achieving the initial read change that introduced it). 
 Now that I've got a driver that wants update to only be called 5seconds after driver init/begin, which uses the sensor poll period taken away from millis +5secs, we need a way of the previous poll period getter+setter taking negative numbers.

brentru: This makes sense to me as `millis()` returns a `ulong` and you want to prevent it from underflowing.

Would we want to change _everything_ within `I2C.cpp/I2C_Driver` to `ulong` while we're at it? I want to make sure we have matching function signatures everywhere.

T-Mosher: Be careful that you don't make a change that breaks backward compatibility. Could code that uses the current data type break if the type is re-defined?

brentru:
@tyeth  We worked on this today! Want to bring in [ #417 ](https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/689) first, and then bring this drive in-line with #417?

tyeth:
Going to break out the changes to time into a separate PR.

SCD work is in https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/709 and [this branch](https://github.com/tyeth/Adafruit_Wippersnapper_Arduino/tree/scd30-scd4x-others-poll-min-1sec)

Polling work will be in https://github.com/tyeth/Adafruit_Wippersnapper_Arduino/tree/ulong-long-previousPollPeriod-start-scd40-in-5s

tyeth:
Happy to close it. I perused it again and can refer when necessary

PR #660, opened 2024-11-12
Added aht21 sensor to the drivers

thussaiththelaw:
the aht21 uses the same library as the aht20, but a different i2c address is used so this needed to be added so the the aht21 sensor can be used. I am working on this within the Wippersnapper components.


tyeth (approved):
Looks good

Issue #659, opened 2024-11-08
Pico (and other boards) connect to the first matching SSID, instead of picking the least negative RSSI value


tyeth:
I've got a mesh network with 3 nodes (plus base station router), all under the same SSID.
The pico detects one at -82db (over ten feet away) and fails to connect to it usually, instead only successfully connecting when it connects to the mesh with -32db (2feet away).

In an ideal world we would evaluate the entire network list from the scan / scan result, or maybe just order by RSSI first then reusing existing logic.

tyeth:
Watchdog timer gets in the way of wifi Multi class (it's not watchdog aware) for Pico. This doesn't present unless using alternative_networks, another spanner in the works. Watchdog cannot be disabled on pico once enabled, so the only option is to not enable it in the first place until the wifi has connected (small risk of device not rebooting in major error like failed to generate device uid, but not rebooting wont matter then). Assuming WDT is on and wifi fails then will attempt reconnect (and assuming over 8seconds will reset and try again without watchdog timer, or if connects then no issue)

brentru:
Resolved by #661

PR #661, opened 2024-11-13
Sorting of scanned wifi networks

tyeth:
Addresses #659

### DIFF
--- a/src/Wippersnapper.h
+++ b/src/Wippersnapper.h
@@ -214,8 +214,7 @@ typedef enum {
   FSM_NET_ESTABLISH_MQTT,
 } fsm_net_t;
 
-#define WS_WDT_TIMEOUT 60000       ///< WDT timeout
-#define WS_MAX_ALT_WIFI_NETWORKS 3 ///< Maximum number of alternative networks
+#define WS_WDT_TIMEOUT 60000 ///< WDT timeout
 /* MQTT Configuration */
 #define WS_KEEPALIVE_INTERVAL_MS                                               \
   5000 ///< Session keepalive interval time, in milliseconds
--- a/src/network_interfaces/Wippersnapper_AIRLIFT.h
+++ b/src/network_interfaces/Wippersnapper_AIRLIFT.h
@@ -103,8 +103,8 @@ public:
 
   /***********************************************************/
   /*!
-  @brief   Performs a scan of local WiFi networks.
-  @returns True if `_network_ssid` is found, False otherwise.
+    @brief   Performs a scan of local WiFi networks.
+    @returns True if `_network_ssid` is found, False otherwise.
   */
   /***********************************************************/
   bool check_valid_ssid() {
@@ -119,26 +119,33 @@ public:
       return false;
     }
 
-    // Was the network within secrets.json found?
-    for (int i = 0; i < n; ++i) {
-      if (strcmp(_ssid, WiFi.SSID(i)) == 0) {
-        WS_DEBUG_PRINT("SSID found! RSSI: ");
-        WS_DEBUG_PRINTLN(WiFi.RSSI(i));
-        return true;
+    bool foundNetwork = false;
+
+    WS_DEBUG_PRINTLN("WipperSnapper found these WiFi networks:");
+    for (uint8_t i = 0; i < n; i++) {
+      if (!foundNetwork && strcmp(WiFi.SSID(i), _ssid) == 0) {
+        foundNetwork = true;
       }
-    }
-
-    // User-set network not found, print scan results to serial console
-    WS_DEBUG_PRINTLN("ERROR: Your requested WiFi network was not found!");
-    WS_DEBUG_PRINTLN("WipperSnapper found these WiFi networks: ");
-    for (int i = 0; i < n; ++i) {
       WS_DEBUG_PRINT(WiFi.SSID(i));
-      WS_DEBUG_PRINT(" ");
+      WS_DEBUG_PRINT(" (");
+      uint8_t BSSID[WL_MAC_ADDR_LENGTH];
+      WiFi.BSSID(i, BSSID);
+      for (int m = 0; m < WL_MAC_ADDR_LENGTH; m++) {
+        if (m != 0)
+          WS_DEBUG_PRINT(":");
+        WS_DEBUG_PRINTHEX(BSSID[m]);
+      }
+      WS_DEBUG_PRINT(") ");
       WS_DEBUG_PRINT(WiFi.RSSI(i));
-      WS_DEBUG_PRINTLN("dB");
+      WS_DEBUG_PRINT("dB (ch");
+      WS_DEBUG_PRINT(WiFi.channel(i))
+      WS_DEBUG_PRINTLN(")");
     }
 
-    return false;
+    if (!foundNetwork) {
+      WS_DEBUG_PRINTLN("ERROR: Your requested WiFi network was not found!");
+    }
+    return foundNetwork;
   }
 
   /********************************************************/
@@ -323,6 +330,7 @@ protected:
 
       // validate co-processor's firmware version
       if (!firmwareCheck()) {
+        // TODO: see if there's a way to add to bootlog without usb reattach
         WS_DEBUG_PRINTLN("Please upgrade the firmware on the ESP module to the "
                          "latest version.");
       }
--- a/src/Wippersnapper_Networking.h
+++ b/src/Wippersnapper_Networking.h
@@ -16,6 +16,11 @@
 #ifndef WIPPERSNAPPER_NETWORKING_H
 #define WIPPERSNAPPER_NETWORKING_H
 
+#ifndef WL_MAC_ADDR_LENGTH
+#define WL_MAC_ADDR_LENGTH 6 ///< MAC address length - from RP2040 BSP
+#endif
+#define WS_MAX_ALT_WIFI_NETWORKS 3 ///< Maximum number of alternative networks
+
 #if defined(ADAFRUIT_METRO_M4_EXPRESS) ||                                      \
     defined(ADAFRUIT_METRO_M4_AIRLIFT_LITE) || defined(ADAFRUIT_PYPORTAL) ||   \
     defined(ADAFRUIT_PYPORTAL_M4_TITANO) || defined(USE_AIRLIFT)
--- a/src/network_interfaces/Wippersnapper_ESP32.h
+++ b/src/network_interfaces/Wippersnapper_ESP32.h
@@ -120,40 +120,40 @@ public:
       return false;
     }
 
-    // Was the network within secrets.json found?
-    for (int i = 0; i < n; ++i) {
-      if (strcmp(_ssid, WiFi.SSID(i).c_str()) == 0) {
-        WS_DEBUG_PRINT("SSID (");
-        WS_DEBUG_PRINT(_ssid);
-        WS_DEBUG_PRINT(") found! RSSI: ");
-        WS_DEBUG_PRINTLN(WiFi.RSSI(i));
-        return true;
-      }
-      if (WS._isWiFiMulti) {
+    bool foundNetwork = false;
+
+    WS_DEBUG_PRINTLN("WipperSnapper found these WiFi networks:");
+    for (uint8_t i = 0; i < n; i++) {
+      if (!foundNetwork && strcmp(WiFi.SSID(i).c_str(), _ssid) == 0) {
+        foundNetwork = true;
+      } else if (!foundNetwork && WS._isWiFiMulti) {
         // multi network mode
         for (int j = 0; j < WS_MAX_ALT_WIFI_NETWORKS; j++) {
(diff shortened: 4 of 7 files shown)


tyeth:
Works on Pico, which is what I was after as it was affecting my testing of other things.

```
---- Reopened serial port COM40 ----
Found single wifi network in secrets.json
Adafruit.io WipperSnapper
-------Device Information-------
Firmware Version: 1.0.0-beta.92
Board ID: rpi-pico-w
Adafruit.io User: tyeth
WiFi Network: free4all_2G
MAC Address: 28:CD:C1:0E:17:E8
-------------------------------
Generating device's MQTT topics...
Running Network FSM...
Establishing network connection...
Performing a WiFi scan for SSID...ERROR: Your requested WiFi network was not found!
WipperSnapper found these WiFi networks: 
free4all -50dB
free4all -52dB
free4all -55dB
Tenda_60C06C -79dB
free4all -82dB
ERROR [WDT RESET]: ERROR: Unable to find WiFi network, rebooting soon...
ERROR [WDT RESET]: ERROR: Unable to find WiFi network, rebooting soon...
```

tyeth:

src/network_interfaces/Wippersnapper_AIRLIFT.h
```diff
@@ -101,10 +103,34 @@ class Wippersnapper_AIRLIFT : public Wippersnapper {
...
   }
 
+  /****************************************************************/
+  /*!
+  @brief  a structure to hold network information for sorting
+  */
+  /****************************************************************/
+  struct WiFiNetwork {
+    char ssid[33]; /*!< SSID (Max 32 characters + null terminator */
+    int rssi;      /*!< Received Signal Strength Indicator */
```
brentru, line 113: RSSI should be type `int32_t` to [match the WiFi library](https://github.com/espressif/arduino-esp32/blob/master/libraries/WiFi/src/WiFiScan.h#L50)

src/network_interfaces/Wippersnapper_AIRLIFT.h
```diff
@@ -119,22 +145,38 @@ class Wippersnapper_AIRLIFT : public Wippersnapper {
       return false;
     }
 
-    // Was the network within secrets.json found?
+    // Dynamically allocate memory for the network list
+    std::vector<WiFiNetwork> networks(n);
```
brentru, line 149: I am not sure about using a vector/dynamic alloc. here, allowing too many possible networks will increase sort time.

Instead, could you define a global `MAX_WIFI_NETWORK` macro of 15 possible network options?

src/network_interfaces/Wippersnapper_AIRLIFT.h
```diff
@@ -101,10 +103,34 @@ class Wippersnapper_AIRLIFT : public Wippersnapper {
     _pass = WS._config.network.pass;
   }
 
+  /****************************************************************/
+  /*!
+  @brief  a structure to hold network information for sorting
+  */
+  /****************************************************************/
+  struct WiFiNetwork {
+    char ssid[33]; /*!< SSID (Max 32 characters + null terminator */
```
brentru, line 112: Looks like `SSID` is a string in the WiFi library, so we should match the type
https://github.com/espressif/arduino-esp32/blob/master/libraries/WiFi/src/WiFiScan.h#L48

src/network_interfaces/Wippersnapper_AIRLIFT.h
```diff
@@ -119,22 +145,38 @@ class Wippersnapper_AIRLIFT : public Wippersnapper {
...
+    // Dynamically allocate memory for the network list
+    std::vector<WiFiNetwork> networks(n);
+
+    // Store the scanned networks in the vector
     for (int i = 0; i < n; ++i) {
-      if (strcmp(_ssid, WiFi.SSID(i)) == 0) {
-        WS_DEBUG_PRINT("SSID found! RSSI: ");
-        WS_DEBUG_PRINTLN(WiFi.RSSI(i));
+      strncpy(networks[i].ssid, WiFi.SSID(i), sizeof(networks[i].ssid) - 1);
+      networks[i].ssid[sizeof(networks[i].ssid) - 1] =
```
brentru, line 154: If you switch SSID to a String type, it will be null-terminated during compilation.

src/network_interfaces/Wippersnapper_AIRLIFT.h
```diff
@@ -26,6 +26,8 @@
 #include "SPI.h"
 #include "WiFiNINA.h"
 #include "Wippersnapper.h"
+#include <algorithm>
```
brentru, line 29: Do we want this here, or moved to be included by WipperSnapper.h or another file (networking.h?) so it's included by all the drivers?

src/network_interfaces/Wippersnapper_AIRLIFT.h
```diff
@@ -119,22 +145,38 @@ class Wippersnapper_AIRLIFT : public Wippersnapper {
...
+      networks[i].ssid[sizeof(networks[i].ssid) - 1] =
+          '\0'; // Ensure null termination
+      networks[i].rssi = WiFi.RSSI(i);
+    }
+
+    // Sort the networks by RSSI in descending order
+    std::sort(networks.begin(), networks.end(), compareByRSSI);
+
+    // Was the network within secrets.json found?
+    for (const auto &network : networks) {
```
brentru, line 163: Why are we using `auto`type assignment here? Instead of this overhead, we could iterate until the end of the array

brentru (changes requested):
@tyeth  I have a few review notes below and also some questions:

1) Do we take the network with the highest RSSI?
2) If we fail to connect to the defined network, do we try again with the next network in our list of networks?
3) What is the purpose of sorting by RSSI if we are not going to take the network with the largest RSSI or if its not defined?

src/network_interfaces/Wippersnapper_AIRLIFT.h
tyeth: as part of the purge of dynamic allocation, this will not be stored as a string, also only applies to esp32, the remaining platforms are const char* returned from WiFi.SSID(int)

src/network_interfaces/Wippersnapper_AIRLIFT.h
tyeth: Agreed, resolved.

src/network_interfaces/Wippersnapper_AIRLIFT.h
tyeth: to keep the allocation not dynamic we'll skip string for these structs

tyeth:
Log of wifi multi connection attempt on ESP8266 with Wifi debug logging on (can't generate for WS firmware due to dbg size)
```
02:03:40.042 -> state: 2 -> 3 (0)
02:03:40.042 -> state: 3 -> 5 (10)
02:03:40.042 -> add 0
02:03:40.042 -> aid 8
02:03:40.042 -> cnt 
02:03:40.042 -> 
02:03:40.042 -> connected with free4all, channel 8
02:03:40.042 -> dhcp client start...
02:03:40.042 -> wifi evt: 0
02:03:42.073 -> ip:192.168.50.226,mask:255.255.255.0,gw:192.168.50.1
02:03:42.073 -> wifi evt: 3
02:03:42.141 -> [WIFIM] Connected:
02:03:42.141 -> [WIFIM]   SSID: free4all
02:03:42.141 -> [WIFIM]   IP:   192.168.50.226
02:03:42.141 -> [WIFIM]   MAC:  82:7F:54:A1:5B:41
02:03:42.141 -> [WIFIM]   CH:   8
02:03:42.141 -> [WIFIM]   RSSI: -71
02:03:42.141 -> WiFi connected: free4all 192.168.50.226
02:03:43.119 -> WiFi connected: free4all 192.168.50.226
02:03:44.136 -> WiFi connected: free4all 192.168.50.226
02:03:45.147 -> WiFi connected: free4all 192.168.50.226
02:03:46.141 -> WiFi connected: free4all 192.168.50.226
02:03:47.149 -> WiFi connected: free4all 192.168.50.226
02:03:48.123 -> WiFi connected: free4all 192.168.50.226
```

src/Wippersnapper.h
```diff
@@ -215,6 +216,7 @@ typedef enum {
 } fsm_net_t;
 
 #define WS_WDT_TIMEOUT 60000       ///< WDT timeout
+#define WS_MAX_SORTED_NETWORKS 15  ///< Maximum number of networks to sort
 #define WS_MAX_ALT_WIFI_NETWORKS 3 ///< Maximum number of alternative networks
```
brentru, line 220: Should this be moved into `https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/blob/main/src/Wippersnapper_Networking.h` instead, since it'd get #define'd before the network iface?

src/network_interfaces/Wippersnapper_AIRLIFT.h
```diff
@@ -101,10 +101,34 @@ class Wippersnapper_AIRLIFT : public Wippersnapper {
...
+  /*!
+  @brief  Comparison function to sort by RSSI in descending order
+  @param  a
+          WiFiNetwork object
+  @param  b
+          WiFiNetwork object
+  @returns True if a.rssi > b.rssi
+  */
+  /*******************************************************************/
+  bool static compareByRSSI(const WiFiNetwork &a, const WiFiNetwork &b) {
```
brentru, line 124: Instead of passing a ptr to a `WiFiNetwork` object, pass the `rssi` values directly into this function from the caller

src/network_interfaces/Wippersnapper_AIRLIFT.h
```diff
@@ -119,22 +143,55 @@ class Wippersnapper_AIRLIFT : public Wippersnapper {
       return false;
     }
 
-    // Was the network within secrets.json found?
+    WiFiNetwork networks[WS_MAX_SORTED_NETWORKS];
+    uint8_t numSavedNetworks = 0;
+
+    // Store the scanned networks in the array
     for (int i = 0; i < n; ++i) {
```
brentru, line 150: I would advise against using the if/else to catch the case where there are too many networks. If you want to truncate on `WS_MAX_SORTED_NETWORKS,` add that to the for() loop's conditional.

src/network_interfaces/Wippersnapper_AIRLIFT.h
```diff
@@ -119,22 +143,55 @@ class Wippersnapper_AIRLIFT : public Wippersnapper {
...
-      if (strcmp(_ssid, WiFi.SSID(i)) == 0) {
-        WS_DEBUG_PRINT("SSID found! RSSI: ");
-        WS_DEBUG_PRINTLN(WiFi.RSSI(i));
+      if (i < WS_MAX_SORTED_NETWORKS) {
+        strncpy(networks[i].ssid, WiFi.SSID(i), sizeof(networks[i].ssid));
+        networks[i].ssid[sizeof(networks[i].ssid) - 1] = '\0';
+        networks[i].rssi = WiFi.RSSI(i);
+        numSavedNetworks++;
```
brentru, line 155: Change to: `++numSavedNetworks`

src/network_interfaces/Wippersnapper_AIRLIFT.h
```diff
@@ -119,22 +143,55 @@ class Wippersnapper_AIRLIFT : public Wippersnapper {
...
+        WS_DEBUG_PRINT(") Ignoring ");
+        WS_DEBUG_PRINT(WiFi.SSID(i));
+        WS_DEBUG_PRINT("(");
+        WS_DEBUG_PRINT(WiFi.RSSI(i));
+        WS_DEBUG_PRINTLN(")");
+      }
+    }
+
+    // Sort the networks by RSSI in descending order
+    std::sort(networks, networks + numSavedNetworks, compareByRSSI);
```
brentru, line 168: If `compareByRSSI` is doing a direct comparison, you could use the `<` comparison from this method rather than create a new one. It might be cleaner not to refactor that one out.

brentru (changes requested):

src/network_interfaces/Wippersnapper_AIRLIFT.h
tyeth: @brentru Would you mind spelling that out for me? Do you mean there's a way to pass the RSSI ints from the network structs directly to the potentially refactored CompareByRSSI, and if so then there's a way to do the simple < here and then need no comparator method?
Alternatively do you mean just chuck a lamda doing the direct access to the network structs and comparing the rssi elements here instead of having the compare method, like
```
std::sort(networks, networks + numSavedNetworks, [](const WiFiNetwork &a, const WiFiNetwork &b) {
    return a.rssi > b.rssi;
});
```

brentru: I'm no longer sure, so you may want to just keep this as-is for now.

brentru (approved):

brentru:
@tyeth  LGTM, merge when tests are passing

PR #664, opened 2024-11-28
Pico2w

tyeth:
This depends on the wifi sorting (I can't easily connect / test otherwise), so hence the reason for the extra initial commits


brentru (approved):
@tyeth  Approved, LGTM!

PR #667, opened 2024-12-09
Fix USB-MSC not re-attaching

brentru:
This pull request resolves the following bug: After writing a default `secrets.json` file to the filesystem, the board must be manually reset. 

This pull request re-initializes USB-MSC after writing to `secrets.json`. It also adds improvements for code clarity to the `createSecretsFile()` function.


src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -327,19 +327,17 @@ void Wippersnapper_FS::createSecretsFile() {
...
-
-  // Flush and close file
   secretsFile.flush();
   secretsFile.close();
-  delay(2500);
 
-  // Signal to user that action must be taken (edit secrets.json)
+  // Re-attach the USB device for file access
+  delay(500);
+  initUSBMSC();
```
tyeth, line 339: Ah I forgot, locally I have this updated to come just one instruction later, after the writeToBootOut

tyeth (changes requested):
Sorry, my bad, I forgot that locally I've changed that tinyusb reattach to come just after the `writeToBootOut` instruction

PR #668, opened 2024-12-12
Stop blinding status lights

tyeth:
The Neopixel brightness has been ignored for a while, and I've been meaning to take a look, but I'm only reminded in serious testing phases. This resolves the issue (only tested with tinyusb devices, but littlefs will be identical)

### DIFF
--- a/src/provisioning/littlefs/WipperSnapper_LittleFS.cpp
+++ b/src/provisioning/littlefs/WipperSnapper_LittleFS.cpp
@@ -138,14 +138,11 @@ void WipperSnapper_LittleFS::parseSecrets() {
            "credentials!\n");
   }
 
-  if (doc["status_pixel_brightness"]) {
-    // check it casts to float and support user specifying 0.0f which is
-    // default, by using the |operator instead of .as
-    // https://arduinojson.org/v7/api/jsonvariant/or/
-    if ((doc["status_pixel_brightness"] | -1.0f) != -1.0f) {
-      WS.status_pixel_brightness = doc["status_pixel_brightness"].as<float>();
-    }
-  }
+  // specify type of value for json key, by using the |operator to include
+  // a typed default value equivalent of with .as<float> w/ default value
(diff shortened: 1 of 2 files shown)


src/provisioning/littlefs/WipperSnapper_LittleFS.cpp
```diff
@@ -138,6 +138,15 @@ void WipperSnapper_LittleFS::parseSecrets() {
            "credentials!\n");
   }
 
+  if (doc["status_pixel_brightness"]) {
+    // check it casts to float and support user specifying 0.0f which is
+    // default, by using the |operator instead of .as
+    // https://arduinojson.org/v7/api/jsonvariant/or/
+    if ((doc["status_pixel_brightness"] | -1.0f) != -1.0f) {
```
brentru, line 145: Set to match the default, `https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/blob/main/src/provisioning/tinyusb/Wippersnapper_FS.cpp#L328` so `| 0.2f`

src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -452,6 +452,14 @@ void Wippersnapper_FS::parseSecrets() {
            "credentials!");
   }
 
+  if (doc["status_pixel_brightness"]){
+    // check it casts to float and support user specifying 0.0f which is default,
+    // by using the |operator instead of .as https://arduinojson.org/v7/api/jsonvariant/or/
+    if ((doc["status_pixel_brightness"] | -1.0f) != -1.0f) {
```
brentru, line 458: Set to match the default, `https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/blob/main/src/provisioning/tinyusb/Wippersnapper_FS.cpp#L328` so `| 0.2f`

brentru (changes requested):

PR #670, opened 2024-12-17
RP2040 WDT=8.3secs - enable after wifi connect

tyeth:
Split out from wifi sorting pr, needed ideally (works without but wifi multi times out at 8seconds) before release of firmware version for #664

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -471,6 +471,12 @@ upload_port = /dev/cu.usbmodem1201
 [env:raspberypi_picow]
 extends = common:rp2040
 
+[env:raspberypi_picow_debug_port_only]
+extends = common:rp2040
+build_flags = 
+    -DUSE_TINYUSB
+    -DDEBUG_RP2040_PORT=Serial
+
 [env:raspberypi_picow_debug]
 extends = common:rp2040
 platform = https://github.com/maxgerhardt/platform-raspberrypi.git
--- a/src/Wippersnapper.cpp
+++ b/src/Wippersnapper.cpp
@@ -2748,8 +2748,10 @@ void Wippersnapper::connect() {
   // Dump device info to the serial monitor
   printDeviceInfo();
 
-  // enable global WDT
+// enable global WDT
+#ifndef ARDUINO_ARCH_RP2040
   WS.enableWDT(WS_WDT_TIMEOUT);
+#endif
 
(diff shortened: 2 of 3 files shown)


platformio.ini
```diff
@@ -471,6 +471,12 @@ upload_port = /dev/cu.usbmodem1201
 [env:raspberypi_picow]
 extends = common:rp2040
 
+[env:raspberypi_picow_debug_port_only]
+extends = common:rp2040
+build_flags = 
+    -DUSE_TINYUSB
+    -DDEBUG_RP2040_PORT=Serial
```
brentru, line 478: Comment out debugging flags using `;`

src/Wippersnapper.cpp
```diff
@@ -2748,8 +2748,10 @@ void Wippersnapper::connect() {
   // Dump device info to the serial monitor
   printDeviceInfo();
 
-  // enable global WDT
+// enable global WDT
+#ifndef ARDUINO_ARCH_RP2040
   WS.enableWDT(WS_WDT_TIMEOUT);
```
brentru, line 2753: Consider moving this guard into the `enableWDT()` call, so we aren't `ifndef''ing each time we call `enableWDT()`

src/Wippersnapper.h
```diff
@@ -214,7 +214,12 @@ typedef enum {
   FSM_NET_ESTABLISH_MQTT,
 } fsm_net_t;
 
-#define WS_WDT_TIMEOUT 60000       ///< WDT timeout
+#ifdef ARDUINO_ARCH_RP2040
+#define WS_WDT_TIMEOUT 8388 ///< Pico Max WDT timeout
```
brentru, line 218: Verbiage Change: `Pico` -> `RP2040` since Pico is the board not the chip which has this issue, and the newer Pico2 or future Picos may solve it.

brentru (changes requested):
Small feedback, looks good. Please ping me when addressed.

platformio.ini
tyeth: I'm keeping this one as a debug output build so the flag can remain

tyeth:
@brentru this one too please (switched enableWDT to after wifi connect)

brentru (approved):
@tyeth  LGTM

PR #671, opened 2024-12-17
DO NOT MERGE - Flush after debug prints, except SAMD (flush routine too big for SAMD21)

tyeth:
### DO NOT MERGE
Retest esp8266 and confirm SAMD51 prints device info


brentru:
@tyeth  Is this stale?

tyeth:
This still has some utility, there is a missing PRINT HEX define when debug is off, and the flushes maybe useful, but now the SAMD block would need removing. Not sure if we want this or not.

brentru:
I think we can close for now, re-open down the line if we need extra debug help for SAMD21.

PR #676, opened 2024-12-27
Review PR: Add "Offline" SDLogger Mode

brentru:
This pull request is not meant to be merged in. It is a review-only pull request for specific regions of code.

### Codebase regions for review
* Added a new class, `ws_sdcard` - for interfacing with a micro SD Card breakout.
  * Request: please take a careful look at how `parseConfigFile` functions and looks
  * Request: please take a careful look at how we switch between production and debug-mode in this class
  * `BuildJSONDoc` and `LogJSONDoc` API interfaces are present for each type of component
  * This class also supports a Real-Time-Clock breakout.
* API calls for interacting with `ws_sdcard` and mode-switching (determining whether the device is online or offline) within `WipperSnapper_V2` class in two functions -`provision()` and `connect()` 
* Added API calls for parsing the SD CS pins from the JSON file on the WIPPER drive within classes - `FS_V2` and `LittleFS_V2` 
* Refactored `network_interfaces/` to `adapters/` to allow for a WipperSnapper device without a network interface (`offline/`).
  * Added new class, `adapters/offline/ws_offline_pico` to support a Raspberry Pi in offline mode.

### Workflow enhancements for review:
* Added a new GitHub Actions workflow file, `run-tests.yml`, to automate testing using Wokwi.


tests/diagrams/offline.json
```diff
@@ -0,0 +1,160 @@
...
+      },
+      { "type": "wokwi-potentiometer", "id": "pot1", "top": -87.7, "left": 143.8, "attrs": {} },
+      { "type": "wokwi-potentiometer", "id": "pot2", "top": -1.3, "left": 143.8, "attrs": {} },
+      {
+        "type": "wokwi-pushbutton",
+        "id": "btn1",
+        "top": 9.2,
+        "left": 74,
+        "rotate": 270,
+        "attrs": { "color": "green" }
```
tyeth, line 39: Should we be using bounce:0 here too? (Like ../diagram.json)

examples/Wippersnapper_demo/Wippersnapper_demo.ino
```diff
@@ -24,12 +24,12 @@ void setup() {
   wipper.provision();
 
   Serial.begin(115200);
-  //while (!Serial) delay(10);
+  while (!Serial) delay(10);
```
tyeth, line 27: Think this should be commented out again

examples/Wippersnapper_demo/Wippersnapper_demo.ino
```diff
@@ -24,12 +24,12 @@ void setup() {
...
 
   wipper.connect();
 
 }
 
 void loop() {
-  wipper.run();
+  //wipper.run();
```
tyeth, line 34: Think this should be re-enabled

generate_json_checksum.py
```diff
@@ -0,0 +1,27 @@
+# generates the 8-bit checksum from a JSON file, provided its path is given as an argument
+import sys
+import json
+
+# load json file from provided path
+with open(sys.argv[1], 'r') as f:
```
tyeth, line 6: Might be good to trap errors here (like no args and missing file), but looks like it's only for our usage anyway
```
C:\dev\arduino\Adafruit_Wippersnapper_Arduino [review-mode-sdlogger ≡ +1 ~1 -0 !]> python .\generate_json_checksum.py 
Traceback (most recent call last):
  File "C:\dev\arduino\Adafruit_Wippersnapper_Arduino\generate_json_checksum.py", line 6, in <module>
    with open(sys.argv[1], 'r') as f:
              ~~~~~~~~^^^
IndexError: list index out of range
C:\dev\arduino\Adafruit_Wippersnapper_Arduino [review-mode-sdlogger ≡ +1 ~1 -0 !]> python .\generate_json_checksum.py a.json
Traceback (most recent call last):
  File "C:\dev\arduino\Adafruit_Wippersnapper_Arduino\generate_json_checksum.py", line 6, in <module>
    with open(sys.argv[1], 'r') as f:
         ^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'a.json'
```

platformio.ini
```diff
@@ -14,12 +14,13 @@ default_envs = adafruit_feather_esp32s3_tft, adafruit_magtag29_esp32s2, adafruit
 [env]
 framework = arduino
 monitor_speed = 115200
+extra_scripts = upload_no_build.py
```
tyeth, line 17: Not intended to be here I believe

examples/config-file/offline-config.json
```diff
@@ -0,0 +1,62 @@
...
+        "mode": "DIGITAL",
+        "sampleMode": "EVENT",
+        "direction": "INPUT",
+        "period": 5,
+        "pull": "UP",
+        "isPin": true
+      },
+      {
+        "componentAPI": "ds18x20",
+        "name": "DS18B20: Temperature Sensor (°F)",
```
tyeth, line 53: This degree symbol is \u encoded in another config (src/config.json), not sure it matters, but was curious how the two ended up differing (were both from exports or one hand coded maybe)

src/components/analogIO/controller.cpp
```diff
@@ -0,0 +1,290 @@
...
+            _analogio_model->GetAnalogIOEvent())) {
+      WS_DEBUG_PRINTLN(
+          "ERROR: Unable to publish analogio voltage event message, "
+          "moving onto the next pin!");
+      return false;
+    }
+    WS_DEBUG_PRINTLN("Published AnalogIOEvent message to broker!")
+  } else {
+
+    // TODO: Log out this data by calling a logging function in sdcard class
```
tyeth, line 201: I'm assuming we're not doing general logging and only datapoints are logged to SD, is that correct?

src/provisioning/tinyusb/Wippersnapper_FS_V2.cpp
```diff
@@ -0,0 +1,737 @@
...
+    @returns  FR_OK if filesystem formatted correctly,
+                otherwise any FRESULT enum.
+*/
+/**************************************************************************/
+FRESULT format_fs_fat12(void) {
+  FATFS elmchamFatfs_v2;
+  uint8_t workbuf_v2[4096];
+
+  // make filesystem
+  FRESULT r = f_mkfs("", FM_FAT | FM_SFD, 0, workbuf_v2, sizeof(workbuf_v2));
```
tyeth, line 72: This result could be checked too

src/provisioning/tinyusb/Wippersnapper_FS_V2.cpp
```diff
@@ -0,0 +1,737 @@
...
+
+    // Now that we have a formatted filesystem, we need to inititalize it
+    if (!wipperFatFs_v2.begin(&flash_v2)) {
+      setStatusLEDColor(RED);
+      fsHalt("FATAL ERROR: Failed to mount newly created filesystem!");
+    }
+  }
+
+  // Write contents to the formatted filesystem
+  if (!writeFSContents()) {
```
tyeth, line 129: I don't think this should be called unless !is_fs_formatted so we avoid writing to the file system unnecessarily. Alternatively the comments throughout the method mention it being a fresh file system which won't be the case most of the time with the current code.
Alternatively the brownout PR can resolve that issue of unnecessarily replacing `bootout.txt` each boot and we can fix this up later.

src/provisioning/sdcard/ws_sdcard.h
```diff
@@ -0,0 +1,117 @@
...
+#else
+#define SPI_SD_CLOCK SD_SCK_MHZ(50) ///< Default SPI clock speed
+#endif
+
+#define SD_FAT_TYPE 3           ///< SD type (3 = FAT16/FAT32 and exFAT)
+#define PIN_SD_CS_ERROR 255     ///< Error code for invalid SD card CS pin
+#define UNKNOWN_VALUE "unknown" ///< Default unknown JSON field value
+#define MAX_SZ_LOG_FILE (512 * 1024 * 1024) ///< Maximum log file size, in Bytes
+#define MAX_LEN_CFG_JSON                                                       \
+  4096 ///< Maximum length of the configuration JSON file, in Bytes
```
tyeth, line 37: This will probably need to be bigger / doubled, the current offline-config is 1.5k with 4 components

src/Wippersnapper_V2.h
```diff
@@ -0,0 +1,292 @@
...
+  CheckinModel *CheckInModel; ///< Instance of CheckinModel class
+  SensorModel *sensorModel;   ///< Instance of SensorModel class
+  DigitalIOController
+      *digital_io_controller; ///< Instance of DigitalIO controller class
+  AnalogIOController *analogio_controller; ///< Instance of AnalogIO controller
+  DS18X20Controller *_ds18x20_controller;  ///< Instance of DS18X20 controller
+
+  // TODO: does this really need to be global?
+  uint8_t _macAddrV2[6];  /*!< Unique network iface identifier */
+  char sUIDV2[13];        /*!< Unique network iface identifier */
```
tyeth, line 233: dupe docstring, should be device not network iface (see [Wippersnapper_V2.cpp:L568](https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/blob/651f5275ef9ec1f467a30fc6c6f628fb68f659f9/src/Wippersnapper_V2.cpp#L568))

src/provisioning/sdcard/ws_sdcard.cpp
```diff
@@ -0,0 +1,1008 @@
...
+  // Attempt to create the new log file
+  if (!file.open(_log_filename, O_RDWR | O_CREAT | O_AT_END))
+    return false;
+  WS_DEBUG_PRINT("[SD] Created new log file on SD card: ");
+  WS_DEBUG_PRINTLN(_log_filename);
+  _sd_cur_log_files++;
+  return true;
+}
+
+bool ws_sdcard::ValidateChecksum(JsonDocument &doc) {
```
tyeth, line 477: Docstring

src/provisioning/sdcard/ws_sdcard.cpp
```diff
@@ -0,0 +1,1008 @@
...
+    return "UNKNOWN";
+  }
+}
+
+void ws_sdcard::BuildJSONDoc(JsonDocument &doc, uint8_t pin, float value,
+                             wippersnapper_sensor_SensorType read_type) {
+  doc["timestamp"] = GetTimestamp();
+  doc["pin"] = "A" + String(pin);
+  doc["value"] = value;
+  doc["si_unit"] = SensorTypeToString(read_type);
```
tyeth, line 764: Feels like this isn't the SI unit, but more the sensor type. Would this be translated to units when converting to spreadsheet?)

src/provisioning/sdcard/ws_sdcard.cpp
```diff
@@ -0,0 +1,1008 @@
...
+  bufferedFile.print("\n"); // JSONL format specifier
+  bufferedFile.flush();     // Send the remaining bytes
+  // print the doc to the serial
+  serializeJson(doc, Serial);
+  Serial.print("\n");
+#else
+  szJson = serializeJson(doc, Serial);
+  Serial.print("\n"); // JSONL format specifier
+#endif
+  _sz_cur_log_file = szJson + 2; // +2 bytes for "\n"
```
tyeth, line 806: Wouldn't it be +1 byte for \n

src/provisioning/sdcard/ws_sdcard.cpp
```diff
@@ -0,0 +1,1008 @@
...
+  if (!LogJSONDoc(doc))
+    return false;
+  return true;
+}
+
+/**************************************************************************/
+/*!
+    @brief  Logs a GPIO sensor event to the SD card.
+    @param  pin
+            The GPIO pin number.
```
tyeth, line 886: DS18b20 not GPIO sensor

src/provisioning/sdcard/ws_sdcard.h
```diff
@@ -0,0 +1,117 @@
...
+                                       ///< card
+  bool is_mode_offline;                ///< True if offline mode is enabled, False otherwise
+  String _serialInput;                 ///< Serial input buffer
+  const char *json_test_data;          ///< Json test data
+  const char *_log_filename;           ///< Path to the log file
+  RTC_DS3231 *_rtc_ds3231 = nullptr;   ///< DS3231 RTC object
+  RTC_DS1307 *_rtc_ds1307 = nullptr;   ///< DS1307 RTC object
+  RTC_PCF8523 *_rtc_pcf8523 = nullptr; ///< PCF8523 RTC object
+  RTC_Millis *_rtc_soft = nullptr;     ///< Software RTC object
+  bool _use_test_data;                 ///< True if sample data is being used for testing
```
tyeth, line 114: This is never used

tests/diagram.json
```diff
@@ -0,0 +1,160 @@
...
+    { "type": "wokwi-potentiometer", "id": "pot2", "top": -1.3, "left": 143.8, "attrs": {} },
+    {
+      "type": "wokwi-pushbutton",
+      "id": "btn1",
+      "top": 9.2,
+      "left": 74,
+      "rotate": 270,
+      "attrs": { "bounce": "0", "color": "green" }
+    },
+    { "type": "board-ds18b20", "id": "temp1", "top": -87.53, "left": 272.88, "attrs": {} },
```
tyeth, line 41: Might be helpful to set the different temps/ids per sensor, something like this for the attrs:
```
"attrs": {
      "deviceID": "010203040506",
      "familyCode": "16",
      "temperature": "-1",
      "minTemp": "-20",
      "maxTemp": "30",
      "tempWaveForm": "sine",
      "tempWaveFreq": "30"
    }
```

See here for details: https://github.com/bonnyr/wokwi-ds1820-custom-chip?tab=readme-ov-file#attributes

src/Wippersnapper_V2.cpp
```diff
@@ -0,0 +1,1314 @@
...
+  WsV2._fileSystemV2->parseDisplayConfig(config);
+  WsV2._display = new ws_display_driver(config);
+  // Begin display
+  if (!WsV2._display->begin()) {
+    WS_DEBUG_PRINTLN("Unable to enable display driver and LVGL");
+    haltErrorV2("Unable to enable display driver, please check the json "
+                "configuration!");
+  }
+
+  WsV2._display->enableLogging();
```
tyeth, line 129: appears to be a no-op

src/Wippersnapper_V2.cpp
```diff
@@ -0,0 +1,1314 @@
...
+  WsV2._ui_helper->show_scr_load();
+  WsV2._ui_helper->set_label_status("Validating Credentials...");
+#endif
+
+#ifdef USE_TINYUSB
+  _fileSystemV2->parseSecrets();
+#elif defined(USE_LITTLEFS)
+  _littleFSV2->parseSecrets();
+#else
+  check_valid_ssidV2(); // non-fs-backed, sets global credentials within network
```
tyeth, line 143: This doesn't set the credentials, instead it scans for wifi networks and verifies the match against already existing network config

src/Wippersnapper_V2.cpp
```diff
@@ -0,0 +1,1314 @@
...
+    // TDOO: This can be cleaned up, get the buffer
+    const std::vector<uint8_t> &buffer = WsV2._sharedConfigBuffers[i];
+    WS_DEBUG_PRINTLN("Creating istream...");
+    pb_istream_t istream = pb_istream_from_buffer(buffer.data(), buffer.size());
+    // Decode the message
+    WS_DEBUG_PRINT("Decoding BrokerToDevice message #");
+    WS_DEBUG_PRINTLN(i);
+    if (!pb_decode(&istream, wippersnapper_signal_BrokerToDevice_fields,
+                   &msg_signal)) {
+      WS_DEBUG_PRINTLN("ERROR: Unable to decode BrokerToDevice message!");
```
tyeth, line 468: useful to know which of the messages failed (i or i+1)

src/Wippersnapper_V2.cpp
```diff
@@ -0,0 +1,1314 @@
...
+                Adafruit IO MQTT /error special topic.
+    @param    errorData
+                Data from MQTT broker.
+    @param    len
+                Length of data received from MQTT broker.
+*/
+/**************************************************************************/
+void cbErrorTopicV2(char *errorData, uint16_t len) {
+  (void)len; // marking unused parameter to avoid compiler warning
+  WS_DEBUG_PRINT("IO Ban Error: ");
```
tyeth, line 487: I was thinking about the device looking for known error strings (mqtt topic doesn't exist if component removed but device about to send periodic data) with it's device uuid in the message (if different uuid then ignore). If not known message then assume ban/backoff still. I think I need to test the ban situation from a second device to confirm what messages we may see.

src/Wippersnapper_V2.cpp
```diff
@@ -0,0 +1,1314 @@
...
+  // Generate device unique identifier
+  // Set machine_name
+  WsV2._boardIdV2 = BOARD_ID;
+  // Move the top 3 bytes from the UID
+  for (int i = 5; i > 2; i--) {
+    WsV2._macAddrV2[6 - 1 - i] = WsV2._macAddrV2[i];
+  }
+  snprintf(WsV2.sUIDV2, sizeof(WsV2.sUIDV2), "%02d%02d%02d", WsV2._macAddrV2[0],
+           WsV2._macAddrV2[1], WsV2._macAddrV2[2]);
+  // Conversion to match integer UID sent by createMsgCheckinRequest()
```
tyeth, line 577: What's the idea here, to drop leading zeros, or is there some potential translation error I'm missing?

tyeth: Another thought just occurred to me, isn't it possible to have two boards with the same first 3 bytes of the mac address? Wouldn't that cause a big issue as the device ids are assumed to be unique?

src/Wippersnapper_V2.cpp
```diff
@@ -0,0 +1,1314 @@
...
+  WS_DEBUG_PRINT("Broker-to-device topic: ");
+  WS_DEBUG_PRINTLN(WsV2._topicB2d);
+  // Subscribe to broker-to-device topic
+  _subscribeB2d = new Adafruit_MQTT_Subscribe(WsV2._mqttV2, WsV2._topicB2d, 1);
+  WsV2._mqttV2->subscribe(_subscribeB2d);
+  _subscribeB2d->setCallback(cbBrokerToDevice);
+
+  // Create global device to broker topic
+  // Attempt to allocate memory for the broker-to-device topic
+#ifdef USE_PSRAM
```
tyeth, line 656: I kind of want to kill these off in favour of our own ws_malloc which does the same switch in one place

src/Wippersnapper_V2.cpp
```diff
@@ -0,0 +1,1314 @@
...
+}
+
+/**************************************************************************/
+/*!
+    @brief    Polls for and handles the checkin response
+              message from the broker.
+*/
+/**************************************************************************/
+void Wippersnapper_V2::PollCheckinResponse() {
+  WsV2.got_checkin_response = false;
```
tyeth, line 1060: I wondered about having this false be set after the publish, on line 1049, but not really much difference

src/Wippersnapper_V2.cpp
```diff
@@ -0,0 +1,1314 @@
...
+    WS_DEBUG_PRINTLN(
+        "[Offline] Hardware configured, skipping network setup and running app...");
+    // Set the status LED to green to indicate successful configuration
+    setStatusLEDColor(0x00A300, WS.status_pixel_brightness);
+    delay(100);
+    // Set the status LED to off during app runtime
+    setStatusLEDColor(0x000000, WS.status_pixel_brightness);
+    return;
+  } else {
+    WS_DEBUG_PRINTLN("Running in online mode...");
```
tyeth, line 1233: Curious about how you feel on this, purely around clarity. The else can be dropped and message moved down, but I was curious about whether you felt it clearer having the else. It could be, as in my mind it forces one to double check the conditional

tests/diagrams/offline.json
brentru: Yes, it should reflect `diagram.json` exactly.

Resolving..

generate_json_checksum.py
brentru: Trapping `FileNotFoundError`.

upload_no_build.py
```diff
@@ -0,0 +1,8 @@
+Import("env")
+env.AddCustomTarget(
+    "uploadnobuild", 
+    None, 
+    'pio run -e %s -t nobuild -t upload' %
+        env["PIOENV"], 
+    title="Uploads without building"
+)
```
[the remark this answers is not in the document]
brentru: Not entirely sure why this artifact was still here, removed for now.

src/provisioning/tinyusb/Wippersnapper_FS_V2.cpp
brentru: We are checking the `FRESULT` value returned by this function, `format_fs_fat12()` on L114:
```
  // If we are not formatted, attempt to format the filesystem as fat12
  if (!is_fs_formatted) {
    FRESULT rc = format_fs_fat12();

    if (rc != FR_OK) {
```

Is there something I'm missing here?

src/provisioning/tinyusb/Wippersnapper_FS_V2.cpp
brentru: This work started at the same time (or before) the brownout PR was submitted. We'll need to merge and include the brownout PR in API-V2 rather than rewriting this.

src/provisioning/sdcard/ws_sdcard.h
brentru: Are you talking about `MAX_LEN_CFG_JSON`?

src/Wippersnapper_V2.h
brentru: Good catch!

src/Wippersnapper_V2.cpp
brentru: The `display` ptr?

src/Wippersnapper_V2.cpp
brentru: I think this may be out of scope for the sdlogger review as it's something I didn't touch for this PR.

src/Wippersnapper_V2.cpp
```diff
@@ -0,0 +1,1314 @@
...
+      WS_DEBUG_PRINTLN("ERROR: Unable to remove DS18X20 sensor!");
+      return false;
+    }
+    break;
+  default:
+    WS_DEBUG_PRINTLN("ERROR: BrokerToDevice message type not found!");
+    return false;
+  }
+
+  // once this is returned, pb_dec_submessage()
```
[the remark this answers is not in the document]
brentru: Yeah, the controller handles instead of `pb_dec_submessage`

src/Wippersnapper_V2.cpp
brentru: Are you interested in logging which _specific_ message failed?

src/Wippersnapper_V2.cpp
brentru: This would be out of scope for this review as it was not touched as part of the offline mode work.

src/Wippersnapper_V2.cpp
brentru: Marking this as out of scope

src/Wippersnapper_V2.cpp
brentru: Agreed, removed the `else` condition as the check within `if` `return`s.

src/provisioning/sdcard/ws_sdcard.h
tyeth: Yes 👍

src/Wippersnapper_V2.cpp
tyeth: I guess not really in general, except once we have a bug report / issue, and at that point we can work out where it fell over.

src/Wippersnapper_V2.cpp
tyeth: not relevant to this PR, apologies, but yeah the enableLogging call has no method body content

src/provisioning/sdcard/ws_sdcard.h
brentru: Increased that buffer size to `8192` bytes. In the grand scheme of things in WS and dynamically parsing a JSON file, it's not overkill

tests/diagram.json
brentru: Even if we're only using one DS18b20?

src/Wippersnapper_V2.cpp
brentru: Here's what I think - this is the _high-level_ decoder method. If an error during decode occurs within the controller, the controller will log specific messages to the user.

src/ws_manager.cpp
```diff
@@ -0,0 +1,101 @@
...
+*/
+/**************************************************************************/
+void Wippersnapper_Manager::checkAPIVersion(int pinNum) {
+  // Check if pin D12 is high
+  pinMode(pinNum, INPUT_PULLUP);
+  bool readButton = digitalRead(pinNum);
+  // NOTE: For debugging right now, we are forcing the API version
+  readButton = true;
+  if (readButton) { // API version 2 if D12 is high
+    ws_instance_v2 = new ws_adapter_wifi_v2();
```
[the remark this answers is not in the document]
brentru: I am confused here. What should L5:15 look like in `ws_manager.h`?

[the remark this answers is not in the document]
tyeth:
```cpp
// For compiling with ESP32
#include "adapters/wifi/ws_wifi_esp32.h"
//typedef ws_wifi_esp32 ws_adapter_wifi;
#include "adapters/wifi/ws_wifi_esp32_v2.h"
//typedef ws_wifi_esp32_v2 ws_adapter_wifi_v2;

// Uncomment the following lines to compile for Pico
// NOTE: When compiling for pico, you must also modify L39-L40 to reflect the following lines
#include "adapters/offline/ws_offline_pico.h"
typedef ws_offline_pico ws_adapter_offline;
#include "adapters/offline/ws_offline_pico_v2.h"
typedef ws_offline_pico_v2 ws_adapter_offline_v2;
typedef ws_offline_pico ws_adapter_wifi;
typedef ws_offline_pico_v2 ws_adapter_wifi_v2;
```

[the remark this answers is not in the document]
tyeth: I should mention I added this at line 194 of boards.h as the default pico target in platformio.ini uses the non-w board:
```
#elif defined(ARDUINO_RASPBERRY_PI_PICO)
#define BOARD_ID "rpi-pico"
#define USE_TINYUSB
#define USE_STATUS_LED
#define STATUS_LED_PIN 32
#elif defined(ARDUINO_RASPBERRY_PI_PICO_2W)
#define BOARD_ID "rpi-pico-2w"
#define USE_TINYUSB
#define USE_STATUS_LED
#define STATUS_LED_PIN LED_BUILTIN
```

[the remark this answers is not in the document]
brentru: Got it! Thanks for clarifying.

src/provisioning/sdcard/ws_sdcard.h
brentru: This has been removed in favor of streaming test data over the serial in debug mode.

tests/diagram.json
tyeth: I think I saw an error/warning logged about having 0 as the serial number, but not fatal. Wouldn't worry too much about it, although having moving data points also seemed useful (but think we talked in the past and said constant)

src/provisioning/sdcard/ws_sdcard.cpp
brentru: Totally, I'm refactoring that in the next commit! Great catch.

src/provisioning/tinyusb/Wippersnapper_FS_V2.cpp
tyeth: Yeah sorry I should be clearer. Line 71 gets the result of f_mkfs into r, then like 75 overwrites that result with the mount attempt, then we check r. I'd expect to check for failure before the mount too (same end result probably not worth changing)

brentru: Got it! Changed the logic to return r if `FR_OK` is not ret'd on that line, also modified L88 to perform a similar check since it's setting `r` but not returning until the function ends.

examples/config-file/offline-config.json
brentru: > This degree symbol is \u encoded in another config (src/config.json)

Removed it, that was a duplicate file.

> Also the sensor component appears to be including two metrics, so should be named differently, and presumably major rework for the front end / rails to accommodate (but totally in line with our talk of one poll period per sensor instead of subcomponent).

Could you please clarify this point - not sure I follow

tyeth: Ignore that last one about the two metrics for the ds18b20 in config, but I was alluding to the fact the "name" mentioned *F but was actually for *C and *F (two metrics), and that we don't have an example config structure for that yet (so was unfamiliar with sensor_count etc) nor frontend support (both expected)

tyeth: To clarify it's just that the name of that component object is for both of the two subcomponents (°C and °F), so I guess we drop the fahrenheit from the name field, and the web front end will already give them units.

brentru:
@tyeth  Please give this a re-review. I have changed what we discussed. I'm going to merge this branch into the api-v2 branch after you give the OK.

src/provisioning/sdcard/ws_sdcard.cpp
```diff
@@ -0,0 +1,933 @@
...
+  case wippersnapper_sensor_SensorType_SENSOR_TYPE_ALTITUDE:
+    return "m";
+  case wippersnapper_sensor_SensorType_SENSOR_TYPE_LUX:
+    return "raw value";
+  case wippersnapper_sensor_SensorType_SENSOR_TYPE_ECO2:
+    return "ppm";
+  case wippersnapper_sensor_SensorType_SENSOR_TYPE_UNITLESS_PERCENT:
+    return "%";
+  case wippersnapper_sensor_SensorType_SENSOR_TYPE_AMBIENT_TEMPERATURE_FAHRENHEIT:
+    return "";
```
tyeth, line 740: This looks like it's missing

src/provisioning/sdcard/ws_sdcard.cpp
```diff
@@ -0,0 +1,933 @@
...
+            The SensorType enum to convert.
+    @returns A string representation of the SensorType enum.
+*/
+/**************************************************************************/
+const char *SensorTypeToSIUnit(wippersnapper_sensor_SensorType sensorType) {
+  switch (sensorType) {
+  case wippersnapper_sensor_SensorType_SENSOR_TYPE_UNSPECIFIED:
+    return "raw value";
+  case wippersnapper_sensor_SensorType_SENSOR_TYPE_ACCELEROMETER:
+    return "m/s/s";
```
tyeth, line 682: feels like this should match gravity/linear-acceleration, but happy to not have it match if preferred

src/provisioning/sdcard/ws_sdcard.cpp
```diff
@@ -675,82 +674,82 @@ uint32_t ws_sdcard::GetTimestamp() {
...
+    return "NOx";
   case wippersnapper_sensor_SensorType_SENSOR_TYPE_TVOC:
-    return "TVOC";
+    return "ppb";
   case wippersnapper_sensor_SensorType_SENSOR_TYPE_BYTES:
-    return "BYTES";
+    return "bytes";
   case wippersnapper_sensor_SensorType_SENSOR_TYPE_BOOLEAN:
-    return "BOOLEAN";
+    return "raw value";
```
tyeth, line 752: I wonder about leaving this blank, but having an empty field might be more frustrating for people

tyeth (changes requested):
Minor changed in units method, the ambient fahrenheit specifically, maybe accelerometer too.

brentru:
Merged into migrate-api-v2. Thank you for the review!

PR #682, opened 2025-01-14
Swap staging root certificate to ISRG Root X1

tyeth:
Tested locally on QTPY S2, needs a normal esp32 and picow / 2w ideally tested too


tyeth:
Tested on PicoW + 2W, verified as required and then working.

PR #688, opened 2025-01-27
DRAFT: Publish returns bool + print info on publishes

tyeth:
This swaps all direct calls to mqtt client .publish (which can access the return bool) to use WS.publish which now returns bool.


tyeth:
Saw issues on network status check, revisit, also see #696 #697

PR #689, opened 2025-01-31
Update WipperSnapper_I2C_Driver_SCD4X.h - renamed include

tyeth:
v1.0.0 of the driver has been released, it renames the header (breaking) and alters the serial read function (not affected) along with a couple of other function names.

See https://github.com/Sensirion/arduino-i2c-scd4x/issues/45


brentru:
@tyeth  Is this ready for review? I experienced this with the refactor and SCD4x

brentru (approved):
Approved, let's get this into `main`. I also need to reformat drvSCD4x in the refactor

PR #690, opened 2025-01-31
fix: timeout datatype in haltError

tyeth:
Spotted while testing the DS18b20 (which has been fixed upstream)

### DIFF
--- a/src/Wippersnapper.cpp
+++ b/src/Wippersnapper.cpp
@@ -2504,9 +2504,9 @@ void Wippersnapper::runNetFSM() {
 void Wippersnapper::haltError(String error, ws_led_status_t ledStatusColor,
                               uint8_t seconds_until_reboot) {
 #ifdef ARDUINO_ARCH_ESP8266
-  uint8_t wdt_timeout_ms = 3200;
+  uint16_t wdt_timeout_ms = 3200;
 #else
-  uint8_t wdt_timeout_ms = 5000;
+  uint16_t wdt_timeout_ms = 5000;
 #endif
   int seconds_until_wdt_enable =
       seconds_until_reboot - (int)(wdt_timeout_ms / 1000);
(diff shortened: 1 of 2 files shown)


src/Wippersnapper.cpp
```diff
@@ -2504,9 +2504,9 @@ void Wippersnapper::runNetFSM() {
 void Wippersnapper::haltError(String error, ws_led_status_t ledStatusColor,
                               uint8_t seconds_until_reboot) {
```
brentru, line 2505: Should `seconds_until_reboot` also be `uint8_t`?

src/Wippersnapper.cpp
```diff
@@ -2504,9 +2504,9 @@ void Wippersnapper::runNetFSM() {
...
 #ifdef ARDUINO_ARCH_ESP8266
-  uint8_t wdt_timeout_ms = 3200;
+  uint16_t wdt_timeout_ms = 3200;
```
brentru, line 2507: The function signature of `enableWDTV2(int timeoutMS)`, which takes in this `wdt_timeout_ms` value is `int` not `uint`. Why are you changing it to uint16_t?

brentru (changes requested):
@tyeth  Reviewed, have some questions

src/Wippersnapper.cpp
tyeth: I think I wasn't really considering downstream use here, more the minimum size of the datatype needed to store necessary milliseconds along with in the back of my mind considering data types using the signedness to hint at acceptable values.

Also with the comment above about function argument type, shall I stick them both over to int and move on?

brentru: Yeah, moving them over to the correct type would be perfect. Please re-request a review when done.

brentru (approved):

PR #695, opened 2025-02-19
Add LPS28DFW

tyeth:


brentru:
@tyeth  This was marked ready for review but I was not assigned, do you want me to take a look at it?

src/components/i2c/drivers/WipperSnapper_I2C_Driver_LPS28DFW.h
```diff
@@ -0,0 +1,147 @@
...
+      if (_lps28->getStatus() & LPS28_STATUS_PRESS_READY) {
+        if (_temp == NULL) {
+          _temp = _lps28->getTemperatureSensor();
+        }
+        if (_pressure == NULL) {
+          _pressure = _lps28->getPressureSensor();
+        }
+        return true;
+      }
+      delay(1);
```
brentru, line 101: This may take more than 100ms in total. Instead of constantly checking the status register, consider implementing a longer delay of 20ms and change the loop condition to match (i < 5)

brentru (changes requested):

src/components/i2c/drivers/WipperSnapper_I2C_Driver_LPS28DFW.h
tyeth: Sure thing!

 I'm curious about the advantage of that, is it just that the status read may take more time than 0ms so the delay plus read times make my 100ms comment incorrect (and a longer time?).
It does "feel" better to only try a few times with a more sensible delay, but then again trying more times and returning more quickly equally has benefit.

I did it this way as the example did, and we were effectively blocking here anyway.

tyeth:
> @tyeth This was marked ready for review but I was not assigned, do you want me to take a look at it?

Thanks for taking a look!

src/components/i2c/drivers/WipperSnapper_I2C_Driver_LPS28DFW.h
brentru: >  is it just that the status read may take more time than 0ms so the delay plus read times make my 100ms comment incorrect (and a longer time?).

Yep

> I did it this way as the example did, and we were effectively blocking here anyway.

If you are following the example, let's leave this be

brentru (approved):

brentru:
LGTM once you are ready

PR #718, opened 2025-04-03
Add INA260

tyeth:
Tested with high side load, running an LED and resistor off of 3v3, reports 6.25mA and sometimes 5/7.5mA, voltage pretty steady at 3.285-7v on Metro S3


tyeth (commented):
@brentru ready for review.

brentru (approved):
@tyeth  LGTM 🚢

PR #746, opened 2025-05-09
Update PM2.5 begin_UART call for PM1006 (PM25AQI v2.0.0)

### LINKED ISSUE #745: Update V1 codebase to use latest release of PM25 library (is_pm1006 flag)
The UART driver for PM25 sensors no longer auto-detects. We need to pass true/false depending on chosen component.

https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/blob/main/src/components/uart/drivers/ws_uart_drv_pm25aqi.h#L85-L92

See similar check at https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/blob/main/src/components/uart/drivers/ws_uart_drv_pm25aqi.h#L151

tyeth:
Closes #745

### DIFF
--- a/src/components/uart/drivers/ws_uart_drv_pm25aqi.h
+++ b/src/components/uart/drivers/ws_uart_drv_pm25aqi.h
@@ -81,14 +81,17 @@ public:
   /*******************************************************************************/
   bool begin() override {
     _aqi = new Adafruit_PM25AQI();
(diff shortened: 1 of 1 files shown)


tyeth:
@brentru  Looks like this is going to pass CI now (new library version being picked up).
Untested, but looked pretty harmless (worth testing before merge).

brentru:
Will review when passing!

PR #760, opened 2025-06-19
Add SH1107 for 128x64/32 OLED featherwings

tyeth:
This ties in with https://github.com/adafruit/Wippersnapper_Components/pull/263

The build task was modified to allow single board builds, due to circuitpython hogging the build agents.

The SH1107 is added, for the featherwing. It uses the width as height and height as width in the learn guide example, then relying on rotation = 1. I've had to include this to make the board usable in the interface, but it now begs the question of whether we add rotation to the protobuf, and as a degrees decimal, or possibly just 0-3.

Multiple SSD1306 options are added to support extra displays/components, and the library args to no longer call i2c begin.

### DIFF
--- a/library.properties
+++ b/library.properties
@@ -1,5 +1,5 @@
 name=Adafruit WipperSnapper
-version=1.0.0-beta.103
+version=1.0.0-beta.104
 author=Adafruit
 maintainer=Adafruit <adafruitio@adafruit.com>
 sentence=Arduino application for Adafruit.io WipperSnapper
--- a/src/Wippersnapper.h
+++ b/src/Wippersnapper.h
@@ -142,7 +142,7 @@
 #endif
 
 #define WS_VERSION                                                             \
-  "1.0.0-beta.103" ///< WipperSnapper app. version (semver-formatted)
+  "1.0.0-beta.104" ///< WipperSnapper app. version (semver-formatted)
 
 // Reserved Adafruit IO MQTT topics
 #define TOPIC_IO_THROTTLE "/throttle" ///< Adafruit IO Throttle MQTT Topic


tyeth:

src/components/i2c/WipperSnapper_I2C.cpp
```diff
@@ -896,7 +927,8 @@ bool WipperSnapper_Component_I2C::initI2CDevice(
         (uint8_t)msgDeviceInitReq->i2c_output_add.config.ssd1306_config.width,
         (uint8_t)msgDeviceInitReq->i2c_output_add.config.ssd1306_config.height,
         (uint8_t)
-            msgDeviceInitReq->i2c_output_add.config.ssd1306_config.text_size);
+            msgDeviceInitReq->i2c_output_add.config.ssd1306_config.text_size,
+        0); // TODO: add rotation to protobuf and IO UI for adapted max len
```
brentru, line 931: `0` is a magic number, can you define this elsewhere, or include in a ctor if it's the default and will always be set?

src/components/i2c/drivers/WipperSnapper_I2C_Driver_Out_Sh1107.h
```diff
@@ -0,0 +1,220 @@
+/*!
+ * @file WipperSnapper_I2C_Driver_Out_SH1107.h
+ *
+ *  Device driver for a SH1107 OLED Display
+ *
+ * Adafruit invests time and resources providing this open source code,
+ * please support Adafruit and open-source hardware by purchasing
+ * products from Adafruit!
+ *
+ * Copyright (c) Brent Rubell for Adafruit Industries 2025
```
brentru, line 10: Change this copyright to your name?

src/components/i2c/drivers/WipperSnapper_I2C_Driver_Out_Sh1107.h
```diff
@@ -0,0 +1,220 @@
...
+ *
+ * MIT license, all text here must be included in any redistribution.
+ *
+ */
+
+#ifndef WIPPERSNAPPER_I2C_DRIVER_OUT_SH1107_H
+#define WIPPERSNAPPER_I2C_DRIVER_OUT_SH1107_H
+
+#include "WipperSnapper_I2C_Driver_Out.h"
+// #include <Adafruit_GrayOLED.h>
```
brentru, line 20: Is this supposed to be commented out?

src/components/i2c/drivers/WipperSnapper_I2C_Driver_Out_Sh1107.h
```diff
@@ -0,0 +1,220 @@
...
+#include <Adafruit_GFX.h>
+#include <Adafruit_SH110X.h>
+#include <Arduino.h>
+
+#define DEFAULT_WIDTH 128 ///< Default width for a sh1107 128x64 display
```
brentru, line 25: Generic definition. Change #define to something specific like `SHT1107_DEFAULT_WIDTH`.

src/components/i2c/drivers/WipperSnapper_I2C_Driver_Out_Sh1107.h
```diff
@@ -0,0 +1,220 @@
...
+#define DEFAULT_HEIGHT 64 ///< Default height for a sh1107 128x64 display
```
brentru, line 26: same as above

src/components/i2c/drivers/WipperSnapper_I2C_Driver_Out_Ssd1306.h
```diff
@@ -53,8 +53,10 @@ class WipperSnapper_I2C_Driver_Out_Ssd1306
   */
   ~WipperSnapper_I2C_Driver_Out_Ssd1306() {
     if (_display != nullptr) {
+      WS_DEBUG_PRINTLN("SSD1306 Destructor called, clearing display buffer.");
```
brentru, line 56: Remove this debugging print for production firmware build

src/components/i2c/drivers/WipperSnapper_I2C_Driver_Out_Ssd1306.h
```diff
@@ -53,8 +53,10 @@ class WipperSnapper_I2C_Driver_Out_Ssd1306
...
       _display->clearDisplay();
       _display->display();
+      WS_DEBUG_PRINTLN("SSD1306 display cleared, turning off display.");
```
brentru, line 59: Remove this debugging print for production firmware build

src/components/i2c/drivers/WipperSnapper_I2C_Driver_Out_Ssd1306.h
```diff
@@ -68,9 +70,10 @@ class WipperSnapper_I2C_Driver_Out_Ssd1306
   bool begin() {
     // Attempt to create and allocate a SSD1306 obj.
     _display = new Adafruit_SSD1306(_width, _height, _i2c);
-    if (!_display->begin(SSD1306_SWITCHCAPVCC, _sensorAddress))
+    if (!_display->begin(SSD1306_SWITCHCAPVCC, _sensorAddress, false, false))
```
brentru, line 73: Why are the default args `bool reset = true, bool periphBegin = true` getting modified here? 

What wasn't happening?

src/components/i2c/drivers/WipperSnapper_I2C_Driver_Out_Ssd1306.h
```diff
@@ -90,11 +93,17 @@ class WipperSnapper_I2C_Driver_Out_Ssd1306
...
+                  The rotation of the display in degrees, default is 0.
   */
-  void ConfigureSSD1306(uint8_t width, uint8_t height, uint8_t text_size) {
+  void ConfigureSSD1306(uint8_t width, uint8_t height, uint8_t text_size,
+                        uint8_t rotation = 0) {
     _width = width;
     _height = height;
     _text_sz = text_size;
+    _rotation = rotation;
+    WS_DEBUG_PRINT("SSD1306 text size: ");
```
brentru, line 105: Remove default prints for prod. build

src/components/i2c/drivers/WipperSnapper_I2C_Driver_Out_Ssd1306.h
```diff
@@ -106,6 +115,9 @@ class WipperSnapper_I2C_Driver_Out_Ssd1306
     if (_display == nullptr)
       return;
 
+    WS_DEBUG_PRINT("SSD1306 Message:");
```
brentru, line 118: debug, remove

brentru (changes requested):
@tyeth  I'm requesting some changes as it's the first PR you've done against the displays. I also have some questions

src/components/i2c/drivers/WipperSnapper_I2C_Driver_Out_Sh1107.h
```diff
@@ -0,0 +1,220 @@
...
+      _display = new Adafruit_SH1107(_width, _height, _i2c);
+    }
+    if (!_display->begin(_sensorAddress, true))
+      return false;
+
+    // Show image buffer on the display hardware.
+    // Since the buffer is intialized with an Adafruit splashscreen
+    // internally, this will display the splashscreen.
+    _display->display();
+    delay(1700);
```
tyeth, line 84: I really like the idea of displaying the adafruit logo at screen start, as it's bundled with the libraries. On the other hand the large delay is sub optimal, and we can save some space by not compiling them in (a future build flag in platform.local.txt etc like SSD1306_NO_SPLASH and SH110X_NO_SPLASH )

src/components/i2c/WipperSnapper_I2C.cpp
```diff
@@ -881,7 +881,38 @@ bool WipperSnapper_Component_I2C::initI2CDevice(
     }
     _drivers_out.push_back(_sevenSeg);
     WS_DEBUG_PRINTLN("7-Segement LED Matrix Initialized Successfully!");
-  } else if (strcmp("oled128x32default", msgDeviceInitReq->i2c_device_name) ==
+  } else if (strcmp("oled128x64wingdefault",
```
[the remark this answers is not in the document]
tyeth: Was going to abbreviate wing to w, and default/large to d/l but then makes me think we should do the others (which are probably set in stone now). Is it worth abbreviating both, or just wing? I feel like the abbreviations won't be too obvious except to you and I.
To be fair people can always go into the definition files if they are aware of the components repo.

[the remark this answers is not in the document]
brentru: Commenting to note that we discussed this over Slack

PR #764, opened 2025-07-08
Xiao s3

tyeth:
This should probably have the same web-native-usb install method as the esp32s3 devkit + funhouse. - UPDATE BOARDS REPO


.github/workflows/build-clang-doxy.yml
```diff
@@ -192,6 +192,7 @@ jobs:
             "qtpy_esp32s2",
             "feather_esp32s3_reverse_tft",
             "qtpy_esp32s3_n4r2",
+            "xiao_esp32s3",
```
tyeth, line 195: swap to the web-native-usb / lvgl build task, to generate appropriate esptool zip

tyeth:
This should also use the same web-native-usb installer method (use esptool and .bin not uf2). Worth waiting until that is merged before adding this as a similar build asset

tyeth:
@brentru this is ready for review, but I want to test the assets via the installer once the boards merged (unpublished) before this PR gets merged.

PR #771, opened 2025-07-14
Support for DFRobot Beetle ESP32C3.

tyeth:
See https://github.com/adafruit/Wippersnapper_Boards/issues/195

### DIFF
--- a/.github/workflows/build-clang-doxy.yml
+++ b/.github/workflows/build-clang-doxy.yml
@@ -318,11 +318,14 @@ jobs:
             "sparklemotionstick_esp32",
             "feather_esp32_v2",
             "itsybitsy_esp32",
+            "dfrobot_beetle_esp32c3",
             "wippersnapper_qtpy_esp32c3",
             "wippersnapper_feather_esp32c6"
           ]
         include:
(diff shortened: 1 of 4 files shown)


tyeth:
As noted in discord, it won't yet build in CI for the dfrobot due to not having an entry here: https://github.com/kbsriram/Adafruit_Wippersnapper_Arduino/blob/main/.github/workflows/build-clang-doxy.yml#L320
and the .test.skip for the debug example folder, and .generate for the demo example folder (like qtpy does) 
[21:44]Tyeth: Oh sorry, and in the same github actions file just beneath add these two for your board too: https://github.com/kbsriram/Adafruit_Wippersnapper_Arduino/blob/main/.github/workflows/build-clang-doxy.yml#L328-L329

kbsriram:
I tested a few, mostly to confirm pin assignments - all these seem to be working for me:

- all digital pins (except for pads on the back) as outputs and PWM outputs (via "dimmable LEDs")
- internal LED and button
- I2C to the extent that I can run a scan from AIO, and it shows the device on the bus (the ones I have at hand don't have an existing integration)

PR #779, opened 2025-07-18
tinyuf2 build task changes (PR #774) plus Feather s2 migration

tyeth:
This combines the build task changes PR #774 with the single board test release for Feather S2.

### DIFF
--- a/.github/workflows/build-clang-doxy.yml
+++ b/.github/workflows/build-clang-doxy.yml
@@ -145,20 +145,14 @@ jobs:
       - name: fetch tinyuf2 combined.bin
         run: |
           echo "Fetching tinyuf2 bootloader..."
-          # if [ "${{ matrix.arduino-platform }}" == "funhouse_noota" ]; then
-          #   wget https://adafruit-circuit-python.s3.amazonaws.com/bootloaders/esp32/adafruit_funhouse_esp32s2/tinyuf2-adafruit_funhouse_esp32s2-0.35.0.zip -O tinyuf2.zip
-          # elif [ "${{ matrix.arduino-platform }}" == "esp32s3_devkitc_1_n8" ]; then
-          #   wget https://adafruit-circuit-python.s3.amazonaws.com/bootloaders/esp32/espressif_esp32s3_devkitc_1/tinyuf2-espressif_esp32s3_devkitc_1-0.35.0.zip -O tinyuf2.zip
-          # elif [ "${{ matrix.arduino-platform }}" == "xiao_esp32s3" ]; then
-          #   wget https://adafruit-circuit-python.s3.amazonaws.com/bootloaders/esp32/seeed_xiao_esp32s3/tinyuf2-seeed_xiao_esp32s3-0.35.0.zip -O tinyuf2.zip
-          # fi
-
           DOWNLOAD_URL=$(curl -s https://api.github.com/repos/adafruit/tinyuf2/releases/latest | \
           grep -o "\"browser_download_url\": \"[^\"]*tinyuf2-${{fromJson(steps.get_board_json.outputs.boardJson).bootloaderBoardName}}-[^\"]*\.zip\"" | \
           cut -d '"' -f 4)
-        
+          if [ -z "$DOWNLOAD_URL" ]; then
+            echo "Error: Could not find the latest tinyuf2 release for board ${fromJson(steps.get_board_json.outputs.boardJson).bootloaderBoardName}."
+            exit 1
+          fi
           wget "$DOWNLOAD_URL" -O tinyuf2.zip
-
           unzip -o tinyuf2.zip -d .
           
           # Copy files where they're expected to make flash_args happy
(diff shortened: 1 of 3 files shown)


tyeth:
@brentru this is ready for review. It obviously needs the build task for the Sx/LvGL (web-native-usb) boards to use the updated definitions in the boards repo, which is why that task has the ref switched.

.github/workflows/build-clang-doxy.yml
```diff
@@ -33,6 +34,8 @@ jobs:
           - offset: "0x0"
           - offset: "0x1000"
             arduino-platform: "funhouse_noota"
+          - offset: "0x1000"
+            arduino-platform: "feather_esp32s2"
```
brentru, line 38: Why is the offset defined here _and_ in the boards repo?

.github/workflows/build-clang-doxy.yml
```diff
@@ -141,6 +145,36 @@ jobs:
             echo $content
             echo EOF
           } >> "$GITHUB_OUTPUT"
+      - name: fetch tinyuf2 combined.bin
+        run: |
+          echo "Fetching tinyuf2 bootloader..."
```
brentru, line 150: don't need the echo on final version

.github/workflows/build-clang-doxy.yml
```diff
@@ -141,6 +145,36 @@ jobs:
...
+          DOWNLOAD_URL=$(curl -s https://api.github.com/repos/adafruit/tinyuf2/releases/latest | \
+          grep -o "\"browser_download_url\": \"[^\"]*tinyuf2-${{fromJson(steps.get_board_json.outputs.boardJson).bootloaderBoardName}}-[^\"]*\.zip\"" | \
+          cut -d '"' -f 4)
```
brentru, line 153: the api../latest is a JSON file. Could you use `jq` to parse the file, rather than pattern-match with regex?

.github/workflows/build-clang-doxy.yml
```diff
@@ -141,6 +145,36 @@ jobs:
...
+          if [ -z "$DOWNLOAD_URL" ]; then
+            echo "Error: Could not find the latest tinyuf2 release for board ${fromJson(steps.get_board_json.outputs.boardJson).bootloaderBoardName}."
+            exit 1
+          fi
+          wget "$DOWNLOAD_URL" -O tinyuf2.zip
+          unzip -o tinyuf2.zip -d .
+
+          # Copy files where they're expected to make flash_args happy
```
brentru, line 161: This should be another step, the current step is "fetch tinyuf2 combined.bin"

.github/workflows/build-clang-doxy.yml
```diff
@@ -141,6 +145,36 @@ jobs:
...
+          echo "0x10000 wippersnapper.${{ matrix.arduino-platform }}.fatfs.${{ env.WS_VERSION }}.bin" >> flash_args
+
+          # Create new_flash_args with esptool parameters first and output file
+          echo "--flash-mode ${{fromJson(steps.get_board_json.outputs.boardJson).esptool.flashMode}}" > new_flash_args
+          echo "--flash-freq ${{fromJson(steps.get_board_json.outputs.boardJson).esptool.flashFreq}}" >> new_flash_args
+          echo "--flash-size ${{fromJson(steps.get_board_json.outputs.boardJson).esptool.flashSize}}" >> new_flash_args
+          echo "-o wippersnapper.${{ matrix.arduino-platform }}.fatfs.${{ env.WS_VERSION }}.combined.bin" >> new_flash_args
+
+          # Append flash_args content skipping the first line
+          tail -n +2 flash_args >> new_flash_args
```
brentru, line 177: This should be its own step.

.github/workflows/build-clang-doxy.yml
brentru: Maybe extend it down to L165 (before the next `#` comment)?

.github/workflows/build-clang-doxy.yml
```diff
@@ -156,23 +190,17 @@ jobs:
           ls /home/runner/Arduino/hardware/espressif/esp32/tools/partitions
       - name: boot_app0 file from arduino-cli core
         if: steps.check_files.outputs.files_exists == 'true'
-        run: mv /home/runner/.arduino15/packages/esp32/hardware/esp32/*/tools/partitions/boot_app0.bin wippersnapper.${{ matrix.arduino-platform }}.fatfs.${{ env.WS_VERSION }}.boot_app0.bin
+        run: cp /home/runner/.arduino15/packages/esp32/hardware/esp32/*/tools/partitions/boot_app0.bin wippersnapper.${{ matrix.arduino-platform }}.fatfs.${{ env.WS_VERSION }}.boot_app0.bin
```
brentru, line 193: Are you changing to `cp` instead of `mv` to keep the original file in the same place? What's the rationale?

brentru (changes requested):

.github/workflows/build-clang-doxy.yml
tyeth: Different things. This is the bootloader offset. The definition file contains the filesystem offset.

.github/workflows/build-clang-doxy.yml
tyeth: I never saw the rationale in moving it, as that requires delete permissions too. I'll revert if you'd prefer

Issue #360, opened 2022-11-08
Support GPS Modules 


brentru:
* Support for I2C GPS modules such as the [Mini GPS PA1010D](https://www.adafruit.com/product/4415)
  * Support generic UART GPS modules such as https://www.adafruit.com/product/5440
* Support sending [NMEA RMC Messages](https://orolia.com/manuals/VSP/Content/NC_and_SS/Com/Topics/APPENDIX/NMEA_RMCmess.htm) every ~ period (in Minutes) for GPS tracking

Issue #699, opened 2025-02-21
Add support for GPS location in log files


brentru:
For Offline Mode, we can implement this without touching IO (yet)!

* Add GPS module's timestamp (#698 ) and location fix to the JSONL log line

Requires: https://github.com/adafruit/Wippersnapper_Protobuf/issues/118

brentru:
Should implement UART first and then AQ sensors and then GPS, as GPS modules would require that as a prerequisite. 

#639

brentru:
@ladyada  Do we want to implement this feature or mark it long-term? All the pieces I need to do this are implemented.

ladyada:
kinda up to you! how much do you want to stay in GPS land?

brentru:
Added to https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/699! This issue close when the pull request is merged.

brentru:
Closed via https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/787#event-19114529302

Issue #639, opened 2024-10-03
[API-v2] Refactor UART 


brentru:
This is the main issue for tracking the UART refactor for API v2.

Issue #755, opened 2025-05-28
Support "Generic UART" Input Components


brentru:
Add support for "Generic UART" input sensors or components. These are devices that output a generic string over UART, like distance sensors. The data is emitted at a periodic interval (100ms, 1000ms) and ends with either a newline or a newline+CR.

PR #787, opened 2025-07-30
Offline Mode - UART and GPS

brentru:
This pull request brings UART and GPS modules into WipperSnapper's offline mode and the upcoming v2 (online) firmware.

Since these two components are completely rewritten for the WipperSnapper API-v2, they are not yet ready to be added to WipperSnapper Online until the IO Production broker supports WipperSnapper API v2.

UART (v2 refactor) 🔧 - UART has been fully refactored and rebuilt to support components other than the PM25 AQ Sensor. The UART component architecture is similar to the i2c component and provides a general driver base class (drvBase) for creating a new UART driver.
- UART now supports sensor.proto API, shared by other sensing classes 
- UART may support output components (such as servos or motors) in the future, its API has been extended to allow for this capability 
- NEW UART Components: PM25 AQ Sensor, US100 Ultrasonic Sensor, Adafruit Ultimate GPS Module 

GPS 🛰️ - GPS functionality has been added to WipperSnapper! This is a new feature ✨
- The GPS component (src/components/gps) can support GPS modules over I2C or UART
- Supports auto-detection of Mediatek (MTK) or Ublox (UBX) GPS modules 
- Support for standard NMEA parsing and logging (RMC and GGA messages only) into a standard interface (new gps.proto container) 
- We have added preliminary support for the following GPS modules and products:
  - [Adafruit Ultimate GPS Breakout with GLONASS + GPS - MTK MTK3339](https://www.adafruit.com/product/5440)
  - [Adafruit Ultimate GPS FeatherWing - MTK MTK3339](https://www.adafruit.com/product/3133)
  - Adafruit Mini GPS PA1010D  - MTK MTK3333
  - [SparkFun GNSS Receiver Breakout - MAX-M10S ](https://www.sparkfun.com/sparkfun-gnss-receiver-breakout-max-m10s-qwiic.html)

Resolves:
https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/360
https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/727
https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/755
https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/753


brentru:
Resolves https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/699

brentru:
Resolves https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/639

brentru:
@tyeth  This pull request is ready for review. Please note that the areas of note are within `ws_sdcard` and the GPS module. UART has already been reviewed.

src/components/gps/hardware.cpp
```diff
@@ -0,0 +1,749 @@
...
+            if (status != UBXSendStatus::UBX_SEND_SUCCESS) {
+              WS_DEBUG_PRINTLN("[gps] Failed to send UBX message");
+            } else {
+              WS_DEBUG_PRINTLN("[gps] OK");
+            } */
+    }
+  } else {
+    WS_DEBUG_PRINTLN("[gps] ERROR: Unsupported GPS driver type!");
+    return false;
+  }
```
tyeth, line 170: These feel like they are the begin/Handle_GPSConfig methods for individual GPS wrapper drivers. Maybe the content needs to be here in orchestration terms, but I was expecting to see some i2c/drivers/gps files, or gps/drivers files, and I imagine calling to the drivers like doing gpsDriverBase.flushInterface instead of
```
if (_iface_type == GPS_IFACE_UART_HW) {
        // Flush the RX/TX buffers before sending
        _hw_serial->flush();
        while (_hw_serial->available() > 0) {
          _hw_serial->read();
        }
      } else if (_iface_type == GPS_IFACE_I2C) {
        WS_DEBUG_PRINT("[gps] Flushing I2C buffers...");
        I2cReadDiscard();
        WS_DEBUG_PRINTLN(" done!");
      }
```
Maybe we don't need driver files for GNSS components 🤷

src/components/gps/hardware.cpp
```diff
@@ -0,0 +1,749 @@
...
+ */
+bool GPSHardware::SetInterface(HardwareSerial *serial, uint32_t baudrate) {
+  if (serial == nullptr)
+    return false;
+  // Configure the hardware serial interface
+  _hw_serial = serial;
+  _iface_type = GPS_IFACE_UART_HW;
+  _baudrate = baudrate;
+  return true;
+}
```
tyeth, line 188: Might need SoftwareSerial version

src/components/gps/hardware.cpp
tyeth, line 1: Not too keen on having each drivers specific methods in this file, feels like driver files might be better for future additions (and base driver methods). It doesn't necessarily matter though, and is probably not worth the refactor right now.

src/protos/gps.pb.h
```diff
@@ -0,0 +1,203 @@
...
+/* * GPSDateTime represents the date and time information from a GPRMC/GPGGA string * */
+typedef struct _wippersnapper_gps_GPSDateTime {
+    int32_t hour; /* * Hour of the day (0-23) * */
+    int32_t minute; /* * Minute of the hour (0-59) * */
+    int32_t seconds; /* * Seconds of the minute (0-59) * */
+    int32_t milliseconds; /* * Milliseconds (0-999) * */
+    int32_t day; /* * Day of the month (1-31) * */
+    int32_t month; /* * Month of the year (1-12) * */
+    int32_t year; /* * Year (e.g., 2023) * */
+} wippersnapper_gps_GPSDateTime;
```
tyeth, line 36: Not certain what happens in error situations / no-lock etc, nor with standard datetime function/property compatibility, but looks like these could all be uint8_t except year+milliseconds.

src/components/uart/drivers/drvUartUs100.h
```diff
@@ -0,0 +1,218 @@
...
+    }
+    return true;
+  }
+
+  Adafruit_GenericDevice *getGenericDevice() { return _generic_dev; }
+
+private:
+  Adafruit_GenericDevice *_generic_dev = nullptr;
+  Stream *_serial_dev = nullptr;
+};
```
tyeth, line 71: move to base class / helpers?

src/components/uart/drivers/drvUartUs100.h
```diff
@@ -0,0 +1,218 @@
...
+  */
+  ~drvUartUs100() { /* TODO: Add back CTOR */ }
+
+  /*!
+      @brief    Initializes the US-100 UART device.
+      @returns  True if initialized successfully, False otherwise.
+  */
+  bool begin() override {
+    // is _hw_serial nullptr?
+    if (_hw_serial == nullptr) {
```
tyeth, line 125: doesn't account for software serial

src/components/uart/drivers/drvUartUs100.h
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  /*!
+      @brief    Gets the US-100 sensor's distance measurement.
+      @param    rawEvent
+                The Raw value, contains the distance in centimeters.
+      @returns  True if the distance value was obtained successfully, False
+                otherwise.
+  */
+  bool getEventRaw(sensors_event_t *rawEvent) {
```
tyeth, line 183: Maybe should be getEventProximity(sensors_event_t *proximityEvent) instead of getEventRaw, matching the ToF VL53x sensors

src/components/uart/controller.cpp
```diff
@@ -0,0 +1,322 @@
...
+                                    cfg_device.device_type,
+                                    cfg_device.device_id, did_begin)) {
+    WS_DEBUG_PRINTLN("[uart] ERROR: Failed to encode UartAdded message!");
+    return false;
+  }
+  WS_DEBUG_PRINTLN("[uart] UartAdded message encoded successfully!");
+
+  WS_DEBUG_PRINTLN("[uart] Publishing UartAdded message to IO...");
+  // TODO: Unsure why this is causing a crash on GPS, figure out later
+  // Currently commented out to prevent crashes
```
tyeth, line 192: Hmm, curious, wonder why it's crashing logging that...

src/components/uart/hardware.cpp
```diff
@@ -0,0 +1,214 @@
...
+    // Create a new SoftwareSerial instance
+    _swSerial = new SoftwareSerial(rx_pin, tx_pin, _config.sw_serial_invert);
+    if (_swSerial == nullptr) {
+      WS_DEBUG_PRINTLN(
+          "[uart] ERROR: Failed to allocate a SoftwareSerial instance!");
+      return false;
+    }
+    _swSerial->begin((unsigned long)_config.baud_rate);
+    _baud_rate = _config.baud_rate;
+#endif // HAS_SW_SERIAL
```
tyeth, line 127: Should we raise an error here if software serial is requested but unavailable on this platform?

src/components/uart/model.cpp
```diff
@@ -0,0 +1,221 @@
...
+    break;
+  case wippersnapper_sensor_SensorType_SENSOR_TYPE_PM25_STD:
+    sensor_event.which_value = wippersnapper_sensor_SensorEvent_float_value_tag;
+    sensor_event.value.float_value = event.pm25_std;
+    break;
+  case wippersnapper_sensor_SensorType_SENSOR_TYPE_PM100_STD:
+    sensor_event.which_value = wippersnapper_sensor_SensorEvent_float_value_tag;
+    sensor_event.value.float_value = event.pm100_std;
+    break;
+  // TODO: Add more cases for other sensor types
```
tyeth, line 186: need temperature event added, I believe raw and proximity will use the data[0] correctly.

src/provisioning/sdcard/ws_sdcard.cpp
```diff
@@ -443,6 +464,97 @@ uint32_t ws_sdcard::HexStrToInt(const char *hex_str) {
...
+  msg_uart_add.cfg_serial.baud_rate = component["baudRate"] | 9600;
+  msg_uart_add.cfg_serial.format =
+      wippersnapper_uart_UartPacketFormat_UART_PACKET_FORMAT_8N1; // Stick to
+                                                                  // default 8N1
+                                                                  // for now
+  msg_uart_add.cfg_serial.timeout =
+      component["timeout"] | 1000; // Use a default UART timeout of 1000ms
+  msg_uart_add.cfg_serial.use_sw_serial = component["useSwSerial"] | false;
+  msg_uart_add.cfg_serial.sw_serial_invert =
+      component["swSerialInvert"] | false;
```
tyeth, line 495: These should eventually be added to the offline config tool for export.
To be fair the use_sw_serial thing is a board/device specific setting which the firmware knows, and swSerialInvert too perhaps? So the defaults of false may be incorrect or want to be #if HAS_SW_SERIAL guarded

src/provisioning/sdcard/ws_sdcard.cpp
```diff
@@ -879,6 +1016,8 @@ uint32_t ws_sdcard::GetTimestamp() {
     TickSoftRTC();
     return cur_time; // early-return as we are not converting this "soft rtc" to
                      // unixtime
+  } else if (WsV2._gps_controller->has_gps) {
+    now = WsV2._gps_controller->GetGPSDateTime();
```
tyeth, line 1020: Do we want to prioritise GPS over RTCs? Or sync our RTC with GPS periodically?

tyeth (changes requested):
Looks great generally

.github/workflows/release-offline.yml
```diff
@@ -214,6 +214,7 @@ jobs:
           git clone --quiet https://github.com/pstolarz/OneWireNg.git /home/runner/Arduino/libraries/OneWireNg
           git clone --quiet https://github.com/pstolarz/Arduino-Temperature-Control-Library.git /home/runner/Arduino/libraries/Arduino-Temperature-Control-Library
           git clone --quiet https://github.com/adafruit/Adafruit_TinyUSB_Arduino /home/runner/Arduino/libraries/Adafruit_TinyUSB_Arduino
+          git clone --quiet --branch fix-for-pico https://github.com/brentru/SdFat.git /home/runner/Arduino/libraries/SdFat_-_Adafruit_Fork/
```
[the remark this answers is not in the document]
brentru: This was picked up by this branch, not sure why. Maybe an issue with merging from main. I don't see it in https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/blob/1.0.0-offline-beta.3/.github/workflows/release-offline.yml

Resolving in next commit.

src/components/gps/controller.cpp
```diff
@@ -0,0 +1,194 @@
...
+  return true;
+}
+
+/*!
+ * @brief Adds a GPS hardware instance to the controller.
+ * @param serial Pointer to the HardwareSerial instance for GPS communication.
+ * @param gps_config Pointer to the GPS configuration message.
+ * @return True if the GPS was added successfully, false otherwise.
+ */
+bool GPSController::AddGPS(HardwareSerial *serial, uint32_t baudrate,
```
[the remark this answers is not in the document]
brentru: We do, we aren't handling `SoftwareSerial` immediately though.

src/components/gps/controller.cpp
```diff
@@ -0,0 +1,194 @@
...
+    return false;
+  }
+
+  gps_hw->SetI2CAddress(i2c_addr);
+  if (!gps_hw->begin()) {
+    WS_DEBUG_PRINTLN("[gps] ERROR: Failed to initialize module!");
+    delete gps_hw;
+    return false;
+  }
+
```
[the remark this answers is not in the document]
brentru: No, no delay required here.

src/components/gps/hardware.cpp
brentru: Would be a good future TODO to reduce complexity within `hardware.cpp`. We can mark an issue for this.

src/components/gps/hardware.cpp
brentru: Not doing SoftwareSerial in this, we can extend it if we need to in the future.

src/components/uart/drivers/drvUartUs100.h
brentru: @tyeth  This sensor returns in SI unit `cm`. Would the proximityEvent handle that? I suppose we could convert cm->m if there's not another unit available and you'd want to use the SI type.

src/components/uart/drivers/drvUartUs100.h
brentru: The `UARTDevice` class is only an abstraction for this specific device because it lacks a library, this driver provides what the library would.

src/components/uart/controller.cpp
brentru: Don't know but working as-is so removing

src/components/uart/hardware.cpp
brentru: No because HAS_SW_SERIAL is defined by the boards.h file

src/components/uart/model.cpp
brentru: Added temp (ambient, ambientF, object, objectF) and prox specific type

src/provisioning/sdcard/ws_sdcard.cpp
brentru: As of now, there is no strict prioritization. RTC will go first if it exists and that's fine. GPS would be a fallback if no RTC exists and sw rtc isnt defined.

src/provisioning/sdcard/ws_sdcard.cpp
tyeth: Should the false default value for useSwSerial be true if HAS_SW_SERIAL=1 ?

src/components/gps/controller.cpp
```diff
@@ -125,14 +125,6 @@ void GPSController::update() {
...
 
   for (GPSHardware *drv : _gps_drivers) {
-    // TODO: Commented out due to parsing failures, stability issue (failed to
-    // parse NMEA acks for this) Perform a keep-alive check by sending an
-    // antenna check command every 2 seconds
-    /*     if (millis() - drv->GetPrvKat() > 2000) {
-          drv->GetAdaGps()->sendCommand(CMD_MTK_CHECK_ANTENNA);
-          drv->SetPrvKat(millis());
-        } */
-
```
tyeth: I thought you said the code works now so removing this makes sense in reference to the comment, not the commented out code block.
Is the antenna status check unnecessary, or just problematic?

src/components/uart/drivers/drvUartUs100.h
tyeth: I'm not too fussed about the types, I just got a memory triggered about the prox/ToF sensors so raised it. 
I think we already have one sensor doing a divide by 10 for similar CM->MM reasons.

src/components/uart/hardware.cpp
tyeth: Brent Rubell
[@Tyeth]( ) I’m confused by [this comment](https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/787#discussion_r2261244145). rp2040 has hardware UART


Tyeth
This bit above (sorry for bad link) says if rp2040 add softserial [https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/787/files#diff-7eff18bcfa8b3bbf92bd0830bfab7c[…]a75f37a5a645fec88746aa4daa9b5R24-R36](https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/787/files#diff-7eff18bcfa8b3bbf92bd0830bfab7c76d6ca75f37a5a645fec88746aa4daa9b5R24-R36)
Is that just future use and we use HW serial for uart on rp2040? (edited) 


Brent Rubell
no it uses the swserial wrapper on rp2040
for uart


Tyeth
So does that mean using software serial for UART gps and pm25 and distance sensor components? (on rp2040)


Brent Rubell
for uart components (non gps yes).
gps doesn’t have swserial impl’d at the moment.
I think I can add it, let me use the bathroom and timebox this to 45min. It shouldn’t be difficult because ws_uart already exposes the swserial wrapper.

brentru: Implemented SoftwareSerial support within UART drivers in 5ef64390..6a54d02b

tyeth (approved):
This looks good Brent, thanks for making the software serial implementation.
Spotted some old debug logging on gps, maybe that's useful for now to leave on or have behind a flag, left minor comment

src/components/gps/hardware.cpp
```diff
@@ -0,0 +1,744 @@
...
+ * otherwise.
+ */
+bool GPSHardware::DetectUbxI2C(uint32_t addr) {
+  _ubx_gps_ddc = new Adafruit_UBloxDDC(addr, _wire);
+  if (!_ubx_gps_ddc->begin())
+    return false;
+  _ubx_gps = new Adafruit_UBX(*_ubx_gps_ddc);
+  if (!_ubx_gps->begin())
+    return false;
+  _ubx_gps->verbose_debug = 3; // TODO: Set this to 1 in production
```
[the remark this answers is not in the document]
brentru: This is part of the `Adafruit_UBX` library, not WipperSnapper. Does this change your opinion?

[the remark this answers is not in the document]
brentru: Hm, we'll see. I'd like to keep it for now.

PR #789, opened 2025-08-13
Add INA228 (included as base lib for INA23x)

tyeth:


tyeth (commented):
Untested as yet.

tyeth:
@brentru this is ready for review.
While doing offline mode I had accidentally plugged in an INA228, and realised that was on the list for after the refactor INA's from .h to .cpp in Online mode. I've now added it to offline mode in the other PR, and this brings to to online mode too.
Also spotted the INA260 was off by a factor of 1000 so fixed that too.

brentru (approved):
@tyeth  LGTM

PR #798, opened 2025-08-21
As5600

tyeth:
Adds the magnetic angle sensor.
Worth noting the device zeros on reset/init.

Not sure my angular calculation from the "scaled" value correctly gives 0-359 which I think is what's expected.
Also didn't have a great magnet for testing, but did see rotation, and returning to approximate expected angle


src/components/i2c/drivers/WipperSnapper_I2C_Driver_AS5600.h
```diff
@@ -0,0 +1,152 @@
...
+    // Check status conditions
+    if (_as5600->isAGCminGainOverflow()) {
+      WS_DEBUG_PRINTLN(" | MH: magnet too strong");
+      return false;
+    }
+    if (_as5600->isAGCmaxGainOverflow()) {
+      WS_DEBUG_PRINTLN(" | ML: magnet too weak");
+      return false;
+    }
+    _angle = ((float)angle / 4095.0) * 360.0;
```
tyeth, line 125: This doesn't feel like it's mathing correctly, wouldn't it be * 359?

src/components/i2c/drivers/WipperSnapper_I2C_Driver_AS5600.h
```diff
@@ -0,0 +1,152 @@
...
+      @param    i2c
+                The I2C interface.
+      @param    sensorAddress
+                7-bit device address.
+  */
+  /*******************************************************************************/
+  WipperSnapper_I2C_Driver_AS5600(TwoWire *i2c, uint16_t sensorAddress)
+      : WipperSnapper_I2C_Driver(i2c, sensorAddress) {
+    _i2c = i2c;
+    _sensorAddress = sensorAddress;
```
tyeth, line 42: part of base driver

src/components/i2c/drivers/WipperSnapper_I2C_Driver_AS5600.h
```diff
@@ -0,0 +1,152 @@
...
+    _as5600 = nullptr;
+    _angle = 0;
+  }
+
+  /*******************************************************************************/
+  /*!
+      @brief    Destructor for an AS5600 sensor.
+  */
+  /*******************************************************************************/
+  ~WipperSnapper_I2C_Driver_AS5600() { delete _as5600; }
```
tyeth, line 52: never called, check reinit of same sensor without reboot!
Otherwise if issues then check object in begin and delete if necessary

src/components/i2c/drivers/WipperSnapper_I2C_Driver_AS5600.h
```diff
@@ -0,0 +1,152 @@
...
+    uint16_t angle = _as5600->getAngle();
+
+    WS_DEBUG_PRINT("AS5600 Raw: ");
+    WS_DEBUG_PRINT(rawAngle);
+    WS_DEBUG_PRINT(" (0x");
+    WS_DEBUG_PRINT(rawAngle, HEX);
+    WS_DEBUG_PRINT(") | Scaled: ");
+    WS_DEBUG_PRINT(angle);
+    WS_DEBUG_PRINT(" (0x");
+    WS_DEBUG_PRINT(angle, HEX);
```
tyeth, line 113: ditch hex version

tyeth:
Note to self - Do not merge without testing reinitialising the as5600

tyeth:
As mentioned in the MLX pull request, I believe (not verified 100%) the destructors are not being called, due to the `delete *iter` line being disabled:
https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/blob/main/src/components/i2c/WipperSnapper_I2C.cpp#L1038
I only say that due to having the delete in there with an if check first plus set to nullptr, but it still crashed - now in actual fact that may have been due to testing reinitialisation too where the main adafruit_mlx driver didn't yet have the i2c_dev=nullptr fix applied that I later locally applied in testing.
I should verify if the destructors are being called, presumably by the erase_ later in the same code block of wipper_i2c.cpp

brentru (approved):
LGTM, feedback addressed

PR #799, opened 2025-08-21
Mlx90632

tyeth:
@brentru this one has had weird intermittent crashes at init, with no output after "attempting to initialise MLX".
The device does come online and report data fine.
Testing removing and re-adding the device it would often crash and reset (but come back as online and added).

I took a bit of time to debug, mainly getting my setup going again, and it seems to be crashing in the begin of the MLX driver, where it checks `if (i2c_dev)` it returns true on first run once when i watched it, and then ran `delete i2c_dev` and crashed.

Adding the `i2c_dev = nullptr` to the constructor resolved that. I've not done extensive testing since as didn't want to use more time on it. If you wouldn't mind checking the situation out (have you got the MLX90632?) it would be much appreciated.
image

Originally I was going to ask why my MLX destructor doesn't get called, but it seems the `delete iter` call has been removed for a few years now (git blame said 3yrs ago), as I thought it was my new driver failing to clear up after itself. I also wondered about the calling i2c begin again in the mlx driver, but that's okay to do now iirc.

==========================================

The PR adds the MLX90632, which comes in a few flavours (-Bxy standard / Dxy medical, plus FoV etc), and the medical one adafruit have released first is the -Dxy variant which can be configured for extended range or medical range. I've therefore added two components for the -D variant, one for medical range and one for extended range.
It made sense to also add the -B variant (one extra component) which just uses standard range.

See related component PR https://github.com/adafruit/Wippersnapper_Components/pull/278


tyeth:
Had planned to but didn't for some reason (distracted or not wanting to consume more time on the issue).
I have this morning, and seeing very similar, so submitted the PR to fix the library.

See https://github.com/adafruit/Adafruit_MLX90632/pull/1

brentru:
@tyeth  A new library version has been released, is this ready to be merged in now? I need it for a project.

PR #810, opened 2025-09-05
Offline qmc5883p+AS5600 (after mlx) - with added Vector value support

tyeth:
Adds the QMC5883P which is the first vector based sensor to be added.
WS already has the type recognised, but not implemented in most places, so there is one commit adding that.
While I was there it made sense to add the other vector based types, as I've been keen to do other accelerometers, not just magnetometers. As it stands I haven't got an RGBA sensor that comes to mind so I've left colour out for now.

I also fought clang-format again and finally triumphed. The implicit rules varied between CI and windows, and I'd previously attempted to produce a rule file to resolve it, but was a battle I lost on two occasions. This time Victory!
There's now a ci-wippersnapper-show-config branch in ci-arduino which adds the --show-config flag to run-clang-format.py which I tested with and then reverted from the offline build script.


src/components/i2c/drivers/drvAs5600.h
```diff
@@ -0,0 +1,161 @@
+/*!
+ * @file drvAs5600.h
+ *
+ * Device driver for the AS5600 Magnetic Angle sensor.
+ *
+ * Adafruit invests time and resources providing this open source code,
+ * please support Adafruit and open-source hardware by purchasing
+ * products from Adafruit!
+ *
+ * Copyright (c) Tyeth Gundry 2024 for Adafruit Industries.
```
tyeth, line 10: 2025

brentru (approved):
@tyeth  Overall, this looks great. I read through it.

Issue #732, opened 2025-04-21
SGP30 + Beta 100


brentru:
**Describe the bug**
SGP30 reportedly worked on Beta 99 but failed to send data on Beta 100. Possibly an issue with either initialization or sending data.

***Arduino board***
Feather ESP32-V2, WipperSnapper Beta 100.

**Which components are connected to your device**
Adafruit ESP32 Feather V2 - 8MB Flash + 2 MB PSRAM - STEMMA QT
BME280
SGP30

**Additional context**
https://forums.adafruit.com/viewtopic.php?p=1053487

brentru:
@tyeth  Could you please take a look at reproducing this, this week?

brentru:
Closed via #811.

PR #811, opened 2025-09-12
SGP30 / SGP40 background sampling

mikeysklar:
Add SGP30 / SGP40 background sampling. Requires frequently 1s polling or returns 0 values.

Untested and only a proof of concept for feedback. This adds fastTick() polling to keep the SGPxx sensors producing usable data. Thanks @Timeline for the key insight of 1s polling being needed.

Difference observed between 15min polling versus 1s with the SGP40.

Screenshot 2025-09-12 at 9 39 50 AM

Address issue [732](https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/732) 

Forum issues:

[SGP30 issue with WipperSnapper](https://forums.adafruit.com/viewtopic.php?t=220078)
[SGP40 Air Quality](https://forums.adafruit.com/viewtopic.php?t=218524)

### DIFF
--- a/src/components/i2c/WipperSnapper_I2C.cpp
+++ b/src/components/i2c/WipperSnapper_I2C.cpp
@@ -1032,7 +1032,7 @@ bool WipperSnapper_Component_I2C::initI2CDevice(
     _drivers_out.push_back(_ssd1306);
     WS_DEBUG_PRINTLN("SSD1306 display initialized Successfully!");
   } else {
-    WS_DEBUG_PRINTLN("ERROR: I2C device type not found!")
+    WS_DEBUG_PRINTLN("ERROR: I2C device type not found!");
     _busStatusResponse =
         wippersnapper_i2c_v1_BusResponse_BUS_RESPONSE_UNSUPPORTED_SENSOR;
     return false;
--- a/src/components/i2c/WipperSnapper_I2C.cpp
+++ b/src/components/i2c/WipperSnapper_I2C.cpp
@@ -1349,7 +1349,8 @@ void WipperSnapper_Component_I2C::update() {
 
   // one fast pass for all drivers every update() call
   for (auto *drv : drivers) {
-    if (drv) drv->fastTick();
+    if (drv)
+      drv->fastTick();
   }
 
   long curTime;
--- a/src/components/i2c/WipperSnapper_I2C.cpp
+++ b/src/components/i2c/WipperSnapper_I2C.cpp
@@ -1347,6 +1347,11 @@ void WipperSnapper_Component_I2C::update() {
   msgi2cResponse.which_payload =
       wippersnapper_signal_v1_I2CResponse_resp_i2c_device_event_tag;
 
+  // one fast pass for all drivers every update() call
+  for (auto *drv : drivers) {
+    if (drv) drv->fastTick();
+  }
+
   long curTime;
   bool sensorsReturningFalse = true;
   int retries = 3;
@@ -1354,6 +1359,7 @@ void WipperSnapper_Component_I2C::update() {
   while (sensorsReturningFalse && retries > 0) {
     sensorsReturningFalse = false;
     retries--;
+    curTime = millis();
 
     std::vector<WipperSnapper_I2C_Driver *>::iterator iter, end;
     for (iter = drivers.begin(), end = drivers.end(); iter != end; ++iter) {
--- a/src/components/i2c/WipperSnapper_I2C.cpp
+++ b/src/components/i2c/WipperSnapper_I2C.cpp
@@ -1347,12 +1347,6 @@ void WipperSnapper_Component_I2C::update() {
   msgi2cResponse.which_payload =
       wippersnapper_signal_v1_I2CResponse_resp_i2c_device_event_tag;
 
-  // one fast pass for all drivers every update() call
-  for (auto *drv : drivers) {
-    if (drv)
-      drv->fastTick();
-  }
-
   long curTime;
   bool sensorsReturningFalse = true;
   int retries = 3;
@@ -1364,6 +1358,9 @@ void WipperSnapper_Component_I2C::update() {
 
     std::vector<WipperSnapper_I2C_Driver *>::iterator iter, end;
     for (iter = drivers.begin(), end = drivers.end(); iter != end; ++iter) {
+      // Per-driver fast tick (non-blocking)
+      (*iter)->fastTick();
+
       // Number of events which occurred for this driver
       msgi2cResponse.payload.resp_i2c_device_event.sensor_event_count = 0;
 
--- a/src/components/i2c/drivers/WipperSnapper_I2C_Driver.h
+++ b/src/components/i2c/drivers/WipperSnapper_I2C_Driver.h
@@ -1413,8 +1413,8 @@ protected:
   long _ambientTempFPeriod = 0L; ///< The time period between reading the
                                  ///< ambient temp. (°F) sensor's value.
   long _ambientTempFPeriodPrv =
-      PERIOD_24HRS_AGO_MILLIS;  ///< The time when the ambient temp. (°F) sensor
-                                ///< was last read.
+      PERIOD_24HRS_AGO_MILLIS; ///< The time when the ambient temp. (°F) sensor
+                               ///< was last read.
   long _objectTempFPeriod = 0L; ///< The time period between reading the object
                                 ///< temp. (°F) sensor's value.
   long _objectTempFPeriodPrv =
--- a/src/components/i2c/drivers/WipperSnapper_I2C_Driver.h
+++ b/src/components/i2c/drivers/WipperSnapper_I2C_Driver.h
@@ -1414,8 +1414,8 @@ protected:
   long _ambientTempFPeriod = 0L; ///< The time period between reading the
                                  ///< ambient temp. (°F) sensor's value.
   long _ambientTempFPeriodPrv =
-      PERIOD_24HRS_AGO_MILLIS;  ///< The time when the ambient temp. (°F) sensor
-                                ///< was last read.
+      PERIOD_24HRS_AGO_MILLIS; ///< The time when the ambient temp. (°F) sensor
+                               ///< was last read.
   long _objectTempFPeriod = 0L; ///< The time period between reading the object
                                 ///< temp. (°F) sensor's value.
   long _objectTempFPeriodPrv =
--- a/src/components/i2c/drivers/WipperSnapper_I2C_Driver.h
+++ b/src/components/i2c/drivers/WipperSnapper_I2C_Driver.h
@@ -1414,8 +1414,8 @@ protected:
   long _ambientTempFPeriod = 0L; ///< The time period between reading the
                                  ///< ambient temp. (°F) sensor's value.
   long _ambientTempFPeriodPrv =
-      PERIOD_24HRS_AGO_MILLIS; ///< The time when the ambient temp. (°F) sensor
-                               ///< was last read.
+      PERIOD_24HRS_AGO_MILLIS;  ///< The time when the ambient temp. (°F) sensor
+                                ///< was last read.
   long _objectTempFPeriod = 0L; ///< The time period between reading the object
                                 ///< temp. (°F) sensor's value.
   long _objectTempFPeriodPrv =
--- a/src/components/i2c/drivers/WipperSnapper_I2C_Driver.h
+++ b/src/components/i2c/drivers/WipperSnapper_I2C_Driver.h
@@ -34,7 +34,7 @@ class WipperSnapper_I2C_Driver {
 public:
   /*******************************************************************************/
   /*!
-      @brief    Instanciates an I2C sensor.
+      @brief    Instanctiates an I2C sensor.
       @param    i2c
                 The I2C hardware interface, default is Wire.
       @param    sensorAddress
@@ -52,6 +52,20 @@ public:
   */
   /*******************************************************************************/
   virtual ~WipperSnapper_I2C_Driver() {}
+
+  /*******************************************************************************/
+  /*!
+      @brief    Per-update background hook for drivers that need faster internal
+                sampling than the user publish interval.
+                Default is a no-op; override in concrete drivers (e.g., SGP30/40)
+                to maintain required ~1 Hz reads and accumulate/condition
+                values for later publish.
+      @note     Call site: WipperSnapper_Component_I2C::update() will invoke
+                this once per loop for each driver. Implementations must be
+                non-blocking (do not delay); use millis()-based timing.
+  */
+  /*******************************************************************************/
+  virtual void fastTick() {}
 
   /*******************************************************************************/
   /*!
@@ -1312,7 +1326,7 @@ public:
       @brief    Updates the properties of a proximity sensor.
       @param    period
                 The time interval at which to return new data from the
-                proimity sensor.
+                proximity sensor.
   */
   /*******************************************************************************/
   virtual void updateSensorProximity(float period) {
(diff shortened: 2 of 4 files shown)


ladyada:
brent and tyeth are both out right now but they'll review when return!

mikeysklar:
@tyeth 

Thank you for the larger explanation. Using a polling frequency value in the JSON config makes a lot of sense.

Another issue @Timeline pointed out is deep sleep mode which I believe is still in the works. I bring it up here only that continuous background polling could cancel out the ability to use sleep mode with these sensors.

A few more options to consider as JSON config values:

* ramp up time (eg. 30sec - 1min) 
* # of initial values to throw away
* deep sleep duration (realistic range)

If you are doing hourly air quality samples it doesn't make a lot of sense to keep the controller awake and polling every second.

src/components/i2c/WipperSnapper_I2C.cpp
```diff
@@ -1347,13 +1347,20 @@ void WipperSnapper_Component_I2C::update() {
   msgi2cResponse.which_payload =
       wippersnapper_signal_v1_I2CResponse_resp_i2c_device_event_tag;
 
+  // one fast pass for all drivers every update() call
+  for (auto *drv : drivers) {
+    if (drv)
+      drv->fastTick();
+  }
+
```
brentru, line 1355: Do not use `auto` here or create a loop. It is redundant and adds a performance overhead for what should be a tight loop.

Instead, try calling `drv->fastTick()` when we are looping through the vectors below. Since `fastTick` is `virtual`, and the vect. of drivers exists, you dont need to check for `nullptr` either.

src/components/i2c/WipperSnapper_I2C.cpp
```diff
@@ -1347,13 +1347,20 @@ void WipperSnapper_Component_I2C::update() {
...
   long curTime;
   bool sensorsReturningFalse = true;
   int retries = 3;
 
   while (sensorsReturningFalse && retries > 0) {
     sensorsReturningFalse = false;
     retries--;
+    curTime = millis();
```
brentru, line 1363: Why would `curTime` be modified here?

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SGP40.h
```diff
@@ -82,13 +109,66 @@ class WipperSnapper_I2C_Driver_SGP40 : public WipperSnapper_I2C_Driver {
...
+  */
+  /*******************************************************************************/
+  void fastTick() override {
+    if (!_sgp40)
+      return;
+    if (!vocEnabled())
+      return;
+
+    uint32_t now = millis();
+    if (now - _lastFastMs >= 1000) {
```
brentru, line 140: `1000` is a magic number, please `#define` what it is at the top of this header.

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SGP40.h
```diff
@@ -82,13 +109,66 @@ class WipperSnapper_I2C_Driver_SGP40 : public WipperSnapper_I2C_Driver {
...
+
 protected:
-  Adafruit_SGP40 *_sgp40; ///< SEN5X driver object
+  Adafruit_SGP40 *_sgp40; ///< SGP40
+
+  /** Millis timestamp of last 1 Hz background read. */
+  uint32_t _lastFastMs = 0;
+
+  /** Number of samples accumulated since last publish. */
+  uint32_t _n = 0;
```
brentru, line 155: `_num_samples`

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SGP40.h
```diff
@@ -50,12 +65,15 @@ class WipperSnapper_I2C_Driver_SGP40 : public WipperSnapper_I2C_Driver {
...
+      _sgp40 = nullptr;
       return false;
     }
-
-    // TODO: update to use setCalibration() and pass in temp/humidity
-
+    _lastFastMs = millis();
+    _n = 0;
+    _vocSum = 0.0f;
+    _rawSum = 0;
```
brentru, line 76: Move these to the constructor

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SGP40.h
```diff
@@ -40,6 +40,21 @@ class WipperSnapper_I2C_Driver_SGP40 : public WipperSnapper_I2C_Driver {
       : WipperSnapper_I2C_Driver(i2c, sensorAddress) {
     _i2c = i2c;
     _sensorAddress = sensorAddress;
+    _sgp40 = nullptr;
```
brentru, line 43: Should be with the object ptr declaration under `protected`, not in ctor

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SGP30.h
```diff
@@ -44,27 +48,90 @@ class WipperSnapper_I2C_Driver_SGP30 : public WipperSnapper_I2C_Driver {
...
 
 protected:
-  Adafruit_SGP30 *_sgp30; ///< Pointer to SGP30 temperature sensor object
+  Adafruit_SGP30 *_sgp30; ///< Pointer to SGP30 sensor object
+
+  /** Millis timestamp of last 1 Hz background read. */
+  uint32_t _lastFastMs = 0;
+
+  /** Number of samples accumulated since last publish. */
+  uint32_t _n = 0;
```
brentru, line 117: Please change to `_sample_num` or something more descriptive

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SGP30.h
```diff
@@ -44,27 +48,90 @@ class WipperSnapper_I2C_Driver_SGP30 : public WipperSnapper_I2C_Driver {
...
+      return true;
+    }
+    if (_sgp30->IAQmeasure()) {
+      senseEvent->tvoc = (uint16_t)_sgp30->TVOC;
+      return true;
+    }
+    return false;
+  }
+
+  void fastTick() override {
```
brentru, line 96: Please add a comment docstring above this func.

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SGP30.h
```diff
@@ -44,27 +48,90 @@ class WipperSnapper_I2C_Driver_SGP30 : public WipperSnapper_I2C_Driver {
...
+    if (!iaqEnabled())
+      return; // nothing enabled, save cycles
+    uint32_t now = millis();
+    if (now - _lastFastMs >= 1000) { // ~1 Hz cadence
```
brentru, line 100: `1000` is a magic number, please #define it.

brentru (changes requested):
Overall, this makes sense to me as an implementation for a few drivers in WipperSnapper beta. 

We're currently implementing the new WipperSnapper API (v2) and have time to consider alternatives for _specific_ things, like background sampling, into the new API design.

@tyeth  If you also approve of this design, I wouldn't be opposed to merging it in. We can talk about it more tomorrow or in this PR.

@mikeysklar  I've requested a few design/syntax changes. Please hold off on these until Tyeth also approves.

tyeth:
> Another issue @Timeline pointed out is deep sleep mode which I believe is
still in the works. I bring it up here only that continuous background
polling could cancel out the ability to use sleep mode with these sensors.
That was my nod to Low Power mode, a.k.a. deep sleep support, which will be
v2 of WipperSnapper (both Offline and Online mode - currently only Offline
mode
<https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/tree/offline-mode/src>
uses v2 codebase).

> A few more options to consider as JSON config values:
>
> * ramp up time (eg. 30sec - 1min)
> * number of initial values to throw away
> * deep sleep duration (realistic range)
Yeah I like these. Less so the deep sleep duration, or at least it then
prompted a bit of confusion. I initially imagined deep-sleep-duration as
unnecessary because it's just the normal polling period people select, but
then realise maybe you're referring to the usable range for low power /
sleep usage which gets more complicated/useful.
 I wonder if ramp up time is the time until good data, or just any
semi-usable data, and whether users would prefer to still report it
(advanced config option?) during that period. Or is ramp up simply until
any data that shouldnt be 0/NaN/etc, like 15s for SGP30 mentioned later,
not until the internal algorithm is happy (do we need another field related
to stability/quality of data after x amount of time).

> If you are doing hourly air quality samples it doesn't make a lot of
sense to keep the controller awake and polling every second.
Agreed, the one issue in the back of my mind is the air quality sensors (or
any type) that have an inbuilt algorithm, rather than raw data (Sensirion
VOC/IAQ index plus some others). They require the baseline
values/coefficients to be trained for ages and regularly stored (ideally
after each reading) and even then are [only valid for the device being
asleep about 10mins max](
https://github.com/adafruit/Adafruit_SGP40/blob/741a139e73e95273d6b5ac074c780aa74597bb9e/src/sensirion_voc_algorithm.h#L133C51-L134C52)
(otherwise [the full baseline compensation run must be performed at next
wake/boot](
https://github.com/adafruit/Adafruit_SGP40/blob/741a139e73e95273d6b5ac074c780aa74597bb9e/src/sensirion_voc_algorithm.h#L122-L123)
).
 I guess that 10minute limit would be the "max deep sleep duration",
assuming the "ramp up time" was 3hours and not performed on wake (so not
after a <10min low power sleep), and we probably want to do the fastest
tick possible when throwing away values (rather than waiting for poll
period before broadcasting first value on a fresh boot - maybe not for a
waking boot).
 In the back of my mind I'm thinking that throw-away-values might fall
apart quickly if the sensor has a slow internal poll period (calling that
fastTicks in this PR), like the 5seconds minimum poll period on one air one
(I forget which - SCD?). I guess we'd just drop the first value and no
others in that situation (it's real CO2 not algorithm based so 30s ramp up
IIRC).
 I'm not sure in the air sensor situation what a user would prefer/expect,
personally I want correct data more than power saving, so ideally I'd see
the first boot not put the sensor to sleep until at least 3hours had
elapsed so the baseline could be saved. I guess it could sleep the mcu
almost immediately as long as the sensor remains powered, and it internally
keeps updating the algorithm without being read (assuming in appropriate
mode) like the SEN4x/5x do (internally updating the algo at 1Hz IIRC).

Forgetting those Sensirion ones for a second, and still looking at VOC,
then we only sell the ENS160 (<3 minutes warm-up <1 hour start-up), and the
BME68x (30mins each startup).
Switching to look at CO2, or better yet eCO2 (estimated/equivalent), then
we only really have the ENS160 + Sensirion SGP and SCD series. Taking the
SGP30 we could benefit from dropping the first 15seconds of data
<https://cdn-learn.adafruit.com/assets/assets/000/050/058/original/Sensirion_Gas_Sensors_SGP30_Datasheet_EN.pdf#:~:text=first%2015s>
(ramp
up time?) before the algo wakes up, use an internal poll of 1second (which
the algo prefers
<https://cdn-learn.adafruit.com/assets/assets/000/050/058/original/Sensirion_Gas_Sensors_SGP30_Datasheet_EN.pdf#:~:text=The%20on>,
but
doesn't work well for low power). It suggests the baseline is useful for up
to a week
<https://files.seeedstudio.com/wiki/Grove-VOC_and_eCO2_Gas_Sensor-SGP30/res/Sensirion_Gas_Sensors_SGP30_Driver-Integration-Guide_HW_I2C.pdf#:~:text=one%20week>,
and initially only valid after 12hours
<https://files.seeedstudio.com/wiki/Grove-VOC_and_eCO2_Gas_Sensor-SGP30/res/Sensirion_Gas_Sensors_SGP30_Driver-Integration-Guide_HW_I2C.pdf#:~:text=12%20hours>.
So how would one expect that to play out?

src/components/i2c/drivers/WipperSnapper_I2C_Driver.h
```diff
@@ -34,7 +34,7 @@ class WipperSnapper_I2C_Driver {
 public:
   /*******************************************************************************/
   /*!
-      @brief    Instanciates an I2C sensor.
+      @brief    Instanctiates an I2C sensor.
```
tyeth, line 37: no middle c

brentru (changes requested):
@mikeysklar  Tyeth and I discussed this fix this morning.  You are on the right track, but we'd like to remove the averaging code.

For the SGP30,
* Remove all the code related to averaging values within the driver.
* `fastTick` is OK but should only just read the driver. It should set a boolean flag if `IAQmeasure` succeeded. 
* `getEventX()` methods should first check the validity of the previous read (return value of `IAQmeasure()`) from fastTick, and fill the Event if the read was OK.

For the SGP40,
* Remove all the code related to averaging values within the driver.
* Add `uint16_t _rawValue` and `int32_t _vocIdx` as `protected:` members
* `fastTick` should just poll for raw value and voc inx, and save those into `_rawValue` and `_vocIdx`.
* Within `getEventX()`, pack `_rawValue` or `_vocIdx` into an Event.

I added two comments below to reflect these changes. lmk if you need more info/have more questions, and thanks for taking a look at this. If you'd like Tyeth to take this over as well, let him know.

brentru:
@mikeysklar  Sorry, what 1Hz timer?

brentru:
Oh, yeah, that should go back in

mikeysklar:
I've been testing the SGP40 and it has been running smoothly with different update periods 1s - 10min. I believe I got all the suggested changes in. I ordered a SGP30 for testing which I can confirm functionality this week.

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SGP30.h
```diff
@@ -24,16 +26,19 @@ class WipperSnapper_I2C_Driver_SGP30 : public WipperSnapper_I2C_Driver {
...
 
   /*******************************************************************************/
   /*!
       @brief    Destructor for an SGP30 sensor.
   */
   /*******************************************************************************/
-  ~WipperSnapper_I2C_Driver_SGP30() {
-    // Called when a SGP30 component is deleted.
-    delete _sgp30;
+  ~WipperSnapper_I2C_Driver_SGP30() override {
```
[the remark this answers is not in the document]
mikeysklar: thx. removed override on SGP30 / SGP40 destructors.

src/components/i2c/drivers/WipperSnapper_I2C_Driver_SGP30.h
```diff
@@ -44,27 +49,97 @@ class WipperSnapper_I2C_Driver_SGP30 : public WipperSnapper_I2C_Driver {
...
+    if (!_sgp30)
+      return false;
+    if (!_did_measure) {
+      _did_measure = _sgp30->IAQmeasure();
+    }
+    if (!_did_measure)
+      return false;
+
+    senseEvent->eCO2 = (uint16_t)_sgp30->eCO2;
+    _did_measure = false; // consume cached reading
```
tyeth, line 73: the `// consume cached reading` comments have me confused.
The flag being false doesn't result in the next getEvent call (TVOC for example) using the cached read/data, instead it will read new data or fail.

Same in the other getEvent method.

This (SGP30) behaves slightly differently in this driver than the SGP40, in that it will do fresh readings for any getEvents after the first one. Whereas the SGP40 wrapper relies on fastTick soley for fetching data, resulting in the data for getEvent (and subsequent / other event types) using data fetching in the same time period.

Not too fussed, mainly the cached comment got me 😀

tyeth (changes requested):

mikeysklar:
SGP30 arrived today so I'll try running both SGP30 and SGP40 with the latest artifacts and report back.

tyeth (approved):

brentru (approved):
@mikeysklar LGTM! Thank you for adding this PR

@tyeth  Please pull this in when you are ready.

Issue #815, opened 2025-09-18
Fruit Jam install issues


tyeth:
Firstly the install uf2 didn't download, but was in release. I've fixed the boards definition (installBoardName) so that now works.

Secondly, I've got circuitpython installed, so now installing WipperSnapper I see the yellow lights, but no drive appears.
The serial console shows a repeating message:
```
Fatal Error: Halted execution!
ERROR Initializing Filesystem
```

I'm assuming initially that I need to format the device first, using the universal nuke from pimoroni, because the official one doesn't erase all the memory (see https://github.com/raspberrypi/pico-examples/issues/642)

-EDIT- Assumption was wrong, same error after erasing. Looks like our fruit jam firmware is being built with zero bytes for the user file system partition. Updated ci-arduino repo in branches/ci-wippersnapper-fruitjam-fs-size
Test PR in https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/814

brentru:
@tyeth  Is this issue still valid, or stale?

PR #816, opened 2025-09-18
Fruit Jam fixes + tweaks (plus airlift FW reported in bootlog)

tyeth:
Updates the platformIO entries to inherit from the main rp2 environment definition (setting arduino framework etc).
Fixes the board defines to match arduino-cli (`ARDUINO_` prefix).
Adds WiFiNINA clone step to CI.
Brings in https://github.com/adafruit/ci-arduino/pull/219 for the user file system partition > 0bytes, fixing #815 

Also adds firmware version reporting in the wipper_boot_log.txt file for AirLift boards.

### DIFF
--- a/src/Wippersnapper.h
+++ b/src/Wippersnapper.h
@@ -372,10 +372,11 @@ public:
   ws_uart *_uartComponent;        ///< Instance of UART class
 
   // TODO: does this really need to be global?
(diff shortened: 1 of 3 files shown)


tyeth (commented):
Looks good. Whitespace commit cleaned up. Tested locally at each stage, will confirm on CI before release.

PR #826, opened 2025-10-10
Feather v2esp32 eink no net interface

tyeth:
Feather ESP32 V2 was showing no network interface when pinged-out / submitting displayResponse from e-ink featherwing after/during initial add.

### DIFF
--- a/src/components/display/controller.h
+++ b/src/components/display/controller.h
@@ -17,8 +17,8 @@
 #include "Wippersnapper.h"
 #include "hardware.h"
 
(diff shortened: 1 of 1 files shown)


brentru (approved):
I do not have this board but the content of the PR looks like it resolves the bug you were having, this was a mistake in my code.

tyeth:
Closing as has been incorporated in the display fixes PR #827, along with in the ESP32v2 linking issue #830 (PR #829) which alters the extern.

PR #827, opened 2025-10-14
Fixes - Display Component

brentru:
* Adds: MagTag displays are now auto-detected as SSD1680 or ILI0373 and configured with the appropriate mode and driver. (related - https://github.com/adafruit/Wippersnapper_Components/pull/287)
* Refactors the display driver selection mechanism to use an enum (`wippersnapper_display_v1_DisplayDriver`) instead of a string identifier.
* Fix - Display drivers clear their screen when their component is deleted, and then free resources.
* Fix - `dispDrvThinkInkGrayscale4T5` now draws and updates the status bar
* Fix - `dispDrvThinkInkGrayscale4T5` adheres to text size/resizing rules
* Fix - Fixed instance naming (`Ws` to `WS`) and corrected class forward declaration to use `Wippersnapper` on the the controller header.
* Fix - Removed redundant watchdog timer (`WS.feedWDT()`) calls from the display controller, priority given towards `NetFSM` call which already updates the WDT.

### DIFF
--- a/src/components/display/hardware.cpp
+++ b/src/components/display/hardware.cpp
@@ -405,22 +405,26 @@ bool DisplayHardware::detect_ssd1680(uint8_t cs, uint8_t dc, uint8_t rst) {
   pinMode(SCK, OUTPUT);
   pinMode(cs, OUTPUT);
   pinMode(dc, OUTPUT);
-  pinMode(rst, OUTPUT);
+  if (rst >= 0)
+    pinMode(rst, OUTPUT);
 
   // Reset the display
   digitalWrite(cs, HIGH);
-  digitalWrite(rst, HIGH);
-  delay(10);
-  digitalWrite(rst, LOW);
-  delay(10);
-  digitalWrite(rst, HIGH);
-  delay(200);
+  if (rst >= 0) {
+    digitalWrite(rst, HIGH);
+    delay(10);
+    digitalWrite(rst, LOW);
+    delay(10);
+    digitalWrite(rst, HIGH);
+    delay(200);
+  }
 
   // Begin transaction by pulling cs and dc LOW
   digitalWrite(cs, LOW);
   digitalWrite(dc, LOW);
   digitalWrite(MOSI, LOW);
-  digitalWrite(rst, HIGH);
+  if (rst >= 0)
+    digitalWrite(rst, HIGH);
   digitalWrite(SCK, LOW);
 
   // Write to read register 0x71


brentru:
@tyeth  Re-requesting a review, a lot of fixes have been made to the display components and I want to ensure we catch them all.

I've tested the display detection code on MagTag ILI and MagTag SSD.

src/components/display/hardware.cpp
```diff
@@ -185,12 +185,6 @@ bool DisplayHardware::beginEPD(
     return false;
   }
 
-  // Validate mode is a correct EPD mode
-  if (config->mode == wippersnapper_display_v1_EPDMode_EPD_MODE_UNSPECIFIED) {
```
tyeth: Good to still warn maybe [but not return], so we notice any invalid configs more easily

src/components/display/hardware.cpp
```diff
@@ -238,8 +247,7 @@ bool DisplayHardware::beginEPD(
     _drvDisp = nullptr;
     return false;
   }
-  WS_DEBUG_PRINTLN("[display] Text Magnification: ");
```
tyeth: Is there an unpushed commit here? It was somewhat useful logging [maybe in px equivalent]. Half expected to see something associated with updating the font size and restoring the previous message assuming it's a display update of just font size.

src/components/display/hardware.cpp
brentru: After the MagTag auto-detection step, I feel like we could validate the mode **and** return. Its unset for magtag until there, though. I will add this in a subsequent commit.

src/components/display/hardware.cpp
```diff
@@ -388,20 +396,28 @@ void DisplayHardware::writeMessage(const char *message) {
...
   // Configure SPI pins to bit-bang
   pinMode(MOSI, OUTPUT);
   pinMode(SCK, OUTPUT);
   pinMode(cs, OUTPUT);
   pinMode(dc, OUTPUT);
   pinMode(rst, OUTPUT);
 
+  // Reset the display
+  digitalWrite(cs, HIGH);
+  digitalWrite(rst, HIGH);
```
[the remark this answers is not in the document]
brentru: Why would I want to guard this? Would I only guard this for `rst` pin?

brentru:
@tyeth  I addressed the potentially undefined behavior with the reset pin above. Please take a look.

PR #828, opened 2025-10-14
fix(e-ink): use correct degree symbols

tyeth:
Fixes degree symbol from a mid-height square to a supertext circle (the degree symbol).
Does look a bit bold, but definitely more appropriate.
image

More obvious with decimal places too:
image


tyeth:
> Thanks for fixing this. Would we need to do this for TFT's too?

Tested and updated! Originally I checked the greenhouse learn guide that showed an external TFT looking fine so left them.

PR #829, opened 2025-10-15
Featherv2 eink linking

tyeth:
Draft PR to be able to leave comments on the changes

### DIFF
--- a/src/Wippersnapper.cpp
+++ b/src/Wippersnapper.cpp
@@ -32,8 +32,11 @@
  */
 
 #include "Wippersnapper.h"
+#include "Wippersnapper_Networking.h"
 
-Wippersnapper WS;
+// Define the global WS instance as the platform-specific derived class
+// This ensures virtual methods like _connect() route to the correct implementation
+Wippersnapper_WiFi WS;
 
 Wippersnapper::Wippersnapper() {
   _mqtt = 0; // MQTT Client object
--- a/src/Wippersnapper.h
+++ b/src/Wippersnapper.h
@@ -502,8 +502,10 @@ protected:
       _outgoingSignalMsg; /*!< Outgoing signal message from device */
 };
 
-// Forward declaration - actual type is platform-specific (Wippersnapper_WiFi)
-// but we declare as base class here to avoid circular dependencies
-extern Wippersnapper WS; ///< Global member variable for callbacks
+// Include networking to get the platform-specific Wippersnapper_WiFi typedef
+#include "Wippersnapper_Networking.h"
+
+// Global WS instance - defined as platform-specific type in Wippersnapper.cpp
+extern Wippersnapper_WiFi WS; ///< Global member variable for callbacks
 
 #endif // ADAFRUIT_WIPPERSNAPPER_H
--- a/src/Wippersnapper.h
+++ b/src/Wippersnapper.h
@@ -501,6 +501,9 @@ protected:
   wippersnapper_signal_v1_CreateSignalRequest
       _outgoingSignalMsg; /*!< Outgoing signal message from device */
 };
+
+// Forward declaration - actual type is platform-specific (Wippersnapper_WiFi)
+// but we declare as base class here to avoid circular dependencies
 extern Wippersnapper WS; ///< Global member variable for callbacks
 
 #endif // ADAFRUIT_WIPPERSNAPPER_H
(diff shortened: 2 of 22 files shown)


src/Wippersnapper.h
```diff
@@ -501,6 +501,11 @@ class Wippersnapper {
   wippersnapper_signal_v1_CreateSignalRequest
       _outgoingSignalMsg; /*!< Outgoing signal message from device */
 };
-extern Wippersnapper WS; ///< Global member variable for callbacks
+
+// Include networking to get the platform-specific Wippersnapper_WiFi typedef
+#include "Wippersnapper_Networking.h"
+
+// // Global WS instance - defined as platform-specific type in Wippersnapper.cpp
+// extern Wippersnapper_WiFi WS; ///< Global member variable for callbacks
```
tyeth: Drop this, but leave comment: global WS now is a Wippersnapper_WiFi and comes via Wippersnapper_Networking.h, but needs to be extern'd in each cpp file.

src/network_interfaces/Wippersnapper_AIRLIFT.h
tyeth: needs fixing up like esp32 (other Arch builds not tested)

src/Wippersnapper.h
tyeth: resolved in #851

src/Wippersnapper.cpp
```diff
@@ -32,8 +32,11 @@
  */
 
 #include "Wippersnapper.h"
+#include "Wippersnapper_Networking.h"
```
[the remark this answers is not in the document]
tyeth: resolved in #851

src/network_interfaces/Wippersnapper_AIRLIFT.h
tyeth: resolved in #851

tyeth-ai-assisted:
Another manifestation of this base-global linking issue, fresh from HIL — and a data point that it isn't limited to the older-gcc ESP32.

On an **ESP32-S3** (QT Py N4R2) running the telemetry PR (adafruit/Adafruit_Wippersnapper_Arduino#946), a `ws.telemetry.Add{type: TM_RSSI}` makes the device read RSSI via `Ws.getRSSI()` — which hits the **base `wippersnapper::getRSSI()` stub**:
```
serial:  Wiippersnapper_V2::getRSSI
serial:  ERROR: Please define a network interface!
protomq: {"telemetry":{"event":{"type":"TM_RSSI","value":{"type":"T_RAW","floatValue":0}}}}
```
WiFi/MQTT connect fine (via the `wipper` = `ws_adapter_wifi` object), but the component-facing global `Ws` is the base type, so its virtual network methods resolve to the stubs — the "two different WS globals" you called out in adafruit/Adafruit_Wippersnapper_Arduino#830. So this happens on S3 too, not just the old-gcc ESP32.

Switching the global to the `Wippersnapper_WiFi` type (this PR) should fix the RSSI telemetry value and the display's RSSI in one go. The `TM_RSSI` telemetry component itself is correct — only the value source is affected. Happy to HIL-verify RSSI over protomq once this lands.

tyeth-ai-assisted:
Design note for this fix given v2's adapter split (so it applies to **both** networked and offline builds):

`ws_platforms.h` typedefs the adapter per target — `ws_adapter_wifi` (`esp32_wifi` / `esp8266_wifi` / `pico_wifi` / `airlift_wifi` / `ninafw_wifi`) for networked builds, and `ws_adapter_offline` (`pico_offline`) for offline builds — and the sketch instantiates one (`ws_adapter_wifi wipper;` vs `ws_adapter_offline wipper;`). All inherit `wippersnapper`.

So the component-facing global can't be hardcoded to a fixed `Wippersnapper_WiFi` type — that would break the offline builds. A polymorphic binding handles both: e.g. a `wippersnapper *Ws` (or reference) pointed at the active adapter `wipper` at startup, with `Ws.` → `Ws->` at the call sites (the "small refactor" from #830). Then `getRSSI()` (and the other virtual network methods) dispatch to whichever adapter the build selected — including `pico_offline::getRSSI()`, which already returns its not-supported sentinel. That keeps the fix correct for wifi and offline without per-build type juggling.

tyeth:
@brentru take a look at this, it's come up before, and needs resolving (to address telemetry rssi).

PR #835, opened 2025-11-07
Add Espressif C5 devkit

tyeth:
This adds the C5 devkit, with 2.4GHz and 5GHz WiFi 6 support.

The C5 flashed fine with platformio but had a similar issue flashing CI assets to the C6 generated when esptool wasn't pinned. Also now that we aren't pinned for lvgl, it felt like a good time to revisit using the latest BSP and ESPtool, previously held up by the flash args issue for the C6.
Our merged bin was suggested (by Jason8266) as having the wrong flash args causing the ets loader issue when using latest esptool, and looking at the arduino-IDE output for the C5 instead it uses "keep" as the args after the initial setup of the bootloader and app partition. I thought I'd tried that before with the C6 although I might have had other issues too then.

There was a mismatch with boot mode DIO and flash mode QIO which the c5 uses, compared to our fixed merge-bin command that uses a single argument. The arduino IDE generates bootloader.bin using `esptool elf2image` from a qio_80m.elf bootloader, but using DIO flash mode arg to that command, even though flash mode was QIO from the FQBN string - checking deeper it needs DIO for boot flash mode)

I've tested the Feather C6 and Espressif C5 devkit, using the combined.bin from CI assets after an erase. (Both now use esptool 5.1.0 instead of 4.6, with "keep" for the flash args to merge-bin, and every esp32 board is moved to BSP 3.3.3 instead of ~3.1)

Boards Repo PR:
https://github.com/adafruit/Wippersnapper_Boards/pull/222

CI Arduino Branch:
https://github.com/adafruit/ci-arduino/tree/ci-wippersnapper-esp333 [Pointed at adafruit/arduino-esp32 BSP3.3.3]

### DIFF
--- a/.github/workflows/build-clang-doxy.yml
+++ b/.github/workflows/build-clang-doxy.yml
@@ -418,7 +418,6 @@ jobs:
         with:
           repository: adafruit/Wippersnapper_Boards
           path: ws-boards
-          ref: add-c5devkit
       - name: Install CI-Arduino
         run: bash ci/actions_install.sh
       - name: Install extra Arduino libraries
@@ -957,7 +956,6 @@ jobs:
         with:
           repository: adafruit/Wippersnapper_Boards
           path: ws-boards
-          ref: add-c5devkit
       - name: Install CI-Arduino
         run: bash ci/actions_install.sh
       - name: Install extra Arduino libraries


.github/workflows/build-clang-doxy.yml
```diff
@@ -72,7 +72,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           repository: adafruit/ci-arduino
-          ref: ci-wippersnapper
+          ref: ci-wippersnapper-esp333
```
brentru: If we are going to use this branch, across this ci, shouldn't we merge the `esp333` work into `ci-wippersnapper` and use that instead?

.github/workflows/build-clang-doxy.yml
```diff
@@ -459,6 +465,7 @@ jobs:
       - name: Rename build artifacts to reflect the platform name
         run: |
           mv examples/Wippersnapper_demo/build/*/Wippersnapper_demo.ino.bin wippersnapper.${{ matrix.arduino-platform }}.littlefs.${{ env.WS_VERSION }}.bin
+          # mv examples/Wippersnapper_demo/build/*/Wippersnapper_demo.ino.merged.bin wippersnapper.${{ matrix.arduino-platform }}.littlefs.${{ env.WS_VERSION }}.auto-merge.bin
```
brentru: If commenting out, please add another commented line explaining why you are commenting this out

.github/workflows/build-clang-doxy.yml
```diff
@@ -939,13 +949,14 @@ jobs:
...
           repository: adafruit/ci-arduino
-          ref: ci-wippersnapper
+          ref: ci-wippersnapper-esp333
           path: ci
       - name: Checkout Board Definitions
         uses: actions/checkout@v4
         with:
           repository: adafruit/Wippersnapper_Boards
           path: ws-boards
+          ref: add-c5devkit
```
brentru: Make sure to change this when the ws-boards branch is merged please!

brentru (changes requested):

tyeth:
Just to add that this was verified again with the Adafruit IO installer, running locally on my machine, both the offline bin and online installer methods. Then also MQTT connection to that local VM.
That required a release to work against (my fork), which then used the wrong jobs (release ci yaml depends on adafruit/main ci steps), so the C5 release asset in my fork was manually replaced from this PR CI job for testing.

It depends on the baudrate field being added to the boards repo ([Wippersnapper_Boards/#222](https://github.com/adafruit/Wippersnapper_Boards/pull/222)), along with the installer PR for the baud rate overrides, and latest esptool-js ([io-rails/#853](https://github.com/AdafruitInternalDev/io-rails/pull/853)). They have been tested locally by myself and briefly by Loren, but would benefit from a run through on staging (as well as a quick smoke test of each platform)

PR #843, opened 2025-11-26
Fix DotStar color format for FunHouse status pixels

### LINKED ISSUE #842: DotStar colours for FunHouse are incorrect when used as Status LED, shows Blue on successful connect.
Needs the colour order setting up correctly for the status pixels. Maybe worth a ws_boards define, or just #ifdef

tyeth:
closes #842


brentru (approved):
Thanks for fixing, LGTM 👍

PR #851, opened 2025-12-17
Add uc8253 3.7" mono 

tyeth:
https://www.adafruit.com/product/6395

https://learn.adafruit.com/bare-e-ink-displays-crash-course/3-7-416x240-monochrome-eink-uc8253

### DIFF
--- a/src/Wippersnapper.cpp
+++ b/src/Wippersnapper.cpp
@@ -33,10 +33,9 @@
 
 #include "Wippersnapper.h"
 
-// Define the global WS instance as the platform-specific derived class
-// This ensures virtual methods like _connect() route to the correct
-// implementation
-Wippersnapper_WiFi WS;
+// Define the global WS instance as the platform-specific derived class,
+// ensuring virtual methods [_connect()] route to the correct implementation
+Wippersnapper_WiFi WS; //!< Global instance of Wippersnapper_WiFi
 
 Wippersnapper::Wippersnapper() {
   _mqtt = 0; // MQTT Client object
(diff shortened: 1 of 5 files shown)


tyeth (commented):
Requires protobuf (https://github.com/adafruit/Wippersnapper_Protobuf/pull/157) and io-rails change (https://github.com/AdafruitInternalDev/io-node/pull/198) to bring in the extra screen support.
Will be the same for the next few screens (two new drivers, think the third may reuse an existing one)

src/components/display/controller.cpp
```diff
@@ -83,15 +83,15 @@ bool DisplayController::Handle_Display_AddOrReplace(
     return false;
   }
 
-  WS.runNetFSM();
+  WS.pingBroker();
```
brentru: Why are we changing from `runNetFSM()` to `pingBroker()` here? 

`runNetFSM()` already pings

src/components/display/controller.cpp
```diff
@@ -83,15 +83,15 @@ bool DisplayController::Handle_Display_AddOrReplace(
...
-  WS.runNetFSM();
+
+  WS.pingBroker();
   display->drawStatusBar(WS._config.aio_user);
-  WS.runNetFSM();
 
   _hw_instances.push_back(display); // Store the display instance
   WS_DEBUG_PRINTLN("[display] Display added or replaced successfully!");
-  WS.runNetFSM();
+  WS.pingBroker();
```
brentru: Same question here

src/components/display/controller.cpp
tyeth: That's what I thought, but looks like no it doesn't in main / v1.

brentru: Oh, okay! In that case, run the FSM to ensure we are connected _then_ ping the broker.

PR #863, opened 2026-01-26
fix(ci): Replace curl with gh command to fetch release info

tyeth:
Switches to using the built in github command line interface `gh` to fetch release info, as it's the most common failure. 
In theory it should be more resilient, and now that I've linked the token it shouldn't be rate limited as easily in the first place.


tyeth:
@brentru this is ready for review, should make the tinyuf2 release fetches less flakey.

Issue #431, opened 2023-05-17
ESP32-V2 Deep Sleep between Updates?


DJDevon3:
**Is your feature request related to a problem? Please describe.**
Would like to have WipperSnapper installed outside with a solar setup (still within wifi range) for a wind gauge, anemometer, and rain gauge (sparkfun weather kit).

**Describe the solution you'd like**
Deep sleep to conserve at much battery life as possible between updates using Espressif's deep sleep capability. I typically poll every 15 minutes.

**Describe alternatives you've considered**
I could go with an ESP32-S2, Circuit Python, MQTT, etc... but I like the simplicity of WipperSnapper and easily installed updates.

**Additional context**
I'll be working with the Adafruit solar charger either product 390 or 4755. 

I have no idea if the Feather ESP32-V2 is already doing deep sleep or not with WipperSnapper.  If yes, then disregard this feature request.

brentru:
Related: https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/191

Issue #837, opened 2025-11-07
Implement Sleep Modes


brentru:
Supported MCUs for initial launch:
- ESP32-x
- RP2350

Sleep Modes for initial launch:
- [ ] Deep Sleep
- [ ] Light Sleep

Issue #327, opened 2022-09-15
Support ESP32 Light Sleep Mode


brentru:
Add support for ESP32 Sleep Mode, light sleep. We may want to use the timer as a primary source for wakeup.

nanogear:
Do ESP32 deep sleep works with wippersnapper?

brentru:
@nanogear  Not currently, this issue is for tracking the request to add sleep mode. It will be added this year.

coolate:
It would be awesome to add a way to wake on the wake-up pin, especially from deep sleep and on other boards that support it too! As a use case, I only want it to wake and and alert me when the switch is switched.

PR #867, opened 2026-02-12
v2 - Add Sleep/Low Power Modes

brentru:
This is a very large PR that implements sleep and low power modes into WipperSnapper 1.0 (v2 API)

WipperSnapper API PR: https://github.com/adafruit/Wippersnapper_Protobuf/pull/159

Resolves: https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/431, https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/191, https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/327, https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/837

### 🆕  Feature - Sleep / Low Power Mode

**Features**
 * Sleep Modes:
   * Deep sleep (ESP32 only)
   * Light sleep (ESP32 + RP2350)
     * Note: RP2350 will use Dormant state. While this is technically similar to ESP32 deep sleep (near zero sleep mode), memory is not powered off, so behaviorally this implementation is more similar to light sleep. 
* Wake Sources:
  * Timer-based: RTC timer on ESP32, AON timer on RP2350
  * GPIO, EXT0 
* Sleep Persistence (memory kept alive during cycles):
  * Software RTC counter persisted (in NVS (ESP32) / tracked via ws_wdt (RP2350)), hardware RTC does not need to persist across cycles
  * 🆕  feature: `.log` filename is persisted across sleep cycles
  * Basic sleep cycle counter is persisted across runs
* Network Handling:
  * Implements WiFi state management for sleep/wake states
  *  Handles NetFSM failure by re-entering sleep (preventing infinite boot loops)
* Power Management Handling:
  * Disable external components (TFT, NeoPixel power, I2C power), SD card shutdown before sleep
  *  ☝️  re-enabled upon Wake
* Special "Wake Enable" Pin for Offline Mode:
  * Override to break out of sleep loop
  * This override is implemented in `sleep.proto`  for Online mode.

**Files**
* New `sleep` component - `src/components/sleep`
  * `controller` - message routing, handles sleep and wake states
  * `hardware` - microcontroller/platform-specific API access for sleep and other peripherals required for sleep mode (i.e: nvs, rtc memory)
  * `model` - Protobuf API for compiled `sleep.proto`

#### 🤓 PR Review Areas
@tyeth 👇 
Entire Files:
- src/components/sleep/controller.cpp
- src/components/sleep/hardware.cpp
- src/components/sleep/model.cpp
Critical Junctures:
- src/helpers/ws_wdt.{h,cpp} - RP2350 sleep implementation
- Sleep handling in main wippersnapper.cpp (new sleep loop vs normal loop)
- src/components/sleep/controller.cpp:132-175 - `WakeFromLightSleep()`
  sequence
- src/components/sleep/controller.cpp:356-416 - `StartSleep()` entry
- src/components/sleep/controller.cpp:423-454 - `HandleNetFSMFailure()`
- src/components/sleep/hardware.cpp - ESP32 `esp_sleep` APIs, `NVS` operations
- Logic within components 
- Anything that looks out of place 

You can kinda skip:
* Driver changes
* CI/build config files

### Test Suite
@tyeth  a test suite for this PR is located in `tests/sleep_mode/`. Read `tests/sleep_mode/README.md` to get started.

### DIFF
--- a/src/wippersnapper.h
+++ b/src/wippersnapper.h
@@ -30,9 +30,7 @@
 */
 #ifdef WS_DEBUG
 #define WS_DEBUG_PRINT(...)                                                    \
-  {                                                                            \
-    WS_PRINTER.print(__VA_ARGS__);                                             \
-  } /**< Print debug message to serial */
+  { WS_PRINTER.print(__VA_ARGS__); } /**< Print debug message to serial */
 #define WS_DEBUG_PRINTLN(...)                                                  \
   {                                                                            \
     WS_PRINTER.println(__VA_ARGS__);                                           \
@@ -44,11 +42,9 @@
   } /**< Print debug message in hexadecimal */
 #else
 #define WS_DEBUG_PRINT(...)                                                    \
-  {                                                                            \
-  } /**< Debug print */
+  {} /**< Debug print */
 #define WS_DEBUG_PRINTLN(...)                                                  \
-  {                                                                            \
-  } /**< Debug println */
+  {} /**< Debug println */
 #endif
 
 /*!


tests/sleep_mode/README.md
```diff
@@ -0,0 +1,255 @@
...
+    {
+      "timerConfig": {
+        "duration": 12
+      }
+    }
+  ],
+  "components": [...]
+}
+```
+
```
tyeth: This doesn't work for me on the Metro S3, invalid sleep mode is reported.

It feels like there should be matching config files (like full files) for the testing process to be recreatable, even if that then limits some tests to specific hardware setups.

src/components/sleep/controller.cpp
```diff
@@ -0,0 +1,491 @@
...
+
+  // Route based on payload type
+  bool res = false;
+  switch (b2d.which_payload) {
+  case ws_sleep_B2D_enter_tag:
+    res = Handle_Sleep_Enter(&b2d.payload.enter);
+    break;
+  default:
+    WS_DEBUG_PRINTLN("[sleep] WARNING: Unsupported Sleep payload");
+    res = false;
```
tyeth: looks unnecessary to set ret=false here as initialised that way just above

src/components/sleep/controller.cpp
```diff
@@ -0,0 +1,491 @@
...
+    WS_DEBUG_PRINT(" with level ");
+    WS_DEBUG_PRINT(msg->config.ext0.level ? "HIGH" : "LOW");
+    WS_DEBUG_PRINTLN("");
+    break;
+  default:
+    WS_DEBUG_PRINTLN("[sleep] WARNING: Unknown sleep config type");
+    break;
+  }
+
+#ifdef ARDUINO_ARCH_RP2350
```
tyeth: #if USE_STATUS_LED would tie in better with ws_Boards. 
I've also come across some without LED_BUILTIN defined, but we don't yet support any. The suggested change should cover that too.

src/components/sleep/controller.cpp
```diff
@@ -0,0 +1,491 @@
...
+  WS_DEBUG_PRINTLN("[sleep] Handle_Sleep_Enter()");
+
+  // Parse and handle lock state
+  _lock = msg->lock;
+  WS_DEBUG_PRINT("[sleep] Sleep lock state: ");
+  WS_DEBUG_PRINTLN(_lock ? "LOCKED" : "UNLOCKED");
+
+  // If config pin is HIGH during startup, override any existing lock
+  _wake_enable_pin_state = CheckWakeEnablePin();
+  if (_wake_enable_pin_state && _lock) {
```
tyeth: Think this needs to switch based on the pin state not matching the pull direction, so we take `_wake_enable_pin_pull` into account (mentioned inside CheckWakeEnablePin).
Alternatively the question is, what should happen if the rp2350 board has an input_pullup versus an input_pulldown button for the wake enable, and presumably an "input" only pull situation we take as-is i.e. existing logic high check, or maybe we need an inverted logic level flag etc.

src/components/sleep/controller.cpp
brentru: Removed redundant assignment.

brentru:
@tyeth  I've gone through your full review, tagging one more time to see if I missed anything

brentru:
@tyeth A TWDT Regression for ESP32-S3 was found in SleepyDog Library, related PR: https://github.com/adafruit/Adafruit_SleepyDog/pull/63

This is unrelated to the sleep PR as it is an external component, non-blocking as that will be merged in

PR #875, opened 2026-03-10
Displays v2

tyeth:
# Displays v2

Pairs with components PR: [adafruit/Wippersnapper_Components#288](https://github.com/adafruit/Wippersnapper_Components/pull/288).

## Proto changes that land here (focus: `display.proto`, `i2c_output.proto`)

`display.proto`:
- B2D/D2B: display `name` lifted to envelope level as a `pb_callback_t` string (was a field on each sub-message).
- D2B responses: replaced per-message `name` / `did_add` / `did_remove` with envelope `name` + a single `did_succeed` bool.
- `Write` collapsed to just a `message` string (content union removed).
- Added `ws_display_Add_ttl_rgb666_tag` payload for RGB666 dotclock panels.
- SPI pin config: `ws_display_SpiPinConfig` replaced by shared `ws_spi_Config` (+ `pin_dc` / `pin_rst` on the parent add config).
- `Init` → `Write` rename on some flows; minor sleep flow tweaks.

`i2c_output.proto`:
- File removed entirely. Output devices now flow through `i2c.proto` (`DeviceDescriptor` + `DisplayProperties`) and the display driver path.

Other protos (already seen): `digitalio` / `pwm` gained `is_inverted`; everything resynced to Wippersnapper_Protobuf @ `1dff4aa`.

## FW side (FYI, no web work needed)

EPD driver ↔ component mapping (components defined in [Wippersnapper_Components#288](https://github.com/adafruit/Wippersnapper_Components/pull/288)):

| Component (PID) | Driver string | FW class | Chip |
|---|---|---|---|
| `eink-29-flexible-monochrome-296x128` ([4262](https://www.adafruit.com/product/4262)) | `UC8151` | `dispDrvThinkInkMonoM06` | UC8151D |
| `eink-magtag` ([4800](https://www.adafruit.com/product/4800)) — older | `ILI0373` (auto) | `dispDrvThinkInkGrayscale4T5` | ILI0373 |
| `eink-magtag` ([4800](https://www.adafruit.com/product/4800)) — 2025 | `SSD1680` (auto) | `drvDispThinkInkGrayscale4Eaamfgn` | SSD1680 |
| `eink-37-monochrome-416x240` ([6395](https://www.adafruit.com/product/6395)) | `UC8253` | `dispDrvThinkInkMonoBAAMFGN` | UC8253 |
| `eink-42-grayscale-300x400` ([6381](https://www.adafruit.com/product/6381)) | `SSD1683` | `dispDrvThinkInkGrayscale4MFGN` | SSD1683 |
| `eink-583-monochrome-648x480` ([6397](https://www.adafruit.com/product/6397)) | `UC8179` | `dispDrvThinkInkMonoAAAMFGN` | UC8179 |

MagTag uses one shared `eink-magtag` component; firmware auto-detects ILI0373 vs SSD1680 at runtime via SPI probing (`hardware.cpp` ~L258).

Other new / migrated drivers:
- [TFT SPI] ST7789
- [RGB666 dotclock] Qualia S3 (2.1\" round 480×480, 3.2\" bar 320×820)
- [I2C OLED] SH1107 ([5297](https://www.adafruit.com/product/5297)), SSD1306
- [I2C LED segment] 7-seg, QuadAlphaNum
- [I2C char LCD] CharLCD

Other:
- New board `ARDUINO_ADAFRUIT_QUALIA_S3_RGB666`.
- I2C output drivers relocated under `components/display/drivers/`.
- Misc: WDT fed in PingV2, GFX lib for Qualia, no-OTA partitions for TFT feathers, ESP8266 flags.

83 files, +5905 / −500.


src/components/display/hardware.cpp
```diff
@@ -0,0 +1,508 @@
...
+}
+
+bool DisplayHardware::write(ws_display_Write *msg) {
+  if (!_drvDisp) {
+    WS_DEBUG_PRINTLN("[display] ERROR: No display driver initialized!");
+    return false;
+  }
+
+  switch (msg->which_content) {
+  case ws_display_Write_message_tag:
```
brentru: Going to err on the 'were not using this right now' and remove it for now, call write directly

brentru (commented):
Overall great, some feedback and questions attached.

Please doxygen and clang, too. There's a lot of files that are undocumented.

As mentioned on the PB PR, I'd like to remove mentions of I2C Display/Outputs from i2c.proto, so i2c's model should reflect that too

src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -27,6 +27,7 @@
     defined(ARDUINO_ADAFRUIT_FEATHER_ESP32S3_REVTFT) ||                        \
     defined(ARDUINO_ADAFRUIT_FEATHER_ESP32S2_REVTFT) ||                        \
     defined(ARDUINO_ADAFRUIT_QTPY_ESP32S3_N4R2) ||                             \
+    defined(ARDUINO_ADAFRUIT_QUALIA_S3_RGB666) || \
```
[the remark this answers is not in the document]
tyeth: Nope

src/components/display/hardware.cpp
```diff
@@ -0,0 +1,508 @@
...
+// ---------------------------------------------------------------------------
+// SPI EPD initialization
+// ---------------------------------------------------------------------------
+bool DisplayHardware::beginSpiEpd(ws_display_Add *msg) {
+  ws_display_EpdSpiConfig *spi = &msg->interface_type.spi_epd;
+
+  int16_t dc = parsePin(spi->pin_dc);
+  int16_t rst = parsePin(spi->pin_rst);
+  int16_t cs = parsePin(spi->pin_cs);
+  int16_t sram_cs = -1, busy = -1;
```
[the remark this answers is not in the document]
tyeth: I'm not sure if it's advantageous or not, visually it's longer with the extra 4 else lines (2 per variable) and we still set the variables at initialisation (just to zero). Maybe more readable though.

src/components/display/hardware.cpp
```diff
@@ -0,0 +1,508 @@
...
+
+  int16_t width = config->width;
+  int16_t height = config->height;
+
+  // Some controller payloads may omit RGB666 dimensions.
+  // Apply panel-based safe defaults to avoid 0x0 display init.
+  if (width <= 0 || height <= 0) {
+    if (strcmp(msg->panel, "TL021WVC02") == 0 ||
+        strcmp(msg->panel, "adafruit-5792") == 0) {
+      width = 480;
```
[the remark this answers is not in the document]
tyeth: This is set in the component config / proto, and is expected to be found at [hardware.cpp:L347](https://github.com/tyeth/Adafruit_Wippersnapper_Arduino/blob/a8240c4dfdf2059f1616a4de0d599716ab042c11/src/components/display/hardware.cpp#L347) so this section is an unneccessary fallback for testing, will remove as can't see when we'd fall through to it (except maybe when people are adding new displays).

src/components/display/controller.h
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * BSD license, all text here must be included in any redistribution.
+ *
+ */
+#ifndef WS_DISPLAY_CONTROLLER_H
+#define WS_DISPLAY_CONTROLLER_H
+#include "hardware.h"
+#include "wippersnapper.h"
+
+#define MAX_DISPLAYS 4 ///< Maximum number of displays
```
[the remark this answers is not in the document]
tyeth: I can't honestly remember, I think there is a driver limit for some display types (like the M4 Mask had to alter the driver to support two), but mainly I was envisaging the user wanting an externally mounted display (e.g. remote displays/indicators over i2c) in addition to an onboard display, also in the future where we can do individual pixel control then we could support a multi-panel matrix display (A single panel composed of 4x LED Matrixes on the MatrixPortalS3)

brentru:
@tyeth This PR is blocked on https://github.com/adafruit/Wippersnapper_Protobuf/pull/174 and there's a large # of conflicting files.

This week - let's work to get this branch ready for review and merged into migrate-api-v2. 

Then, integrate the latest `display.proto` and `spi.proto` into this branch. When you're done, request a review and I'll take another look at the codebase.

tyeth:
This was already integrated with the new SPI proto, but is now inline with the refactor this week (and Lorens proto suggestions). It builds, the boards come online etc, and a watchdog timer issue was resolved (in pingV2).

The component names and paths are still not ideal, which I will resolve next week.

tyeth:
I've updated the description as you mentioned before Brent, so @lorennorman can have visibility too

brentru:
@tyeth ☝️  Take a look at Copilots output. Feel free to ignore/resolve anything, it's a bot. I will do a full review pass over this on Monday when I have a clear 🧠.

brentru:
@tyeth This PR is +5,905/-500. This is a **lot**. Could you split out the display changes only, alternatively - I will only review the display changes.

src/components/display/controller.h
```diff
@@ -0,0 +1,65 @@
...
+  ~DisplayController();
+  bool Router(pb_istream_t *stream);
+  /*!
+      @brief  Handles a request to add a display.
+      @param  msg   The Display Add message.
+      @param  name  The unique name for the new display.
+      @return True if the display was added successfully, False otherwise.
+  */
+  bool Handle_Display_Add(ws_display_Add *msg, const char *name);
+  bool removeExistingDisplayByName(const char *name);
```
brentru: Make `private`

src/components/display/controller.h
```diff
@@ -0,0 +1,65 @@
...
+  /*!
+      @brief  Handles a request to remove a display.
+      @param  name  The unique name of the display to remove.
+      @return True if the display was removed, False otherwise.
+  */
+  bool Handle_Display_Remove(const char *name);
```
brentru: This should take in a pointer to a `ws_display_` object, not a `const char`. Look at the other `Handle_Display_` functions as a reference.

src/components/display/controller.h
```diff
@@ -0,0 +1,65 @@
...
+      @return True if the write succeeded, False otherwise.
+  */
+  bool Handle_Display_Write(ws_display_Write *msg, const char *name);
+  void update(int32_t rssi, bool is_connected);
+
+private:
+  DisplayHardware *_displays[MAX_DISPLAYS] = {nullptr};
+  uint8_t _num_displays;
+  unsigned long _last_bar_update; ///< Timestamp of last status bar update
+  int8_t findDisplayByName(const char *name);
```
brentru: This is finding an _index_, pls reflect that in the func. name.

src/components/display/controller.cpp
```diff
@@ -0,0 +1,349 @@
...
+    WS_DEBUG_PRINTLN("[display] ERROR: Unable to decode Display B2D envelope");
+    return false;
+  }
+
+  switch (b2d.which_payload) {
+  case ws_display_B2D_add_tag:
+    return Handle_Display_Add(&b2d.payload.add, display_name);
+  case ws_display_B2D_remove_tag:
+    return Handle_Display_Remove(display_name);
+  case ws_display_B2D_write_tag:
```
brentru: As I said in controller, `Handle_Display_Remove` needs to take a `&b2d.payload.` reference like the other cases in this `switch` do.

src/components/display/controller.cpp
```diff
@@ -0,0 +1,349 @@
...
+  if (msg->which_interface_type != ws_display_Add_spi_epd_tag)
+    return;
+  if (msg->which_config != ws_display_Add_config_epd_tag)
+    return;
+
+  ws_display_EPDConfig *config = &msg->config.config_epd;
+
+  // MagTag auto-detection is handled at hardware level (needs SPI probing),
+  // but we can still set mode default
+  if (strncmp(name, "eink-magtag", 11) == 0) {
```
brentru: Use `strcmp` for exact-match.

brentru: `11` is a magic number, if you're comparing directly do something like `sizeof("eink-magtag") - 1`

src/components/display/controller.cpp
```diff
@@ -0,0 +1,349 @@
...
+    @param  msg  The Display Add message.
+    @return True if successful, False otherwise.
+*/
+/*!
+    @brief  Resolves component-name based driver/mode defaults for EPD displays.
+            Called before passing the Add message to hardware so that the
+            controller owns the "what driver" decision and hardware just inits.
+    @param  msg  The Display Add message (may be modified in place).
+*/
+static void resolveEpdDefaults(ws_display_Add *msg, const char *name) {
```
brentru: This function needs to return something if the `name` isn't found,

src/components/display/controller.cpp
```diff
@@ -0,0 +1,349 @@
...
+    }
+  }
+}
+
+/*!
+    @brief  Resolves component-name based driver/panel defaults for RGB666
+            dotclock displays (Qualia).
+    @param  msg  The Display Add message (may be modified in place).
+*/
+static void resolveRgb666Defaults(ws_display_Add *msg, const char *name) {
```
brentru: Same as above func, you should `return` a `bool` on failure modes.

src/components/display/controller.cpp
```diff
@@ -0,0 +1,349 @@
...
+      WS_DEBUG_PRINTVAR(msg->driver);
+      WS_DEBUG_PRINT("', panel '");
+      WS_DEBUG_PRINTVAR(msg->panel);
+      WS_DEBUG_PRINTLN("'");
+      return;
+    }
+  }
+}
+
+bool DisplayController::Handle_Display_Add(ws_display_Add *msg,
```
brentru: Please run doxygen across this PR.

src/components/display/controller.cpp
```diff
@@ -0,0 +1,349 @@
...
+
+  // Resolve component-name defaults before passing to hardware
+  if (msg->type == ws_display_DisplayClass_DISPLAY_CLASS_EPD) {
+    resolveEpdDefaults(msg, name);
+  } else if (msg->type == ws_display_DisplayClass_DISPLAY_CLASS_TFT &&
+             msg->which_interface_type == ws_display_Add_ttl_rgb666_tag) {
+    resolveRgb666Defaults(msg, name);
+  }
+
+  removeExistingDisplayByName(name);
```
brentru: Add a comment saying why you're doing this

src/components/display/controller.cpp
```diff
@@ -0,0 +1,349 @@
...
+  if (!hw->begin(msg, name)) {
+    WS_DEBUG_PRINTLN("[display] ERROR: Failed to initialize display hardware!");
+    delete hw;
+    // Publish failure response
+    ws_display_D2B d2b = ws_display_D2B_init_zero;
+    d2b.name.funcs.encode = encode_string_cb;
+    d2b.name.arg = (void *)name;
+    d2b.did_succeed = false;
+    d2b.which_payload = ws_display_D2B_added_or_replaced_tag;
+    Ws.PublishD2b(ws_signal_DeviceToBroker_display_tag, &d2b);
```
brentru: Abstract building the responses out into a `model` class, and pass them the arguments. Then call `Publish`

src/components/display/controller.cpp
```diff
@@ -0,0 +1,349 @@
...
+    WS_DEBUG_PRINTLN("[display] Processing initial write...");
+    Handle_Display_Write(&msg->write, name);
+  }
+
+  // Publish success response
+  ws_display_D2B d2b = ws_display_D2B_init_zero;
+  d2b.name.funcs.encode = encode_string_cb;
+  d2b.name.arg = (void *)name;
+  d2b.did_succeed = true;
+  d2b.which_payload = ws_display_D2B_added_or_replaced_tag;
```
brentru: See comment about `model`

src/components/display/controller.cpp
```diff
@@ -0,0 +1,349 @@
...
+    @return True if a display was removed, False if no existing display had the
+   same name.
+*/
+bool DisplayController::removeExistingDisplayByName(const char *name) {
+  // If display with same name exists, remove it first
+  int8_t existingIdx = findDisplayByName(name);
+  if (existingIdx >= 0) {
+    WS_DEBUG_PRINTLN("[display] Replacing existing display");
+    delete _displays[existingIdx];
+    _displays[existingIdx] = nullptr;
```
brentru: Redundant, can remove

src/components/display/controller.cpp
```diff
@@ -0,0 +1,349 @@
...
+/*!
+    @brief  Removes an existing display with the same name as the new one being
+   added. This ensures that adding a display with a duplicate name will replace
+            the old one instead of creating a conflict.
+    @param  name  The name of the display to remove.
+    @return True if a display was removed, False if no existing display had the
+   same name.
+*/
+bool DisplayController::removeExistingDisplayByName(const char *name) {
+  // If display with same name exists, remove it first
```
brentru, line 291: Add a `null` check for `name` at the top of the func

src/components/display/controller.cpp
```diff
@@ -0,0 +1,349 @@
...
+  if (!did_remove) {
+    WS_DEBUG_PRINTLN("[display] WARNING: Display not found");
+  }
+
+  // Publish response
+  ws_display_D2B d2b = ws_display_D2B_init_zero;
+  d2b.name.funcs.encode = encode_string_cb;
+  d2b.name.arg = (void *)name;
+  d2b.did_succeed = did_remove;
+  d2b.which_payload = ws_display_D2B_removed_tag;
```
brentru: See comment about using `model` for encoding

src/components/display/controller.cpp
```diff
@@ -0,0 +1,349 @@
...
+}
+
+/*!
+    @brief  Handles a request to write to a display.
+    @param  msg  The Display Write message.
+    @return True if successful, False otherwise.
+*/
+bool DisplayController::Handle_Display_Write(ws_display_Write *msg,
+                                             const char *name) {
+  WS_DEBUG_PRINT("[display] Writing to display: ");
```
brentru, line 340: Add a `(!msg || !name)` guard/check here

src/components/display/controller.cpp
```diff
@@ -0,0 +1,349 @@
...
+  WS_DEBUG_PRINTLNVAR(name);
+
+  int8_t idx = findDisplayByName(name);
+  if (idx < 0) {
+    WS_DEBUG_PRINTLN("[display] ERROR: Display not found!");
```
brentru: Might be useful to print out the display's name here, too.

src/components/display/controller.cpp
```diff
@@ -0,0 +1,349 @@
...
+
+  unsigned long now = millis();
+  if (now - _last_bar_update < 60000)
+    return;
+  _last_bar_update = now;
+
+  for (uint8_t i = 0; i < _num_displays; i++) {
+    if (_displays[i]) {
+      WS_DEBUG_PRINTLN("[display] Updating status bar...");
+      _displays[i]->updateStatusBar(rssi, 100, is_connected);
```
brentru: magic #: `100`

src/components/display/controller.cpp
```diff
@@ -0,0 +1,349 @@
...
+    @brief  Updates the status bar on all displays every 60 seconds.
+    @param  rssi  The current WiFi RSSI.
+    @param  is_connected  Whether MQTT is currently connected.
+*/
+void DisplayController::update(int32_t rssi, bool is_connected) {
+  if (_num_displays == 0)
+    return;
+
+  unsigned long now = millis();
+  if (now - _last_bar_update < 60000)
```
brentru: magic #: `60000`

src/components/display/controller.cpp
```diff
@@ -0,0 +1,349 @@
...
+    }
+  }
+}
+
+/*!
+    @brief  Finds a display by name.
+    @param  name  The display name to search for.
+    @return Index of the display, or -1 if not found.
+*/
+int8_t DisplayController::findDisplayByName(const char *name) {
```
brentru: Guard `null` name first

brentru: and return `-1` if null

src/components/display/controller.cpp
```diff
@@ -0,0 +1,349 @@
...
+  for (uint8_t i = 0; i < _num_displays; i++) {
```
brentru, line 394: You're returning the iterator which is a `uint8_t`, but your function's return signature is `int8_t`.

src/components/display/hardware.h
```diff
@@ -0,0 +1,98 @@
...
+#include "drivers/dispDrvThinkInkGrayscale4MFGN.h"
+#include "drivers/dispDrvThinkInkGrayscale4T5.h"
+#include "drivers/dispDrvThinkInkMonoAAAMFGN.h"
+#include "drivers/dispDrvThinkInkMonoBAAMFGN.h"
+#include "drivers/dispDrvThinkInkMonoM06.h"
+
+/*!
+    @brief  Interface for interacting with display hardware (TFT, eInk,
+            OLED, etc.) in the V2 API.
+*/
```
brentru, line 41: Keep style consistent - don't add doxygen comments in .h files, we comment in the implementation file unless it does not exist (in the case of drivers)

src/components/display/hardware.h
```diff
@@ -0,0 +1,98 @@
...
+      @brief  Updates status bar icons based on current state.
+      @param  rssi            Current WiFi RSSI value.
+      @param  bat             Current battery level (0-100).
+      @param  mqtt_connected  True if MQTT is connected.
+  */
+  void updateStatusBar(int8_t rssi, uint8_t bat, bool mqtt_connected);
+
+private:
+  char _name[64];
+  ws_display_DisplayClass _type;
```
brentru: Shouldn't this be `_class` to match `ws_display_DisplayClass`?

src/components/display/hardware.cpp
```diff
@@ -0,0 +1,490 @@
...
+ * Copyright (c) Tyeth Gundry 2025 for Adafruit Industries.
+ *
+ * BSD license, all text here must be included in any redistribution.
+ *
+ */
+#include "controller.h"
+
+DisplayHardware::DisplayHardware() {
+  memset(_name, 0, sizeof(_name));
+  _type = ws_display_DisplayClass_DISPLAY_CLASS_UNSPECIFIED;
```
brentru: Set the `_type` in the class instead of the ctor.

src/components/display/hardware.cpp
```diff
@@ -0,0 +1,490 @@
...
+const char *DisplayHardware::getName() { return _name; }
+
+int16_t DisplayHardware::parsePin(const char *pinStr) {
+  if (!pinStr || strlen(pinStr) < 2 || pinStr[0] != 'D') {
+    return -1;
+  }
+  return atoi(pinStr + 1);
+}
+
+bool DisplayHardware::begin(ws_display_Add *addMsg, const char *name) {
```
brentru: Add a check for `addMsg` at the top

src/components/display/hardware.cpp
```diff
@@ -0,0 +1,490 @@
...
+  WS_DEBUG_PRINTLNVAR(addMsg->panel);
+
+  switch (addMsg->which_interface_type) {
+  case ws_display_Add_spi_tft_tag:
+    return beginSpiTft(addMsg);
+  case ws_display_Add_spi_epd_tag:
+    return beginSpiEpd(addMsg);
+  case ws_display_Add_ttl_rgb666_tag:
+    return beginTtlRgb666(addMsg);
+  // DSI + i8080 todo
```
brentru: Move this TODO before the `default` case or rm it.

src/components/display/hardware.cpp
```diff
@@ -0,0 +1,490 @@
...
+  dispDrvRgb666 *drv = new dispDrvRgb666(msg->panel);
+  if (!drv)
+    return false;
+
+  int16_t width = config->width;
+  int16_t height = config->height;
+
+  // Some controller payloads may omit RGB666 dimensions.
+  // Apply panel-based safe defaults to avoid 0x0 display init.
+  if (width <= 0 || height <= 0) {
```
brentru, line 500: **BUG**: This fails to initialize but keeps going, you should clean up and `return false` here

src/components/display/hardware.cpp
```diff
@@ -0,0 +1,490 @@
...
+}
+
+bool DisplayHardware::beginSpiTft(ws_display_Add *msg) {
+  ws_display_TftSpiConfig *spi = &msg->interface_type.spi_tft;
+  ws_spi_Config *dev = &spi->spi;
+
+  int16_t cs = parsePin(dev->pin_cs);
+  int16_t dc = parsePin(spi->pin_dc);
+  int16_t mosi = parsePin(dev->pin_mosi);
+  int16_t sck = parsePin(dev->pin_sck);
```
brentru, line 111: These pins are required for SPI, add a check here and bail out early if they're `-1`

src/components/display/hardware.cpp
```diff
@@ -0,0 +1,490 @@
...
+  }
+  ws_display_DisplayProperties *config = &msg->config.config_display;
+
+  if (_drvDisp) {
+    delete _drvDisp;
+    _drvDisp = nullptr;
+  }
+
+  if (strcmp(msg->driver, "ST7789") == 0) {
+    _drvDisp = new dispDrvSt7789(cs, dc, mosi, sck, rst, miso);
```
brentru, line 155: `new` can fail. Before calling `_drvDisp->` functions below, you need to check if it failed, otherwise we're gonna crash.

src/components/display/hardware.cpp
```diff
@@ -0,0 +1,490 @@
...
+  WS_DEBUG_PRINTVAR(rst);
+  WS_DEBUG_PRINT(" MISO:");
+  WS_DEBUG_PRINTLNVAR(miso);
+
+  if (dev->bus != 0) {
+    WS_DEBUG_PRINTLN("[display] ERROR: Non-default SPI bus not supported!");
+    return false;
+  }
+
+  if (msg->which_config != ws_display_Add_config_display_tag) {
```
brentru, line 143: Below we detect a specific EPD display tag.. the print below mentions a display for **SPI TFT**. Is there a SPI TFT tag we're supposed to be checking here?

src/components/display/hardware.cpp
```diff
@@ -0,0 +1,490 @@
...
+  ws_display_EPDConfig *config = &msg->config.config_epd;
+
+  if (_drvDisp) {
+    delete _drvDisp;
+    _drvDisp = nullptr;
+  }
+
+  // MagTag hardware auto-detection (requires SPI probing)
+  const char *driver = msg->driver;
+  if (strncmp(_name, "eink-magtag", 11) == 0 && strlen(driver) == 0) {
```
brentru: Like I mentioned in the `controller`, replace `11` with the sizeof() of what you're comparing against

src/components/display/hardware.cpp
```diff
@@ -0,0 +1,490 @@
...
+  } else if (strcmp(driver, "UC8179") == 0) {
+    _drvDisp = new dispDrvThinkInkMonoAAAMFGN(dc, rst, cs, sram_cs, busy);
+  } else if (strcmp(driver, "UC8253") == 0) {
+    _drvDisp = new dispDrvThinkInkMonoBAAMFGN(dc, rst, cs, sram_cs, busy);
+  } else if (strcmp(driver, "UC8151") == 0) {
+    _drvDisp = new dispDrvThinkInkMonoM06(dc, rst, cs, sram_cs, busy);
+  } else {
+    WS_DEBUG_PRINT("[display] ERROR: Unsupported EPD driver: ");
+    WS_DEBUG_PRINTLNVAR(driver);
+    return false;
```
brentru, line 437: There are a lot of `new` calls here! I said above that `new` can fail. Before calling `_drvDisp->` functions below, you need to check if it failed, otherwise we're gonna crash.

src/components/display/hardware.cpp
```diff
@@ -0,0 +1,490 @@
...
+  }
+
+  // Parse EPD mode
+  if (config->mode == ws_display_EPDMode_EPD_MODE_UNSPECIFIED) {
+    WS_DEBUG_PRINTLN("[display] ERROR: EPD mode is unspecified!");
+    delete _drvDisp;
+    _drvDisp = nullptr;
+    return false;
+  }
```
brentru: If we're gonna fail out and this is only parsing and validation not requiring the init.... move this check up. Place it above the initialization code to where the other validation code is.

src/components/display/hardware.cpp
```diff
@@ -0,0 +1,490 @@
...
+  if (config->text_size > 0)
+    drv->setTextSize(config->text_size);
+
+  if (!drv->begin()) {
+    WS_DEBUG_PRINTLN("[display] ERROR: Failed to begin RGB666 driver!");
+    delete drv;
+    return false;
+  }
+
+  _drvDisp = drv;
```
brentru: Both `beginX()` calls above utilize `_drvDisp` directly. here, you assign it. Why did you change the pattern? Is there a reason?

src/components/display/hardware.cpp
```diff
@@ -0,0 +1,490 @@
...
+#else
+  WS_DEBUG_PRINTLN("[display] ERROR: TTL RGB666 not supported on this board!");
+  return false;
+#endif
+}
+
+// ---------------------------------------------------------------------------
+// I2C display initialization (OLED, CharLCD, LED backpack, etc.)
+// ---------------------------------------------------------------------------
+bool DisplayHardware::beginI2cDisplay(ws_display_Add *msg) {
```
brentru, line 538: validate `msg` first

src/components/display/hardware.cpp
```diff
@@ -0,0 +1,490 @@
...
+  WS_DEBUG_PRINTLN("[display] Calling I2C driver begin()...");
+  if (!drv->begin()) {
+    WS_DEBUG_PRINTLN("[display] ERROR: Failed to begin I2C display driver!");
+    delete drv;
+    return false;
+  }
+
+  _drvDisp = new dispDrvBaseI2c(drv);
+  WS_DEBUG_PRINTLN("[display] I2C display initialized successfully!");
+  return true;
```
brentru: **BUG ** Check result of `new` before returning! If it fails, we're going to be in trouble because it'll fail silently (always `return true`)

src/components/display/hardware.cpp
```diff
@@ -0,0 +1,490 @@
...
+    drv->ConfigureI2CBackpack(cfg->brightness, cfg->alignment);
+  } else {
+    // No matching config — for OLEDs, apply safe defaults to prevent
+    // crash from uninitialized width/height in drvOutSsd1306::begin()
+    WS_DEBUG_PRINTLN("[display] WARNING: No config for I2C display, "
+                     "applying defaults");
+    if (strcasecmp(driverName, "SSD1306") == 0) {
+      drv->ConfigureSSD1306(128, 32, 1);
+    } else if (strcasecmp(driverName, "SH1107") == 0) {
+      drv->ConfigureSSD1306(128, 64, 1);
```
brentru: Why are these defaults being applied? If they are **ALWAYS** applied for the SSD - that should be within the driver.

src/components/display/hardware.cpp
```diff
@@ -0,0 +1,490 @@
...
+    }
+  }
+
+  WS_DEBUG_PRINTLN("[display] Calling I2C driver begin()...");
```
brentru: debug remove

src/components/display/drivers/drvOutSh1107.h
```diff
@@ -0,0 +1,131 @@
...
+ * Copyright (c) Tyeth Gundry for Adafruit Industries 2025
+ *
+ * MIT license, all text here must be included in any redistribution.
+ *
+ */
+
+#ifndef DRV_OUT_SH1107_H
+#define DRV_OUT_SH1107_H
+
+#include "drvOutputBase.h"
```
brentru: Why is this and other drivers using the Output base header, `out` prefix, and the rest use the `dispDrvBase.h`?

brentru (changes requested):
@tyeth  This was a massive amount of work, thank you for taking it on.

Please review my review, along with Copilot's, before resubmitting. Mine is a mix of questions and comments.

src/components/display/controller.cpp
```diff
@@ -0,0 +1,312 @@
...
+      return;
+    }
+  }
+}
+
+bool DisplayController::Handle_Display_Add(ws_display_Add *msg) {
+  WS_DEBUG_PRINT("[display] Adding display: ");
+  WS_DEBUG_PRINTLNVAR(msg->name);
+
+  // Resolve component-name defaults before passing to hardware
```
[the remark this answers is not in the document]
tyeth: I don't see the value / logic for that, but this is an older comment.
Only EPD or RGB666 need the added defaults (rest rely on proto), what are you thinking here that we check config instead of type?

src/components/display/hardware.h
tyeth: one of the linting tools kicked off that the class wasn't documented. I'll remove and try to identify which tool / version it came from.

src/components/display/hardware.cpp
tyeth: I've decided to bail after the printing of the processed pins so debugging is easier.

src/components/display/hardware.cpp
tyeth: this tag is confusingly auto-named, but its the generic ws_display_DisplayProperties (generic display config, not epd/charlcd/seg config)

src/components/display/controller.h
```diff
@@ -0,0 +1,63 @@
...
+class DisplayController {
+public:
+  DisplayController();
+  ~DisplayController();
+  bool Router(pb_istream_t *stream);
+  /*!
+      @brief  Handles a request to add a display.
+      @param  msg   The Display Add message.
+      @return True if the display was added successfully, False otherwise.
+  */
```
brentru: @tyeth - This comment style (in .h) doesn't line up with `display/hardware.cpp`. Please switch to doxygen'ing class methods in the implementation file.

build.ps1
```diff
@@ -0,0 +1 @@
... | Select-Object -First 1 -ExpandProperty Name; $m=Select-String -Path 'platformio.ini' -Pattern '^\s*default_envs\s*=\s*(.+)$' | Select-Object -First 1; $defaultEnv=if($m){ ($m.Matches[0].Groups[1].Value -split ',')[0].Trim() } else { $null }; $envName=if($argEnv){$argEnv}elseif($lastBuilt){$lastBuilt}elseif($defaultEnv){$defaultEnv}else{'adafruit_qualia_s3_rgb666'}; Write-Host "Using env: $envName"; $kErase=$null; $kClean=$null; $sw=[Diagnostics.Stopwatch]::StartNew(); Write-Host 'Press E within 2s to erase flash (default: no)...'; while($sw.ElapsedMilliseconds -lt 2000 -and -not $kErase){ if([Console]::KeyAvailable){ $kErase=[Console]::ReadKey($true).KeyChar } else { Start-Sleep -Milliseconds 50 } }; Write-Host 'Press F within 1.5s for fullclean, C for clean (default: none)...'; $sw.Restart(); while($sw.ElapsedMilliseconds -lt 1500 -and -not $kClean){ if([Console]::KeyAvailable){ $kClean=[Console]::ReadKey($true).KeyChar } else { Start-Sleep -Milliseconds 50 } }; if($kClean -match '^[fF]$'){ pio run -t fullclean -e $envName } elseif($kClean -match '^[cC]$'){ pio run -t clean -e $envName }; if($kErase -match '^[eE]$'){ pio run -t erase -e $envName }; pio run -t upload -e $envName
```
brentru: @tyeth  - What is this file?  May want to remove it

src/components/display/controller.cpp
```diff
@@ -0,0 +1,378 @@
...
+    case ws_display_Add_i2c_tag:
+      Ws.error_handler->publishComponentError(
+          msg->interface_type.i2c,
+          "Failed to initialize I2C display hardware for add request");
+      break;
+    default:
+      Ws.error_handler->publishComponentError(
+          "Unknown interface",
+          "Failed to initialize display hardware for add request");
+      break;
```
brentru: @tyeth You call this giant switch case here and within `Handle_Display_Remove`. Refactor this out into a function instead.

src/components/display/controller.cpp
```diff
@@ -0,0 +1,378 @@
...
+    }
+
+    return false;
+  }
+
+  // Show splash screen and status bar
+  hw->showSplash();
+  hw->drawStatusBar(Ws._configV2.aio_user);
```
brentru: Because these are called by default and in succession, can we instead add a new function within drvBase, like `initialize(char *aio_user)`, to perform this functinality?

src/components/display/hardware.h
```diff
@@ -0,0 +1,73 @@
...
+  ~DisplayHardware();
+  bool begin(ws_display_Add *addMsg);
+  bool write(ws_display_Write *msg);
+  const char *getName();
+  void showSplash();
+  void drawStatusBar(const char *io_username);
+  void updateStatusBar(int8_t rssi, uint8_t bat, bool mqtt_connected);
+
+private:
+  char _name[64];
```
brentru: Shouldn't this match the `display.pb.h` enum field - `char driver[32]; /* * Driver name (e.g., "UC8179", "SSD1306", "HT16K33") */`- and be 32 bytes instead?

src/components/display/hardware.cpp
```diff
@@ -0,0 +1,644 @@
...
+    @param  msg  The Display Add message with SPI EPD interface and config.
+    @return True on success, False on failure.
+*/
+bool DisplayHardware::beginSpiEpd(ws_display_Add *msg) {
+  ws_display_EpdSpiDescriptor *spi_epd_config = &msg->interface_type.spi_epd;
+  ws_spi_Descriptor *spi_pin_config = &spi_epd_config->spi;
+
+  int16_t dc = parsePin(spi_epd_config->pin_dc);
+  int16_t rst = parsePin(spi_epd_config->pin_rst);
+  int16_t cs = parsePin(spi_pin_config->pin_cs);
```
brentru: You should validate `dc`, `rst`, and `cs` the same way you validate the other pins below (check for != -1)

src/components/display/hardware.cpp
```diff
@@ -0,0 +1,644 @@
...
+*/
+uint8_t
+DisplayHardware::EpdBitBangReadRegister(uint8_t cmd,
+                                        ws_display_EpdSpiDescriptor *config) {
+  int16_t mosi = DisplayHardware::parsePin(config->spi.pin_mosi);
+  int16_t miso = DisplayHardware::parsePin(config->spi.pin_miso);
+  int16_t sck = DisplayHardware::parsePin(config->spi.pin_sck);
+  int16_t cs = DisplayHardware::parsePin(config->spi.pin_cs);
+  int16_t dc = DisplayHardware::parsePin(config->pin_dc);
+  int16_t rst = DisplayHardware::parsePin(config->pin_rst);
```
brentru: 1) you can just call `parsePin()` directly, do not need `DisplayHardware::parsePin` - this is the same caller from the same class.
2) `parsePin()` returns a `-1` if False, `EpdBitBangReadRegister` should validate the response rather than just writing to the storage variable for the pin.

src/components/display/hardware.cpp
```diff
@@ -0,0 +1,644 @@
...
+*/
+bool DisplayHardware::write(ws_display_Write *msg) {
+  if (!_drvDisp) {
+    WS_DEBUG_PRINTLN("[display] ERROR: No display driver initialized!");
+    return false;
+  }
+
+  WS_DEBUG_PRINT("[display] Writing message: ");
+  WS_DEBUG_PRINTLNVAR(msg->message);
+  _drvDisp->writeMessage(msg->message);
```
brentru, line 686: Are the variables on `writeMessage()` ever called? Or do we always just pass it the `message`? 

If we don't pass them, we should delete them from the `virtual void` ctor.
```
virtual void dispDrvBase::writeMessage(const char *message, bool clear_first = true, int32_t cursor_x = 0L, int32_t cursor_y = 0L)
```

src/components/display/hardware.cpp
```diff
@@ -0,0 +1,644 @@
...
+ *
+ * BSD license, all text here must be included in any redistribution.
+ *
+ */
+#include "controller.h"
+
+/*!
+    @brief  Constructs a DisplayHardware with no active driver.
+*/
+DisplayHardware::DisplayHardware() { memset(_name, 0, sizeof(_name)); }
```
brentru: I think the `memset` call is legacy - but it is also not required, redundant, and can be dropped.

src/components/display/hardware.cpp
```diff
@@ -0,0 +1,644 @@
...
+  if (_drvDisp)
+    _drvDisp->updateStatusBar(rssi, bat, mqtt_connected);
+}
+
+/*!
+    @brief  Writes a Display Write message to the display.
+    @param  msg  The Display Write message to render.
+    @return True if the write succeeded, False otherwise.
+*/
+bool DisplayHardware::write(ws_display_Write *msg) {
```
brentru, line 674: Oh, we want to check and validate `message` before sending it off to the driver...

src/components/display/controller.cpp
```diff
@@ -0,0 +1,378 @@
...
+       ws_display_EPDMode_EPD_MODE_MONO},
+      {"eink-37-monochrome-416x240", "UC8253",
+       ws_display_EPDMode_EPD_MODE_MONO},
+      {"eink-42-grayscale-300x400", "SSD1683",
+       ws_display_EPDMode_EPD_MODE_GRAYSCALE4},
+      {"eink-583-monochrome-648x480", "UC8179",
+       ws_display_EPDMode_EPD_MODE_MONO},
+  };
+
+  for (auto &m : mappings) {
```
brentru: I don't like use of `auto`, can we be explicit here?

src/components/display/controller.cpp
```diff
@@ -0,0 +1,378 @@
...
+  struct {
+    const char *component;
+    const char *driver;
+    const char *panel;
+  } mappings[] = {
+      {"qualia-round-21-480x480", "ST7701S", "TL021WVC02"},
+      {"qualia-bar-32-320x820", "ST7701S", "TL032FWV01"},
+  };
+
+  for (auto &m : mappings) {
```
brentru: I don't like the use of `auto`, can we be explicit here?

brentru (changes requested):
@tyeth  Overall, this is good. I have a few (12) changes I'd like to see implemented. Once implemented, please tag me in the comments and I will verify + Approve so we can get this merged in.

src/components/display/controller.h
```diff
@@ -0,0 +1,48 @@
...
+*/
+class DisplayController {
+public:
+  DisplayController();
+  ~DisplayController();
+  bool Router(pb_istream_t *stream);
+  bool Handle_Display_Add(ws_display_Add *msg);
+  bool Handle_Display_Remove(ws_display_Remove *msg);
+  bool Handle_Display_Write(ws_display_Write *msg);
+  void update(int32_t rssi, bool is_connected);
```
tyeth, line 40: change `update` to be updateStatusBarMetrics or ideally split apart to UpdateRSSI, Connectivity, Battery.

build.ps1
tyeth: It rebuilds the last build target easily from the commandline, optionally allowing a target name arg, and asks to clean and/or erase first.
I'll drop it.

brentru (approved):
Thanks for addressing feedback, looks good to merge in

PR #900, opened 2026-04-02
fix(ping): Avoid Device Connection Flipping, KeepAlive Fixup

### LINKED ISSUE #899: Device Connection Flipping, KeepAlive Fixup
On the IO frontend, WipperSnapper devices flip "disconnected"/"connected" frequently. It's a pain because:
- it can wipe out frontend activity if you're in a form that relies on being connected
- it "junks up" a device's connection log with many spurious connection events

I believe this can be fixed entirely in the device code, per the following 3 snippets:

constant is set to 5 seconds:
```cpp
#define WS_KEEPALIVE_INTERVAL_MS                                               \
  5000 ///< Session keepalive interval time, in milliseconds
```

broker is told "disconnect me if i don't see you after 5 seconds":
```cpp
      WS._mqtt->setKeepAliveInterval(WS_KEEPALIVE_INTERVAL_MS / 1000);
```

device gives itself a half-second to:
* finish up whatever it was doing
* get the ping packet to the broker
```cpp
void Wippersnapper::pingBroker() {
  // ping within keepalive-10% to keep connection open
  if (millis() > (_prv_ping + (WS_KEEPALIVE_INTERVAL_MS -
                               (WS_KEEPALIVE_INTERVAL_MS * 0.10)))) {
```

Analysis: the device is inadvertently giving itself a mere 500ms to finish up whatever it is doing and get a "ping" packet to the broker. So every 5 seconds, there's a chance this lands after the timeout, causing a "disconnect" following closely by a "connect".

We discussed quickly as a team to make sure this wasn't intended, and we agree it is not. It's more that it's not obvious this is how this all works!

Proposed Fix:
Instead of using 1 constant plus math to express everything, break it into 2 obvious constants and remove the math from the callsites.

```cpp
// ...
#define WS_BROKER_KEEPALIVE_MS 10000 // maximum time without a ping before a disconnect
#define WS_DEVICE_PING_MS 5000 // minimum time before a ping
// ...
// tell the broker how long to wait for a ping
WS._mqtt->setKeepAliveInterval(WS_BROKER_KEEPALIVE_MS / 1000);
// ...
void Wippersnapper::pingBroker() {
  // if it's past time to send the next ping
  if (millis() > (_prv_ping + WS_DEVICE_PING_MS)
// ...
```

I hope this will fix the problem and be easier to reason about going forward.

tyeth:
Fixes #899

### DIFF
--- a/src/Wippersnapper.cpp
+++ b/src/Wippersnapper.cpp
@@ -1746,16 +1746,18 @@ void cbThrottleTopic(char *throttleData, uint16_t len) {
   WS_DEBUG_PRINTVAR(throttleDuration);
   WS_DEBUG_PRINTLN("ms and blocking command execution.");
 
+  const uint32_t pingDelayMs = WS_DEVICE_PING_MS;
+
   // If throttle duration is less than the keepalive interval, delay for the
   // full keepalive interval
-  if (throttleDuration < WS_DEVICE_PING_MS) {
-    delay(WS_DEVICE_PING_MS);
+  if (throttleDuration < pingDelayMs) {
+    delay(pingDelayMs);
   } else {
-    // round to nearest millis to prevent delaying for less time than req'd.
-    float throttleLoops = ceil(throttleDuration / WS_DEVICE_PING_MS);
+    // Round up so throttling never ends earlier than requested.
+    uint32_t throttleLoops = (throttleDuration + pingDelayMs - 1) / pingDelayMs;
     // block the run() loop
     while (throttleLoops > 0) {
-      delay(WS_DEVICE_PING_MS);
+      delay(pingDelayMs);
       WS.feedWDT();
       WS._mqtt->ping();
       throttleLoops--;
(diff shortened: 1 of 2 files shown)


src/Wippersnapper.cpp
```diff
@@ -2474,7 +2474,7 @@ void Wippersnapper::runNetFSM() {
       fsmNetwork = FSM_NET_CHECK_NETWORK;
       break;
     case FSM_NET_ESTABLISH_MQTT:
-      WS._mqtt->setKeepAliveInterval(WS_KEEPALIVE_INTERVAL_MS / 1000);
+      WS._mqtt->setKeepAliveInterval(WS_BROKER_KEEPALIVE_MS / 1000);
```
brentru: We should pre-calculate and cache this arithmetic into a value, possibly within the start of `connect()` instead of calculating every time during runtime

brentru (changes requested):

brentru (changes requested):
@tyeth  Some more small changes, I wrote this func. a while ago!

brentru (approved):

lorennorman:
Just a followup thought here: might it be useful to change these in specific circumstances? Like if:
- a user has a particularly flaky network, "i know my wifi drops daily and it takes 2 minutes for the network to reboot"
- a user needs a tighter disconnect period, "5 seconds is make or break, kick my device at 7 seconds without heartbeat!"
...if they could control these values they could self-help wrt disconnect behavior

We could allow these values to be overridden by, for instance:
- user controls in the frontend
- special `secrets.json` vars

PR #908, opened 2026-04-14
Add VCNL4030

tyeth:
Adds Vishay VCNL4030, matching PR: 
https://github.com/adafruit/Wippersnapper_Components/pull/316

Also adds a commit to ensure weak matching of skills by agents returns something (opus once did .agents/skills/* and found no files so assumed no skills):
* fix(skills): Ensure .agents/skills/* returns a match/file


.agents/skills/SKILLS.md
```diff
@@ -0,0 +1 @@
+Check subfolders for skills in this repo.
```
brentru, line 1: @tyeth  Did you mean to add this?

src/components/i2c/drivers/WipperSnapper_I2C_Driver_VCNL4030.h
```diff
@@ -0,0 +1,134 @@
...
+  bool begin() {
+    if (_vcnl4030) {
+      delete _vcnl4030;
+    }
+    _vcnl4030 = new Adafruit_VCNL4030();
+    if (!_vcnl4030->begin(_sensorAddress, _i2c))
+      return false;
+
+    // Enable and configure sensors, compound result catches any failure
+    bool result = _vcnl4030->enableALS(true);
```
brentru, line 65: If the first call fails, can we do each of these API calls, or would it cause a cascading failure?

brentru (changes requested):

.agents/skills/SKILLS.md
tyeth: Yes.

src/components/i2c/drivers/WipperSnapper_I2C_Driver_VCNL4030.h
tyeth: Assuming a basic register write doesn't fail then it should never happen, but in theory if it did I would expect other calls to potentially fail, and we'd waste a couple of millis doing them instead of potentially bailing after each possible register interaction.

Don't think it's worth changing.

Image

tyeth:
@brentru please re-review / read comments

brentru (approved):
@tyeth  LGTM maybe bump release ver up in here, if you are going to cut one

PR #909, opened 2026-04-15
Test picotool uploadmode for arduino-cli to allow picotool reboot from running firmware

tyeth:
I want to be able to reset an RP2 based board via Picotool.
That requires enabling the reset usb custom vendor interface, which in earle core is gated by ENABLE_PICOTOOL_USB that then allows the extra descriptor required for the reset interface. Which effectively follows https://arduino-pico.readthedocs.io/en/latest/install.html#uploading-sketches-with-picotool

<details><summary>Discord conversation / Addition context: </summary>
<p>

Tyeth  [4:50 PM]
Anyone got a handle on what the situation with picotool reboot is? Particularly in windows, with arduino projects running on device (i.e. not BOOTSEL mode), although any context (circuitpython/win-usb) would be helpful.
From what I've gathered it appears there is a usb version thing (bcd) that needs to be version 0x210 for windows to then ask for extra vendor interface info (including the pico reset interface), otherwise running as bcd 0x200 means simpler windows driver support (no picotool reset). (forum/GH links: [#1](https://forums.raspberrypi.com/viewtopic.php?t=396793) [#2](https://forums.raspberrypi.com/viewtopic.php?t=394703) [#3](https://forums.raspberrypi.com/viewtopic.php?t=394730) [#4](https://github.com/raspberrypi/pico-sdk/issues/1802))
In cmake/pico-sdk it appears to be enabled by default, but [for arduino-pico (earle core) there is a guard around the uploadmethod=picotool](https://github.com/search?q=repo%3Aearlephilhower%2Farduino-pico%20ENABLE_PICOTOOL_USB&type=code) which then switches to add the extra usb interface descriptor for vendor reset, and that is not usually set (0x210) for tinyusb projects (like our Adafruit WipperSnapper Arduino firmware), but in theory fully supported (and only not set due to uploadmode choice).
See https://arduino-pico.readthedocs.io/en/latest/install.html#uploading-sketches-with-picotool
 So my plan is to switch our wippersnapper firmware to use the uploadmethod=picotool during arduino-cli compilation and that will mean I can then reset running wipper firmwares on rp2 devices via picotool. I would imagine that circuitpython is already using cmake+pico-sdk so probably gets the functionality for free. I'll verify that later via lsusb -v so no worries for circuitpython.
 Does that sound about right? Anyone gone through this already? (edited) 
 
Dan Halbert  [5:26 PM]
I would imagine that circuitpython is already using cmake+pico-sdk so probably gets the functionality for free. I'll verify that later via lsusb -v so no worries for circuitpython.Not sure abou this. We do use pico-sdk, and use cmake to some extent, but this may not be a concern for us. How would this affect CircuitPython UF2 updates? We don't use picotool right now for anything.

Tyeth  [5:37 PM]
I don't think you'd notice any difference, as the uf2 doesn't really change (only the firmware does - to include the extra usb vendor descriptor+interface for reset), it's mainly about not having to use the bootsel button to get a device back to bootloader (assuming one has no ability to modify the device code running or ask it to reboot nicely via repl etc).
I think picotool may get used for uf2 generation on the rp235x to ensure various flags are set correctly (although not required as can be done other ways), or that was my impression from earles related comment https://forums.raspberrypi.com/viewtopic.php?t=392235#p2340175 (edited) 

Dan Halbert  [6:28 PM]
[@Tyeth](https://adafruit.slack.com/team/U067P18H41L) We can do this programmatically already from CircuitPython:
>>> import microcontroller
>>> microcontroller.on_next_reset(microcontroller.RunMode.UF2)
>>> microcontroller.reset[6:28 PM]or
>>> import microcontroller
>>> microcontroller.on_next_reset(microcontroller.RunMode.BOOTLOADER)
>>> microcontroller.reset()[6:28 PM]reboots to the bootloader, in both cases (edited) 

Tyeth  [6:29 PM]
yeah that's the thing, I'm trying to reset from host from any running firmware, so in cpy it's easy as long as repl responds, but in WS we don't yet implement the picotool reset support
[6:29 PM]I want to avoid soldering boot pins for HIL test devices (along with extra gpio wires hooked) (edited) 

Dan Halbert  [6:30 PM]
got it - thanks - i wasn't sure what your goa lwas

Tyeth  [6:30 PM]
to be fair having the boot pins is more helpful long term, so the button use or safemode activation etc could be part of tests
[6:31 PM]It was fine til I flashed our WS firmware to the pico and then tried to install something else :joy:

</p>
</details>


tyeth:
Didn't see what I was expecting, not once did `picotool reboot` work for a device running the user application (various).
Running picotool reboot from a bootloader ready device works fine, but it should be possible to reboot into bootloader.

In the end the old serial baud rate at 1200bps did successfully reboot a WS-running PicoW into bootloader mode:
`stty -F /dev/ttyACM0 1200`

tyeth:
Closing this as didn't do what I'd hoped / desired. In the end the usb vendor interface wasn't shown with tinyusb, but the arduino serial reset (baud 1200) did work. With micropython (I was loading WS + CircuitPy + MicroPy in testing) couldn't be reset into bootloader without bootsel button, whereas arduino/circuitpython could.
I'll require bootsel button test point connection in an ideal HIL scenario (the i2c 8x solenoid driver used to control the hub has spare channels for this on the B-port pins).

PR #910, opened 2026-04-15
feat(sensor): add MAX44009 ambient light sensor

tyeth:
Matches https://github.com/adafruit/Wippersnapper_Components/pull/318


src/components/i2c/drivers/WipperSnapper_I2C_Driver_MAX44009.h
```diff
@@ -0,0 +1,119 @@
...
+      @returns  True if initialized successfully, False otherwise.
+  */
+  /*******************************************************************************/
+  bool begin() {
+    if (_max44009) {
+      delete _max44009;
+    }
+    _max44009 = new Adafruit_MAX44009();
+    if (!_max44009->begin((uint8_t)_sensorAddress, _i2c))
+      return false;
```
brentru, line 64: Shouldn't we also `delete _max44009` prior to return? Is this handled automatically?

src/components/i2c/drivers/WipperSnapper_I2C_Driver_MAX44009.h
```diff
@@ -0,0 +1,119 @@
...
+      @returns  True if the sensor was read successfully, False otherwise.
+  */
+  /*******************************************************************************/
+  bool _readSensor() {
+    unsigned long now = millis();
+    if (_lastRead != 0 && (now - _lastRead < ONE_SECOND_IN_MILLIS))
+      return true; // recently read, use cached value
+
+    float lux = _max44009->readLux();
+    if (isnan(lux))
```
brentru, line 89: Returns: `A non-zero value (true) if x is a NaN value; and zero (false) otherwise.`

You may want to do an explicit check here for `isnan()`

brentru (changes requested):

src/components/i2c/drivers/WipperSnapper_I2C_Driver_MAX44009.h
tyeth: It will be handled as this function is called subsequently (first bit of this method does the delete if exists).

Although this raised an interesting assessment that the destructors are not being called as expected. See #911

tyeth:
@copilot resolve the merge conflicts in this pull request

src/components/i2c/drivers/WipperSnapper_I2C_Driver_MAX44009.h
tyeth: Sorry can you explain that to me please? I get the impression the boolean check on line 89 is exactly what we want, and returns true if lux is a NaN otherwise false.

brentru (approved):

brentru:
@tyeth  Workflows werent approved, running now

tyeth:
> @tyeth Workflows werent approved, running now

Ah, that's just the copilot library.properties merge conflict resolution commit needing authorization first time. Merged.
Original work passed:
image

PR #913, opened 2026-04-30
Add expanders

tyeth:
Draft PR to add review comments to these provisional changes


src/components/expander/controller.cpp
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  ws_i2c_DeviceDescriptor desc = msg->cfg.device_description;
+  uint8_t addr = (uint8_t)desc.device_address;
+
+  // Check if this expander has already been added
+  if (GetDriver(addr) != nullptr) {
+    WS_DEBUG_PRINTLN("[expander] ERROR: Expander exists at this address!");
+    return false;
+  }
```
tyeth, line 142: Do we want to allow overwriting with a different mux (assuming the user picked wrong one or wanted to swap/change the current one - diff device same address).
I assume we'd remove the existing one first so no that won't happen.

The "this" in comment wrongly suggests it's only the identical driver possible.

src/components/expander/hardware.h
```diff
@@ -0,0 +1,44 @@
...
+            so higher layers don't need to know the chip.
+*/
+class ExpanderHardware {
+public:
+  virtual ~ExpanderHardware();
+
+  virtual bool begin(uint8_t i2c_addr, TwoWire *wire) = 0;
+  virtual void pinMode(uint8_t pin, uint8_t mode) = 0;
+  virtual void digitalWrite(uint8_t pin, uint8_t value) = 0;
+  virtual uint8_t digitalRead(uint8_t pin) = 0;
```
tyeth: Might be worth having overloads that take an ExpanderPin / DigitalIOPin, like for the nesso, but not sure it's necessary:
https://github.com/espressif/arduino-esp32/blob/e4d54aeb410e5993612d11222c4283f9d773a3d2/variants/arduino_nesso_n1/pins_arduino.h#L47-L54

For adafruit gfx we need compatible constructors, but alternatively we could switch to the M5Stack M5unified library just for the Nesso-N1 board (pretending it's a uniquely special display driver - different to a standard ST7789P3 - to use ExpPins)

src/components/digitalIO/controller.h
```diff
@@ -34,6 +35,9 @@ struct DigitalIOPin {
   ulong prv_pin_time;                   ///< The pin's previous time.
   bool
       did_read_send; ///< True if the last read was sent to IO, False otherwise.
+  ExpanderHardware
+      *expander_drv; ///< If this pin is on an expander, pointer to expander's
+                     ///< driver instance, otherwise nullptr
 };
```
tyeth: Feels like this DigitalIOPin struct should live somewhere else, ws_helpers, or hardware/model.
The GH diff highlight only covers one line as it was a broken diff section

tyeth:
@brentru had a look, nice work, couple of minor thoughts

Issue #914, opened 2026-05-06
Huzzah esp8266 crashing with PWM + LCF i2c batt monitor


tyeth:
https://forums.adafruit.com/viewtopic.php?p=1087845#p1087845

tyeth:
This might be two separate issues, or intertwined during user testing.

The first report is just an LCF battery charger attached - see https://forums.adafruit.com/viewtopic.php?p=1087844#p1087844

The second report is a huzzah pogo-ing, i.e. coming online and offline again repeatedly. There are logs provided, which show a few boots, some initialising PWM first then i2c, even two pwm pins at one point (@tyeth check magic config for huzzah 8266 as looks like two leds defined [blue/red]). https://forums.adafruit.com/viewtopic.php?t=223708

brentru:
@tyeth Thanks for raising this issue, do you want to try resolving this bug? Could you timebox this task to 2.5hr? If you can't resolve it - please add a postmortem of what you discovered, how far you got, and then tag me in here and I'll take a look.

tyeth:
Unable to recreate issue. Had PWM LEDs and LC7 lipo fuel gauge. Various recreation attempts. Using version 129. See https://forums.adafruit.com/viewtopic.php?p=1089649#p1089649

Think we let this one lie for now @brentru and hope that my battery flattening test shows something, or the user comes back with more info.

It is worth noting that the fix raised (https://github.com/tyeth-ai-assisted/adafruit-Adafruit_Wippersnapper_Arduino/pull/1) attempts to avoid timer clashes while doing sensitive i2c operations, which seems sensible enough and could probably be released as is.

tyeth:
Also to confirm the RED LED is near the USB port (other side from orange charging LED), and the Blue LED is on the esp8266 SOM (near wifi aerial), with both tested as working simultaneously as dimmable LED components (PWM) at the same time as the LC7xxxF lipo fuel gauge, so the component magic config should not be adjusted.

PR #922, opened 2026-05-22
Fix: HUZZAH8266 Crash, PWM

tyeth:
An attempt to prevent the PWM issues seen in #914

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,7 @@ examples/Wippersnapper_demo/build/
 
 # Platformio artifacts
 .pio/
+output.map
 
 # Secrets
 data/
--- a/src/Wippersnapper.cpp
+++ b/src/Wippersnapper.cpp
@@ -615,7 +615,9 @@ bool cbDecodeI2CDeviceInitRequestList(pb_istream_t *stream,
     return true;
   }
 
+  pauseStatusLEDPWM();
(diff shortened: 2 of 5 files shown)


brentru:
@tyeth Have you tested this on ESP8266 hardware and confirmed the crash no longer occurs?

tyeth:
@brentru No, I didn't find my debug setup so didn't want to waste much time on recreating it at the end of the week. 
Had planned to do a few runs on Prod to recreate then set it against protomq repeatedly. 
I don't think it's reliable as the user managed to get online subsequently.
Instead I chucked the map + elf + stack decode into claude with the additional context and it suggested the likelyhood of pwm and i2c timer clashes in critical sections.

tyeth:
To note that I attempted to recreate the issue, updated notes in https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/issues/914 and the forum, but no joy. I'll test this next week and we'll see if drwho comes back in the forums with any more info.

Reminder to self, update the learn guide page to not recommend the LC7xxxxF lipo fuel guage and instead users should purchase the MAX1704x or another lipo chip. Also stick short bit in component json, description field, so shows in works-with-wippersnapper page.

tyeth:
Also if required to hold pwm for i2c transactions then v2 should take on the fix too

PR #925, opened 2026-05-29
Add - I2c Sensor Settings

brentru:
### Overview
Users can change the polling interval of an I2C component. Some I2C components have extra settings which may influence sensor readings - gain, measurement time, resolution, read rate, etc.

Currently, those settings align with the defaults from the example code (or the datasheet) and users are unable to set up these extra settings.

This pull request, in conjunction with https://github.com/adafruit/Wippersnapper_Protobuf/pull/197 and https://github.com/adafruit/Wippersnapper_Components/pull/327, opens the door for users to configure these settings from Adafruit IO Web.

#### Diagram
```mermaid
flowchart TD
    Router("I2cController::Router()<br><small>Add arrives, buffered for decode</small>")
    HandleAdd("Handle_Add()<br><small>forwards settings to driver</small>")
    Configure("drvBase::configure(settings)<br><small>walks settings, matches key</small>")
    Virtual("setTempOversampling()<br><small>virtual key → method hook</small>")
    Override("BME680::setTempOversampling()<br><small>maps value, writes I2C register</small>")
    Defaults("configureDefaults()")

    Router --> HandleAdd
    HandleAdd --> Configure
    Configure -->|settings present| Virtual
    Configure -->|no settings| Defaults
    Virtual --> Override

    classDef controller fill:#E6F1FB,stroke:#185FA5,color:#0C447C;
    classDef driver fill:#E1F5EE,stroke:#0F6E56,color:#085041;
    class Router,HandleAdd controller;
    class Configure,Virtual,Override,Defaults driver;
```

#### Explation

On IO Web;
* User selects BME680, "Advanced Settings", **Temperature Oversampling** and selects 1x. Clicks Create Component.

* For BME680, the definition has a settings object:
```
"settings": {
    "temp_oversampling": {
      "displayName": "Temperature Oversampling",
      "default": 0,
      "options": ["None", "1x", "2x", "4x", "8x", "16x"]
    },
...
```
* For sensors that support them, the `settings` are packed into the I2C `Add` message under `ws.config.Settings settings` where `settings` is a map of a string and the index (within `options`). 

On the Device; 
* For I2c components, the `Add` message arrives in the `I2cController`'s `Router()` function and is passed to decode in the model (we buffer this to avoid overwriting the stream) and `Handle_Add` is called.
* Within `Handle_Add`, the array of settings is sent to the base driver (`drvBase`)'s `configure()` function.
* `drvBase::configure` walks the `settings` by index (specified by the json's `options`) and performs a check for the key (specified by the keys within the json's `options`, in this case it is `temp_oversampling`)
* drvBase contains `virtual` functions for each key mapping, `temp_oversampling` maps to `setTempOversampling`
* The driver for the BME680 implements an `override`  for setTempOversampling, which performs the driver-specific mapping and function call to configure the i2c device's registers.
* If `settings` are not provided by IO, the `configureDefaults()` function will be called. 
  * This **new function** in the drivers provides a separation between the transport/sensor initialization (`begin()` ) and configuration (`configureDefaults()`) and will likely needed to be implemented across all `drivers/`


tyeth (commented):
Not sure we want to have a settings method for every custom setting key in the drvBase class.

I'd half expected a map of key to private driver setting func inside each driver, or a single method that effectively switch cased all the settings keys. 
In theory the configureDefaults would be called first so any settings would be overloading.

I guess there's no reason for my preference, just stylistic and previous expectations, and it would be more repetition of maps.

src/components/i2c/drivers/drvVeml7700.h
```diff
@@ -55,24 +55,132 @@ class drvVeml7700 : public drvBase {
...
+    // Get sensor event populated in lux via the configured lux method
+    lightEvent->light = _veml->readLux(_luxMethod);
 
     return true;
   }
 
 protected:
   Adafruit_VEML7700 *_veml; ///< Pointer to VEML7700 light sensor object
+  luxMethod _luxMethod =
+      VEML_LUX_NORMAL; ///< Lux reading method used in getEventLight().
```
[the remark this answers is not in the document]
tyeth: Yes, agreed. If we can avoid changing functionality / expectations from v1 to v2 in the same components, then we should do so wherever possible.

src/components/i2c/drivers/drvBmp280.h
```diff
@@ -60,21 +60,191 @@ class drvBmp280 : public drvBase {
...
+      @returns  True if configured successfully, False otherwise.
+  */
+  bool configureDefaults() override {
     _bmp->setSampling(Adafruit_BMP280::MODE_NORMAL,  /* Operating Mode. */
                       Adafruit_BMP280::SAMPLING_X2,  /* Temp. oversampling */
                       Adafruit_BMP280::SAMPLING_X16, /* Pressure oversampling */
                       Adafruit_BMP280::FILTER_X16,   /* Filtering. */
                       Adafruit_BMP280::STANDBY_MS_500); /* Standby time. */
+    return true;
+  }
```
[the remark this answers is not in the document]
tyeth: Good catch

src/components/i2c/drivers/drvBase.h
```diff
@@ -287,6 +288,216 @@ class drvBase {
   */
   virtual void SelectMUXChannel(uint8_t channel) { return; }
 
+  /*!
+      @brief    Applies a set of decoded settings to the driver. When no
+                settings are provided, the driver's default configuration is
+                applied instead. Dispatches each setting key to the matching
+                set* method; unknown keys are reported via the return value.
```
tyeth, line 211: False: unknown keys are not reported via the return value

src/components/i2c/drivers/drvBase.h
```diff
@@ -287,6 +288,216 @@ class drvBase {
   */
   virtual void SelectMUXChannel(uint8_t channel) { return; }
 
+  /*!
+      @brief    Applies a set of decoded settings to the driver. When no
+                settings are provided, the driver's default configuration is
```
tyeth, line 209: Makes more sense to call ConfigureDefaults and then call Configure with custom settings to overload specific ones.

src/components/i2c/drivers/drvBme680.h
```diff
@@ -61,22 +61,137 @@ class drvBme680 : public drvBase {
...
+      is_success = false;
     if (!_bme->setPressureOversampling(BME680_OS_4X))
-      return false;
+      is_success = false;
     if (!_bme->setIIRFilterSize(BME680_FILTER_SIZE_3))
-      return false;
+      is_success = false;
     // 320*C for 150 ms
     if ((!_bme->setGasHeater(320, 150)))
+      is_success = false;
```
tyeth: wonder if we want to join these with `&& (!blah)` as each one is run currently regardless of outcome. Or use || to maintain existing behaviour (which maybe makes sense as we still want to attempt to set each param despite one failure).

src/components/i2c/drivers/drvBmp280.h
```diff
@@ -114,13 +284,35 @@ class drvBmp280 : public drvBase {
...
+  Adafruit_BMP280::sensor_mode _mode =
+      Adafruit_BMP280::MODE_NORMAL; ///< Operating mode
+  Adafruit_BMP280::sensor_sampling _temp_sampling =
+      Adafruit_BMP280::SAMPLING_X16; ///< Temperature oversampling
+  Adafruit_BMP280::sensor_sampling _press_sampling =
+      Adafruit_BMP280::SAMPLING_X16; ///< Pressure oversampling
+  Adafruit_BMP280::sensor_filter _filter =
+      Adafruit_BMP280::FILTER_OFF; ///< Filter setting
+  Adafruit_BMP280::standby_duration _duration =
+      Adafruit_BMP280::STANDBY_MS_1; ///< Standby duration
```
tyeth, line 336: match configure defaults (copilot spotted)

src/components/i2c/drivers/drvApds9999.h
```diff
@@ -0,0 +1,286 @@
...
+      break;
+    case 3:
+      _apds9999->setLightGain(APDS9999_LIGHT_GAIN_9X);
+      break;
+    case 4:
+      _apds9999->setLightGain(APDS9999_LIGHT_GAIN_18X);
+      break;
+    default:
+      return false;
+    }
```
tyeth, line 147: The other drivers in this PR use a default value for all custom settings instead of returning false. I'd quite like it if we returned the failed key + value so it could be traced/noticed. I assume in future io-rails can identify the ws version upon error and offer sensible advice (update firmware as you're using an unsupported new setting)

src/components/i2c/drivers/drvLtr329_Ltr303.h
tyeth, line 242: This wont work anymore for the RAW event, as raw comes after light, and needs the appropriate read delay (integration time plus 10ms) as mentioned in code comment.
Alternatively a single caching read method would resolve.

src/components/i2c/drivers/drvTmp117.h
```diff
@@ -57,6 +57,112 @@ class drvTmp117 : public drvBase {
...
+  */
+  bool setMode(int32_t mode) override {
+    tmp117_mode_t meas_mode;
+    switch (mode) {
+    case 0:
+      meas_mode = TMP117_MODE_CONTINUOUS;
+      break;
+    case 1:
+      meas_mode = TMP117_MODE_SHUTDOWN;
+      break;
```
tyeth, line 167: What should the driver do if the mode == shutdown?

src/components/i2c/model.cpp
```diff
@@ -284,6 +284,93 @@ bool I2cModel::DecodeI2cDeviceAddReplace(pb_istream_t *stream) {
...
+  str_buf[len] = '\0';
+  return true;
+}
+
+bool I2cModel::cbDecodeSettingsEntry(pb_istream_t *stream,
+                                     const pb_field_t *field, void **arg) {
+  I2cModel *model = (I2cModel *)*arg;
+  if (model->_decoded_settings_count >= MAX_I2C_SETTINGS) {
+    return false;
+  }
```
tyeth, line 338: We want to report this back to the broker, as it's a useful signal. One level higher we also print if a settings decode re-add fails, so maybe that wants to be hoisted to broker too, but they feel like different error sources/values so worth distinguishing between them.

src/components/i2c/drivers/drvBase.h
brentru: Yeah but then you are potentially setting the same registers twice vs once?

src/components/i2c/drivers/drvBme680.h
brentru: Could you propose a code snippet of what you're thinking using ``` syntax below?

src/components/i2c/drivers/drvBmp280.h
brentru: I don't understand what you want me to change here, could you elaborate?

tyeth (commented):
might have left comments unsubmitted so submitting again

src/components/i2c/drivers/drvApds9999.h
brentru: Handling this within the drvBase rather than the driver configuration func call itself

src/components/i2c/drivers/drvTmp117.h
brentru: When the TMP117 is set to shutdown mode, all temperature conversions are aborted and the TMP117 will enter low power shutdown mode. The user can change the mode if they want to bring it back up.

src/components/i2c/model.cpp
brentru: Added in the caller, not the decode function

src/components/i2c/drivers/drvLtr329_Ltr303.h
brentru: Doing a single-caching read instead

brentru:
@tyeth  - I've done some significant refactoring, could you pls. re-review?

src/components/expander/controller.cpp
```diff
@@ -181,15 +193,25 @@ bool ExpanderController::Handle_Add(ws_expander_Add *msg) {
...
-                                                "Unsupported setting key!");
-        continue;
+        found = true;
+        if (!(drv->*handler.setter)(DecodedSettingToValue(settings[i]))) {
+          char err[64];
+          snprintf(err, sizeof(err), "Failed to apply setting: %s",
+                   settings[i].key);
+          Ws.error_handler->publishComponentError(desc, err);
+        }
+        break;
```
tyeth, line 208: I'd be tempted to continue / not break out the settings collection loop. 
We can probably safely process further settings. 
Assuming the proto map is not ordered (no idea) then when we add a new setting all older firmwares could fail to apply any settings (or at least some valid key/values that came after an invalid key/value) because of not recognising one of them.

tyeth (changes requested):

src/components/expander/controller.cpp
tyeth: Ignore that, I missunderstood the double loop (one for settings collection and another for the key matching, we are just breaking from key matching.

tyeth (approved):

src/components/i2c/drivers/drvBase.h
tyeth: In my head that's safer / still okay, although I'm sure some odd chips do weird poorly documented things when certain registers are set.

Also we will miss some defaults if only specific settings are received / exposed.

src/components/i2c/drivers/drvBmp280.h
tyeth: e.g. this member is initialised to a different value that we set in configureDefaults, 1millisecond versus 500ms in configureDefaults. The same with all settings, they should ideally match, although I suppose no reason if configureDefaults is always called (still up for debate).
```
Adafruit_BMP280::standby_duration _duration =
      Adafruit_BMP280::STANDBY_MS_1; ///< Standby duration
```

tyeth (commented):
I think this looks great. 

I'm still not convinced that calling configureDefaults only if no settings are supplied is correct. 
Mainly due to not being sure if we will send every setting over proto or a subset of the init parameters for the sensor/component. Also I'd expect the majority of configuration registers to accept multiple sets idempotently, but am aware the odd sensor does a mode change/reset when key registers change (something like alter integration time and sampling mode goes to standby until recalled).

PR #927, opened 2026-06-05
fix(pixels): initialise all members of strands array

tyeth:
Fixes #926

Previously the strands array was only initialising the first member with the supplied initialiser (and the remainder zero initialised rather than using the supplied values):
```
strand_s strands[MAX_PIXEL_STRANDS]{
    nullptr,
    nullptr,
    wippersnapper_pixels_v1_PixelsType_PIXELS_TYPE_UNSPECIFIED,
    0,
    0,
    wippersnapper_pixels_v1_PixelsOrder_PIXELS_ORDER_UNSPECIFIED,
    -1,
    -1,
    -1};
```

This meant the safety check of getStrandIndex(atoi(pin)) did a check on the first element and was fine, but would presumably crash because the pins being checked were not -1 when iterating over further strands array elements, allowing the write to proceed on an unsetup PWM channel.

Change is to do the default intialising of all members in a new constructor.

Issue #926, opened 2026-06-05
Setting value on DotStar causes crash


brentru:
> i clicked a color on the dotstar, value is sent to the board, firmware crashes. basically the firmware got a Write for a component it never got an Add for

We should just log this and drop the packet, rather than crashing

Firmware: WipperSnapper Beta 129
Hardware: Adafruit FunHouse

PR #930, opened 2026-06-14
Create hil-test-suite.yml for HIL tests

tyeth:
Add initial content for HIL test suite configuration. This is a placeholder PR to experiment with in PR feedback and runners etc.


tyeth:
## 🔌 HIL test suite added (pixelWrite / uninitialised-strand regression)

Replaced the `hil-test-suite.yml` stub with a hardware-in-the-loop workflow that exercises the **#926/#927** fix on a **real QT Py ESP32-S3** via our `usbip-hil-controller`.

**What it checks** — a v1 `ws.signal.pixelWrite` to an *uninitialised* NeoPixel strand (pin **D0**, colour **200**) after MQTT check-in:
- **release `1.0.0-beta.127`** → expected to **crash + reboot** (null-deref: pin `0` collides with the zero-init `strands[]` sentinel → `getStrandIdx` false-matches → `neoPixelPtr->fill()` on `nullptr`).
- **this PR / fixed build (#927)** → hits the `ERROR: Pixel strand not found…` guard and **continues, no reset**.

**How it runs**
1. After the build workflow, joins Tailscale and asks the controller `GET /v1/targets` which boards are available — **only `qtpy_esp32s3_n4r2` is online today**; the rest of the bench is reported offline and **skipped + listed** (no red CI for unavailable hardware).
2. Pulls the **release** fatfs zip and **this PR's** build artifact, flashes each, boots against a per-session **protomq** broker (the broker host:port is supplied by the controller — CI never passes it), then injects the v1 pixelWrite and records `PIXELWRITE_VERDICT rebooted=true|false`.
3. Captures **`serial.log` / `protomq.log` / `flash.log`** + the firmware version from the FAT `*boot_out.txt`, attaches them as the `hil-assets` artifact, and posts a per-target pass/fail table here. Uploaded firmware is tracked as a job asset (and purged with the job).

**Contract:** *low = a published release, high = this PR's build.* Any other combination routes to a custom script run.

**Triggering** — `workflow_run` fires only from the **default branch**, so it auto-runs once this PR merges; **before merge, run it via `workflow_dispatch`** (`low_ref`, `high_ref`) once the secrets below are set.

**Required** — repo secrets `TAILSCALE_AUTHKEY_TYETH`, `HIL_API_KEY_TYETH` (+ `HIL_IO_USERNAME`/`HIL_IO_KEY`/`HIL_WIFI_SSID`/`HIL_WIFI_PASSWORD`); optional vars `HIL_API_BASE` / `HIL_LOW_REF` / `HIL_TARGETS`.

> Status: workflow + bench-side capabilities (target availability, firmware upload/URL fetch with asset tracking, the `inject_pixelwrite` stage, job-asset API) are in place and deployed; pending its first live run (set secrets → `workflow_dispatch`, or merge).

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

tyeth (commented):
Looks relatively good. Should surface all the ENV VARs like Lines Before and After the test phrases / conditions in logs provided in the PR summaries. Either as workflow options (with defaults) and/or at top of file. Possibly a readme. Possibly move script elsewhere (lib subfolder?). Tests could move too. The WS Python HIL tests will alter things a bit and maybe bring more clarity on reusable fixtures (like a dual library set of mqtt protobuf scenarios in pytest and then individual lib / serial expectations)

> final step:
> we probably want serial since boot in the quotes by default for these specific tests (if the lines-before proof summary env var is -1), and the timestamps of the two quoted log sections for the last test run show different time periods by a minute or so which shows they don't line up and the protomq log quote is therefore useless (missed event) especially for comparison with the serial log

tyeth:
Note supposedly when testing bossa/bossac on SAMD51, that pyportal titano didn't like v128, but did like v78, so bisect/investigate using HIL suite (it's on rpi-displays host).

Think it was on an alpha debug copy of airlift so probably hated semver or saw unexpected serial msgs at boot etc. v78 + 128 + 130 fine after airlift firmware change to latest.

PR #931, opened 2026-06-16
fix(fs): gate flash reformat on flash health, not brownout reset reason

tyeth:
A battery sag during a flash write or WiFi TX often does NOT surface as RTCWDT_BROWN_OUT_RESET - it commonly reports as POWERON/SW_CPU/watchdog resets instead - so WS.brownOutCausedReset is unreliable as the guard for whether to (re)format the user partition.

Replace the reset-reason-based format decision in the FS constructor with flashSafeToFormat(), which is independent of the reset reason:

- Mount without formatting; on failure, settle (delay) and retry once.
- Only (re)format if the flash is proven healthy:
  - ESP32 on-chip flash: validate the partition-table magic at 0x8000 (0xAA50) and a valid NVS page-header state at 0x9000 via esp_flash_read.
  - External QSPI/SPI (SAMD51) and RP2040/RP2350: confirm a real JEDEC ID, then require the FAT boot sector to read back genuinely erased (all 0xFF) before formatting.
- If reads look unstable (likely voltage/brownout), halt without formatting and without powering USB/LED, to conserve a dying battery.

The existing brownOutCausedReset battery-saving guards (eraseCPFS, createBootFile, parseSecrets) are left intact.

### DIFF
--- a/src/provisioning/tinyusb/Wippersnapper_FS.cpp
+++ b/src/provisioning/tinyusb/Wippersnapper_FS.cpp
@@ -45,6 +45,12 @@ Adafruit_FlashTransport_SPI flashTransport(EXTERNAL_FLASH_USE_CS,
 // ESP32-S2 uses same flash device that stores code.
 // Therefore there is no need to specify the SPI and SS
 Adafruit_FlashTransport_ESP32 flashTransport;
+// The FAT user partition lives on the same on-chip flash as the ESP32
+// bootloader, partition table and NVS - so we can validate those known-good
+// structures to tell a blank/corrupt filesystem apart from unstable
+// (brownout) flash reads. See flashSafeToFormat().
+#define WS_USE_ESP32_INTERNAL_FLASH 1
+#include <esp_flash.h>
 #elif defined(ARDUINO_ARCH_RP2040)
 // RP2040 use same flash device that store code.
 // Therefore there is no need to specify the SPI and SS
@@ -89,6 +95,84 @@ bool setVolumeLabel() {
 
 /**************************************************************************/
 /*!
+    @brief    Decides whether it is safe to (re)create the filesystem after a
+              mount failure, i.e. whether the failure looks like a genuinely
+              blank/corrupt FAT user partition (safe to format) rather than a
+              flash chip whose reads are momentarily unstable due to a voltage
+              sag / brownout (must NOT be formatted - that would erase the
+              user's data on recovery). A brownout is frequently NOT reported
+              as a brownout reset reason, so we cannot rely on the reset
+              reason alone before doing something destructive.
+
+              On the ESP32 the FAT partition shares the on-chip flash with the
+              bootloader, partition table (0x8000) and NVS (0x9000), so those
+              known-good structures are validated directly. Every other
+              FAT-path board (external QSPI/SPI flash on SAMD51, or the on-chip
+              flash on RP2040/RP2350) has no such off-filesystem structure, so
+              the FAT volume's own boot sector is inspected instead - it may
+              only be formatted if it reads back genuinely erased (all 0xFF).
+    @returns  True if a (re)format is safe, False if the flash is unreadable /
+              unstable / not genuinely blank, in which case it must NOT be
+              formatted.
+*/
+/**************************************************************************/
+bool flashSafeToFormat() {
+#ifdef WS_USE_ESP32_INTERNAL_FLASH
+  uint8_t buf[2];
+
+  // Partition table @ 0x8000 - the first entry must begin with the ESP-IDF
+  // partition magic 0x50AA (stored little-endian: 0xAA, 0x50).
+  if (esp_flash_read(esp_flash_default_chip, buf, 0x8000, sizeof(buf)) !=
+      ESP_OK)
+    return false;
+  if (buf[0] != 0xAA || buf[1] != 0x50)
+    return false;
+
+  // NVS partition @ 0x9000 - the first page header state must be a valid NVS
+  // page state: UNINITIALIZED 0xFFFFFFFF / ACTIVE 0xFFFFFFFE / FULL
+  // 0xFFFFFFFC / FREEING 0xFFFFFFF8. The high byte is always 0xFF; the low
+  // byte distinguishes the state. Random garbage from a browning-out SPI bus
+  // is very unlikely to match this.
+  if (esp_flash_read(esp_flash_default_chip, buf, 0x9000, sizeof(buf)) !=
+      ESP_OK)
+    return false;
+  if (buf[1] != 0xFF ||
+      (buf[0] != 0xFF && buf[0] != 0xFE && buf[0] != 0xFC && buf[0] != 0xF8))
+    return false;
+
+  return true;
+#else
+  // External/QSPI flash (SAMD51: PyPortal, Metro M4 AirLift, Titano) or the
+  // RP2040/RP2350 on-chip flash: there is no off-filesystem structure to
+  // validate, so inspect the FAT volume's own boot sector. SPIFlash reads are
+  // relative to the filesystem on every transport, so offset 0 is the boot
+  // sector.
+
+  // First confirm a real flash chip is responding - a dead/disconnected SPI
+  // bus often floats high and reads back as all 0xFF, which would otherwise be
+  // mistaken for a genuinely-erased (blank) chip.
+  uint32_t jedecID = flash.getJEDECID();
+  if (jedecID == 0 || jedecID == 0xFFFFFF)
+    return false;
+
+  // A (re)format is only appropriate when the boot sector is genuinely erased
+  // (all 0xFF). Any other unmountable content is treated as possible
+  // voltage/brownout corruption and must NOT be erased.
+  uint8_t sector[256];
+  for (uint32_t addr = 0; addr < 512; addr += sizeof(sector)) {
+    if (flash.readBuffer(addr, sector, sizeof(sector)) != sizeof(sector))
+      return false; // could not read the volume -> unsafe to format
+    for (uint32_t i = 0; i < sizeof(sector); i++) {
+      if (sector[i] != 0xFF)
+        return false; // non-erased content -> not blank -> refuse to format
+    }
+  }
+  return true; // boot sector is fully erased -> genuinely blank -> safe
+#endif
+}
+
+/**************************************************************************/
+/*!
     @brief    Initializes USB-MSC and the QSPI flash filesystem.
 */
 /**************************************************************************/
@@ -106,20 +190,31 @@ Wippersnapper_FS::Wippersnapper_FS() {
   // Wait for detach
   delay(500);
 
-  // If a filesystem does not already exist - attempt to initialize a new
-  // filesystem
+  // Attempt to mount the existing filesystem WITHOUT formatting.
   if (!initFilesystem()) {
-    if (WS.brownOutCausedReset) {
-      // try once more for good measure
-      delay(10); // let power stablise after failure
-      if (!initFilesystem()) {
-        // no lights, save power as we're probably on a low battery
-        fsHalt("Brownout detected. Couldn't initialise filesystem.");
+    // Mount failed. A battery sag / brownout can momentarily corrupt SPI
+    // flash reads - often WITHOUT being reported as a brownout reset - so
+    // give the power rail a moment to settle and retry before doing anything
+    // destructive.
+    delay(50); // let power stabilise after failure
+    if (!initFilesystem()) {
+      // Still failing. Only (re)format if flashSafeToFormat() proves the flash
+      // chip itself is healthy. This is independent of the reset reason (which
+      // frequently does NOT report a brownout): if the reads are unstable, a
+      // format would needlessly erase the user's data on recovery.
+      if (!flashSafeToFormat()) {
+        // Likely a voltage/brownout event. Keep USB and the status LED off to
+        // conserve a dying battery, and halt.
+        fsHalt("Flash reads unstable (possible brownout) - refusing to "
+               "format. Recharge/repower the board and reset.");
       }
-    } else if (!WS.brownOutCausedReset && !initFilesystem(true)) {
-      TinyUSBDevice.attach();
-      setStatusLEDColor(RED);
-      fsHalt("ERROR Initializing Filesystem");
+      // Flash is healthy: the FAT partition is genuinely blank or corrupt, so
+      // it is safe to create a new filesystem. [DESTRUCTIVE]
+      if (!initFilesystem(true)) {
+        TinyUSBDevice.attach();
+        setStatusLEDColor(RED);
+        fsHalt("ERROR Initializing Filesystem");
+      }
     }
   }
 


src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -87,6 +93,84 @@ bool setVolumeLabel() {
...
+  // byte distinguishes the state. Random garbage from a browning-out SPI bus
+  // is very unlikely to match this.
+  if (esp_flash_read(esp_flash_default_chip, buf, 0x9000, sizeof(buf)) !=
+      ESP_OK)
+    return false;
+  if (buf[1] != 0xFF ||
+      (buf[0] != 0xFF && buf[0] != 0xFE && buf[0] != 0xFC && buf[0] != 0xF8))
+    return false;
+
+  return true;
```
tyeth, line 143: PArtition table can be 0x9000 so do not assume 0x8000 is only valid location. Pretty sure there's an idf command to validation partition table / print location. We should attempt partition validation at both 0x8000 || 0x9000

tyeth: Also, still abort format on brownout reasons, skip flash checks. Also, if failures to get secrets/flash access, then should do a log or two with flush and 1s wait and flush and 0.1s wait then deepsleep for 30m, and if wake reason is deepsleep could return straight to sleep (after flash init check - could be solar so gaining voltage slowly).

tyeth: @brentru when you get your battery for the magtag testing, then take a look at this PR

flavio-fernandes:
## MagTag battery-death test — secrets survived with PR #931

I flashed this branch onto a MagTag (ESP32-S2) and repeated the same test that proved the bug in beta.130: non-charging USB cable (TX/RX only, no VBUS) so the device runs purely on battery while logging to serial.

### Result

The fix held. After the battery died mid-operation and recovered, the device came back with `secrets.json` intact and connected normally to WiFi and AIO. No wipe occurred through five subsequent low-battery disconnect/reconnect cycles.

### Comparison

| | beta.130 | PR #931 |
|---|---|---|
| Last good disconnect | 01:17:03 EDT | 11:19:32 EDT |
| First message after recovery | `Fatal Error: Please edit the secrets.json file.` | `WipperSnapper found these WiFi networks...` |
| Recovery gap | 2 min 53 sec | 3 sec |
| Secrets intact? | **No** | **Yes** |

### Key log excerpt (PR #931)

```
[... ~2 hours of normal operation ...]
Sending MQTT PING: SUCCESS!
WiFi RSSI: -57
Sending MQTT PING: SUCCESS!
WiFi RSSI: -58
Sending MQTT PING: SUCCESS!
WiFi RSSI: -55
Sending MQTT PING: SUCCESS!
WiFi RSSI: -56
Sending MQTT PING: 
                               ← battery died mid-ping, device cut out

---------- 2026-06-25 11:19:32 -0400: Disconnected ----------


---------- 2026-06-25 11:19:35 -0400: Connected ----------

WipperSnapper found these WiFi networks:
[ssids redacted]
Connecting to WiFi (attempt #0)
Connecting to WiFi (attempt #1)
Connecting to WiFi (attempt #2)
Connecting to WiFi (attempt #3)
Connecting to WiFi (attempt #4)
Connected to WiFi!
WiFi Status: 20
Registering hardware with WipperSnapper...
...
GOT Registration Response Message:
Hardware Response Msg:
        GPIO Pins: 18
        Analog Pins: 5
        Reference voltage: 2.60v
Completed registration process, configuration next!
...
Registration and configuration complete!
                                        Running application...
```

The device resumed from where it left off — no halt, no format, no secrets loss. This continued through several more low-battery reconnects until logging was stopped manually.

---

I'll run a few more cycles to build confidence, but it's looking very promising. Great work on the `flashSafeToFormat()` approach — the partition table + NVS health check clearly catches what the reset-reason guard was missing.

src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -45,6 +45,12 @@ Adafruit_FlashTransport_SPI flashTransport(EXTERNAL_FLASH_USE_CS,
 // ESP32-S2 uses same flash device that stores code.
 // Therefore there is no need to specify the SPI and SS
 Adafruit_FlashTransport_ESP32 flashTransport;
+// The FAT user partition lives on the same on-chip flash as the ESP32
+// bootloader, partition table and NVS - so we can validate those known-good
+// structures to tell a blank/corrupt filesystem apart from unstable
+// (brownout) flash reads. See flashSafeToFormat().
+#define WS_USE_ESP32_INTERNAL_FLASH 1
```
brentru, line 52: @tyeth  Will we always define `WS_USE_ESP32_INTERNAL_FLASH` for _all_ ESP32 platforms?

src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -45,6 +45,12 @@ Adafruit_FlashTransport_SPI flashTransport(EXTERNAL_FLASH_USE_CS,
...
+#include <esp_flash.h>
```
brentru, line 53: @tyeth  Could we move the #include and #define here into FS.h, keeping the `ARDUINO_ARCH_ESP32` guard?

src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -87,6 +93,84 @@ bool setVolumeLabel() {
...
+              FAT-path board (external QSPI/SPI flash on SAMD51, or the on-chip
+              flash on RP2040/RP2350) has no such off-filesystem structure, so
+              the FAT volume's own boot sector is inspected instead - it may
+              only be formatted if it reads back genuinely erased (all 0xFF).
+    @returns  True if a (re)format is safe, False if the flash is unreadable /
+              unstable / not genuinely blank, in which case it must NOT be
+              formatted.
+*/
+/**************************************************************************/
+bool flashSafeToFormat() {
```
brentru, line 119: -> `isFlashSafeToFormat()`

src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -87,6 +93,84 @@ bool setVolumeLabel() {
...
+#ifdef WS_USE_ESP32_INTERNAL_FLASH
+  uint8_t buf[2];
+
+  // Partition table @ 0x8000 - the first entry must begin with the ESP-IDF
+  // partition magic 0x50AA (stored little-endian: 0xAA, 0x50).
```
brentru, line 124: The `0x8000` and `0x9000` default offsets may not always be where the NVS partition is located. 

 Instead, could you try locating the NVS by using the [esp partition API](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/storage/partition.html#_CPPv424esp_partition_find_first20esp_partition_type_t23esp_partition_subtype_tPKc) such as `esp_partition_find_first()` to locate the NVS? 

Example: https://github.com/espressif/esp-idf/blob/master/examples/storage/partition_api/partition_find/README.md

If you can't do this, or don't see success when working with this api - please stop and tag me in to help rather than moving forward.

src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -87,6 +93,84 @@ bool setVolumeLabel() {
...
+  uint8_t sector[256];
+  for (uint32_t addr = 0; addr < 512; addr += sizeof(sector)) {
+    if (flash.readBuffer(addr, sector, sizeof(sector)) != sizeof(sector))
+      return false; // could not read the volume -> unsafe to format
+    for (uint32_t i = 0; i < sizeof(sector); i++) {
+      if (sector[i] != 0xFF)
+        return false; // non-erased content -> not blank -> refuse to format
+    }
+  }
+  return true; // boot sector is fully erased -> genuinely blank -> safe
```
brentru, line 170: I'd read the JEDECID after the format occurs, and check its validity again.

src/provisioning/tinyusb/Wippersnapper_FS.cpp
```diff
@@ -87,6 +93,84 @@ bool setVolumeLabel() {
...
+  // 0xFFFFFFFC / FREEING 0xFFFFFFF8. The high byte is always 0xFF; the low
+  // byte distinguishes the state. Random garbage from a browning-out SPI bus
+  // is very unlikely to match this.
+  if (esp_flash_read(esp_flash_default_chip, buf, 0x9000, sizeof(buf)) !=
+      ESP_OK)
+    return false;
+  if (buf[1] != 0xFF ||
+      (buf[0] != 0xFF && buf[0] != 0xFE && buf[0] != 0xFC && buf[0] != 0xF8))
+    return false;
+
```
brentru, line 142: In the ESP32 branch, you only inspect NVS - not the FAT volume. Why wouldn't we inspect the FAT volume first, then the NVS?

brentru (changes requested):
@tyeth  I think, overall, this function and its reasoning is solid. It's been validated by @flavio-fernandes on-hardware as well. I do have some concerns - mostly about the hardcode defaults and difference within the function's branches.

src/provisioning/tinyusb/Wippersnapper_FS.cpp
tyeth: The partition table will always be OX8000 but that's potentially one kilobyte off and so the OX9000 check, although with custom ESP-IDF SDK config you can alter it. 
  I was a little bit curious how the ESPSDK functions would perform if there was an issue compared to the lower level flash attempt which can fail more gracefully when the flash chip or SPI bus misbehave or are unreadable. 
  It's pretty much definitely worth relying on the SDK if possible but it's not an easy thing to test so I initially went this direction for more reliability. 
 I really don't care what the NVS looks like, only the valid partition table, and its magic bytes being present means the flash is correctly readable as expected.

brentru: > The partition table will always be OX8000 

On **every** ESP32 MCU, and future MCUs?  This feels fragile.

flavio-fernandes:
## Update: secrets **were** wiped on PR #931 — a deeper drain got through

Bad news, and I want to correct my earlier optimism. After 5 clean sag/recovery cycles with secrets intact, a **deeper** battery-drain episode wiped the filesystem on this same `flash-health-format-guard` build. So #931 reduces the failure rate but does **not** close the hole. My earlier "fix held" comment was premature — apologies.

### The wipe window (SSIDs redacted)

Five shallow sags recovered cleanly (`11:55:05`, `11:56:58`, `11:57:18`, `11:57:33`, `11:57:47`) — each a full boot with secrets intact. Then a deep drain:

```
---------- 11:57:56 Disconnected ----------          ← deep drain begins
---------- 11:59:55 Connected ----------    (~2 min dead)
---------- 11:59:56 Disconnected ----------          ← brief flicker, browning out
---------- 11:59:58 Connected ----------
WipperSnapper found these WiFi networks: [redacted]
Connecting to WiFi (attempt #0)
---------- 12:00:03 Disconnected ----------          (~1.5 min dead)
---------- 12:01:33 Connected ----------
ERROR: Please edit the secrets.json file. Then, reset your board.   ← WIPED
```

From there it's the same beta.130 post-wipe loop (alternating *"Please edit the secrets.json file"* and *"Invalid IO credentials"*) through 13:12.

### Why it got through — this is exactly @brentru's line-142 note

The pattern that survived (shallow sag) vs. the one that wiped (deep drain) tells the story. `isFlashSafeToFormat()` on the ESP32 branch validates the **partition table (0x8000)** and **NVS (0x9000)** — regions that are never written during normal operation. Those always read back fine **as soon as the board has recovered enough voltage to boot and run the FS init at all.** They are effectively a "is the SPI flash bus alive right now" probe, not a "should we wipe the FAT" probe.

So on the deep cycle: the battery sagged enough to corrupt/interrupt a FAT write, then recovered enough to boot. By the time the constructor ran, 0x8000 and 0x9000 read perfectly → `isFlashSafeToFormat()` returned **true** → `f_mkfs()` ran → secrets gone. The check never looked at the FAT region itself — which is precisely what @brentru flagged:

> In the ESP32 branch, you only inspect NVS - not the FAT volume. Why wouldn't we inspect the FAT volume first, then the NVS?

### The deeper principle

**"The flash is readable" ≠ "the FAT should be reformatted."** Once the battery recovers, the flash is *always* readable — that's the whole problem. You cannot distinguish *blank-from-the-factory* from *transiently-corrupted-by-a-brownout* by reading regions that look identical in both cases. Reset reason (#929) and always-good-region reads (#931) are both proxies for the real question, and both leak.

The real question is: **has this device ever had a working, mounted filesystem?** If yes, a mount failure is corruption and we must **never** silently `f_mkfs` it. If no, it's a genuine first boot and formatting is correct.

### Proposed gate: a persistent "provisioned" marker in NVS

Record the one fact that actually answers the question, in a place engineered to survive both power loss and a FAT format — **NVS** (separate partition, power-fail-safe page state machine, untouched by `f_mkfs`). This also sidesteps the hardcoded-offset fragility you two were already discussing, since it uses the high-level `Preferences`/`nvs` API rather than magic addresses.

```cpp
// After a SUCCESSFUL mount (FS provably exists):
Preferences p; p.begin("ws", false);
p.putBool("fs_ok", true);
p.end();

// In the constructor, when initFilesystem() fails:
Preferences p; p.begin("ws", true);
bool wasProvisioned = p.getBool("fs_ok", false);
p.end();

if (wasProvisioned) {
  // We KNOW a good FS existed here. A mount failure now is brownout/
  // corruption, never "blank from factory". Do NOT format.
  fsHalt("Filesystem unreadable but this board was previously provisioned "
         "- likely a brownout. Recharge and reset; refusing to reformat to "
         "protect your secrets.json.");
}
// No marker -> genuine first boot -> safe to create the filesystem. [DESTRUCTIVE]
initFilesystem(true);
```

Behavior:
| Scenario | Outcome |
|---|---|
| Factory-fresh (no marker) | formats once → template secrets → after reset, mounts → marker set |
| Provisioned, brownout corrupts FAT | marker present → **refuse to format**, halt, conserve battery |
| Genuine first boot | formats ✅ |

The tradeoff is deliberate: a provisioned board that can't mount will now **halt and ask for a recharge/reset instead of wiping**. Failing safe (preserve data) beats failing destructive (silent wipe) — and in practice a full recharge can let a corrupted-but-largely-intact FAT remount. An explicit reformat gesture (button-hold, or erasing the NVS key) can cover the rare genuine-corruption case.

Defense-in-depth: keep the FAT **boot-sector blank check** the non-ESP32 branch already does (`format only if the volume reads back genuinely erased / all 0xFF`) and apply it to ESP32 too — that's the "inspect the FAT volume" part of @brentru's note, and it's a good second gate behind the NVS marker.

I'd keep @tyeth's flash-readable probe, but only for deciding whether to light the LED / attach USB / drop into the deep-sleep-and-retry path — not for the destructive format decision.

Happy to test whatever you land on. @brentru — you offered to help wire up the partition/NVS API; this might be a good moment for that, since the marker approach leans on it. I'll keep the rig set up and can turn builds around quickly.

flavio-fernandes:
Following up on the wipe I reproduced above — I put the proposed fix into a PR against this branch so it's concrete rather than just a wall of text: **#937** (targets `tyeth/flash-health-format-guard`).

It does two things on top of what's here:

1. **Closes the hole that let the wipe through.** `initFilesystem(force_format=false)` was still calling `f_mkfs()` on a mount failure, so the constructor's first `initFilesystem()` reformatted the corrupted FAT *before* `flashSafeToFormat()` ever ran — the guard was effectively dead code (which is why we saw `Please edit the secrets.json file`, not `Flash reads unstable`). #937 makes `initFilesystem()` return false on mount failure so only the guarded path can format.

2. **Adds the gate that actually answers the question.** A persistent `fs_ok` marker in NVS (survives brownout *and* a FAT format): if it's set and the FAT won't mount, that's corruption → refuse to format and protect secrets.json. The flash-health probe can't distinguish blank-from-factory from brownout-corrupted once the battery recovers, but this marker can.

It also folds in @brentru's review notes from this PR: `isFlashSafeToFormat()` rename, `esp_partition_find_first()` instead of the hardcoded `0x8000`/`0x9000`, inspecting the FAT boot sector on ESP32 too, the post-format JEDEC re-check, and moving the ESP32 includes into the header.

Builds clean for the MagTag. I'll flash #937 and re-run the same sacrificed-USB battery-drain test — and I'll keep hammering it until it stops making a fool of me. :) Totally your call on whether to merge it into this branch, cherry-pick, or just use it as a starting point — happy either way.

flavio-fernandes:
A consolidated write-up of what we've pinned down on hardware, and how this PR and the two that build on it (#937, #942) fit together. The short version: **the evidence directly backs the call made in this PR — you can't trust the reset reason — and it points to a clean three-layer fix.**

## What actually happens on a hard battery drain (MagTag, on the bench)

We instrumented a build to print the real reset reason in the protective halt, drained a provisioned MagTag, then dumped the flash to see exactly what survived.

1. **The reset reason is not reliable — confirmed.** On a full drain the board came back as `POWERON_RESET (1)`, **not** `BROWNOUT (15)`. A complete drain is a clean power-off then a cold boot, so the brownout detector never trips. This is exactly why gating on flash *health* rather than the reset reason (this PR) is the right call — we have the hardware to prove the reset reason would have lied.

2. **The damage is always one 4 KB block, and it's a flash glitch — not our code.** A flash dump showed exactly the first **4 KB** of the `ffat` partition (`0x310000`) erased to `0xFF`, with everything past it byte-perfect. No firmware path writes that region except `f_mkfs`/`f_setlabel` (format), and on the instrumented runs no format had run. The cause is the SPI-NOR flash taking a stray erase as Vcc collapses on power-down — a known unclean-shutdown failure mode. A read can't erase flash; nothing in our boot path erases that block.

3. **The user's data is intact — it's a failed *mount*, not corruption.** In the dump, `secrets.json` was a byte-perfect 206-byte valid JSON; the root directory and all files were fine. The board only failed to *mount*. (This lines up with @tyeth's framing earlier in this thread: a misread of a failed flash operation, not a corrupted file.)

4. **What's in that 4 KB matters for the fix.** It is not just the 512-byte boot sector — it's the **boot sector + both FAT tables** (the root directory and data start right after, at `0x1000`). FAT's two-FAT redundancy can't help (one erase takes both), and a FAT12 super-floppy has no backup boot sector. So the volume is unmountable until that one block is rebuilt.

5. **Reproducible on demand.** We can recreate the exact failure in seconds without a battery drain:
   ```
   esptool erase-region 0x310000 0x1000   # wipe boot sector + FATs
   ```

## How the three PRs layer up

Each one handles a distinct part of the problem; together they cover it end to end.

| PR | Role | One-line |
|----|------|----------|
| **#931** (this PR) | **Don't reformat on a bad signal** | Gate a reformat on actual flash health, not the (unreliable) brownout reset reason. |
| **#937** | **Don't reformat a known-good board at all** | `initFilesystem()` no longer auto-formats on a failed mount; a persistent NVS "this board was provisioned" marker means a previously-working board is never wiped, even if the reset reason looks innocent. |
| **#942** | **Heal it, don't just refuse** | After a clean mount, back up the 4 KB boot region to NVS; on a later boot, if that region is blank, restore it and re-mount — the board recovers itself with every file intact. |

The way to read the stack:

- **#931** stops the firmware from reformatting because of a *misread reset reason*.
- **#937** goes further: a board that has ever had a working filesystem is **never** reformatted on a mount failure, full stop — the worst case becomes "halt and protect the data" instead of "wipe it." (The NVS marker survives both a brownout and a format, so it's a reliable gate where the reset reason isn't.)
- **#942** turns that protective halt into an automatic recovery: it keeps a copy of the one block that gets erased and writes it back. It is **paranoid** — it only restores when the region is *entirely* `0xFF` and the backup passes a CRC32 + boot-signature check — so it can only help, never make things worse.

**Dependency note:** #942 only works on top of #937. On a plain `main` base, `initFilesystem()` reformats on the failed mount *before* the restore could run, so the boot region is gone before we look at it. #937's "return `false` instead of auto-format" is the precondition the rescue needs. I validated the two together on the branch `test/fs-937-plus-942-rescue` — the `erase-region 0x310000 0x1000` trick above wipes the FS on `main`+#942, but on the combined branch the board restores the boot region from NVS and boots straight back with `secrets.json` intact. A full battery-drain run is underway to confirm the real-world path.

Suggested merge order is the natural stack: **#931 → #937 → #942**.

Happy to fold any of this into the individual PR descriptions if that's easier to review. Thank you all for the careful back-and-forth on this one.

tyeth:
See comment with possible method not requiring nvm: https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/937#issuecomment-5372614782

PR #933, opened 2026-06-18
Backport 22 missing I2C sensor drivers to API v2

tyeth:
## Summary

Brings `migrate-api-v2` in line with v1 (`main`) and the `offline-mode` alpha by backporting **22 missing I2C sensor drivers** into the v2 `drvBase` style, ordered by the date each was originally added to v1. Also adds a `fastTick()` hook for sensors that need a fixed internal sampling cadence.

Each new driver documents candidate **custom settings** (gains, integration/conversion times, modes, calibration, etc.) in comments, ready to wire into the v2 per-device settings API.

## Sensors added

| Date added (v1) | Sensor | Driver | Metrics | Source |
|---|---|---|---|---|
| 2023-07-14 | SGP30 | `drvSgp30` | eCO2, TVOC | offline |
| 2023-09-27 | HTU31D | `drvHtu31d` | temp (+F), humidity | offline |
| 2024-08-23 | HDC302x | `drvHdc302x` | temp (+F), humidity | offline |
| 2024-12-19 | VCNL4200 | `drvVncl4200` | light, proximity | offline |
| 2024-12-23 | SEN66 | `drvSen6x` | temp (+F), humidity, PM1/2.5/10, VOC, NOx, CO2 | offline |
| 2025-02-19 | LPS28DFW | `drvLps28dfw` | temp (+F), pressure | offline |
| 2025-04-04 | INA260 | `drvIna260` | voltage, current | offline |
| 2025-06-29 | INA237 | `drvIna237` | voltage, current | offline |
| 2025-06-29 | INA238 | `drvIna238` | voltage, current | offline |
| 2025-07-02 | D6T-1A | `drvD6t1a` | temp (+F), object-temp (+F) | offline |
| 2025-08-13 | INA228 | `drvIna228` | voltage, current | offline |
| 2025-08-20 | MLX90632 | `drvMLX90632` | temp (+F), object-temp (+F) | offline |
| 2025-08-21 | AS5600 | `drvAs5600` | raw (angle) | offline |
| 2025-08-26 | QMC5883P | `drvQmc5883p` | raw (magnitude, gauss) | offline |
| 2025-08-27 | BMP5xx | `drvBmp5xx` | temp (+F), pressure, altitude | v1 |
| 2025-09-24 | SPA06-003 | `drvSpa06_003` | temp (+F), pressure | v1 |
| 2026-02-20 | SGP41 | `drvSgp41` | VOC index, NOx index, raw | v1 |
| 2026-03-24 | TMP119 | `drvTmp119` | ambient-temp (+F) | v1 |
| 2026-04-09 | AS7331 | `drvAs7331` | raw (UVB) | v1 |
| 2026-04-14 | STCC4 | `drvStcc4` | CO2, temp, humidity | v1 |
| 2026-04-14 | VCNL4030 | `drvVcnl4030` | light, proximity | v1 |
| 2026-04-15 | MAX44009 | `drvMax44009` | light (lux) | v1 |

(APDS9999 was already present in `migrate-api-v2`.)

## How they were ported

- **From `offline-mode`** (already `drv*` style): mechanical port; dropped the legacy `ConfigureDefaultSensorTypes()` (v2 enables sensor types from the cloud/offline config).
- **From v1** (`WipperSnapper_I2C_Driver_*`): rebased onto `drvBase` -- 4-arg constructor, `_address`, `getEvent*` handlers; dropped v1-only base machinery; `ONE_SECOND_IN_MILLIS` -> `1000`, `ulong` -> `unsigned long`.
- Registered every driver in `controller.h` / `controller.cpp` `I2cFactorySensor`. BMP5xx is registered under `bmp580`/`bmp581`/`bmp585`; VCNL4030/VCNL4200 under both the correct `vcnl*` id and the legacy-misspelled `vncl*` to match existing sibling keys.
- Libraries added to `library.properties` (CI-critical) and `platformio.ini`. TMP119 uses the existing Adafruit TMP117 library; SGP41's gas-index algorithms ship inside the Adafruit SGP41 library.
- Copyright year set to **2026** on all backported drivers.

## Architecture notes

**ESP8266 build fix** -- `WS_DEBUG_PRINTLN` already wraps its argument in `F()` on ESP8266 (`ws_helper_macros.h`); the ported HDC302x/MLX90632 drivers passed `F("...")`, producing `F(F(...))` (a hard `PSTR` error on 8266, silently fine on ESP32). Fixed to plain string literals, matching the existing convention. `WS_DEBUG_PRINTVAR`/`WS_DEBUG_PRINTLNVAR` are used for variable logging.

**Single-poll read model** -- per v2 design (one `_sensor_period` per device; the controller reads all enabled metrics in one pass), removed the hardcoded per-metric `1000ms`/`200ms` cache guards copied from v1/offline. On-demand sensors read fresh in each getEvent; sensors that yield all metrics from one transaction (STCC4, MLX90632, D6T-1A, SEN6X) read-all in one method and distribute, with data-ready-gated parts returning the last good sample for repeat metric reads within a pass instead of failing the whole device read.

**`fastTick()` hook** -- for sensors that need a fixed internal sampling cadence independent of the publish period:
- `drvBase`: `virtual void fastTick() {}` (no-op default).
- `I2cController::update()` calls `drv->fastTick()` once per loop iteration (before the publish-period gate), mirroring the v1 `WipperSnapper_I2C` loop.
- **SGP30 / SGP40 / SGP41** sample at ~1 Hz inside `fastTick()` (millis-guarded, non-blocking) and cache results; their `getEvent*` return the cached values. This is required for the Sensirion gas-index conditioning/algorithms to settle regardless of how often metrics are published.

## Out of scope (tracked for follow-up)

I2C **output** components and **displays** -- see `docs/COMPONENT_BACKPORT_PLAN.md`, which tracks the full backport and remaining work.

## Testing

Full CI build matrix green (ESP8266, ESP32/Cx/Sx incl. all `_debug` variants, RP2040, SAMD) on the latest commit.


src/components/i2c/drivers/drvStcc4.h
```diff
@@ -0,0 +1,153 @@
...
+    //  - Single-shot vs continuous measurement mode.
+  }
+
+  /*******************************************************************************/
+  /*!
+      @brief    Reads all sensor data from the STCC4, caching the results.
+                Only performs a new I2C read if the last read was more than
+                one second ago.
+      @returns  True if the sensor was read successfully, False otherwise.
+  */
```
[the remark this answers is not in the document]
tyeth: Definitely should be implemented, and additionally mention in docstring if i2c read or <1s since, then cached members left unchanged.

src/components/i2c/drivers/drvStcc4.h
tyeth, line 1: The cached STCC4 readings are initialized to 0, and the getEvent* handlers always return true even when ReadSensorData() fails. Before the first successful measurement this can publish bogus 0 values instead of reporting a read failure.
src/components/i2c/drivers/drvStcc4.h:128
getEventRelativeHumidity() always returns true even if no valid humidity sample has been read yet. Add a validity check so the first publish doesn’t report a bogus value.
src/components/i2c/drivers/drvStcc4.h:143
getEventCO2() always returns true even if no valid CO2 sample has been read yet. Guard against the initial invalid cache value so the first publish doesn’t report a bogus 0 ppm reading.

src/components/i2c/drivers/drvMlx90632.h
```diff
@@ -0,0 +1,313 @@
...
+    mlx90632_mode_t currentMode = _mlx90632->getMode();
+    if (currentMode == MLX90632_MODE_STEP ||
+        currentMode == MLX90632_MODE_SLEEPING_STEP) {
+      // Trigger single measurement (SOC bit) for step modes
+      if (!_mlx90632->startSingleMeasurement()) {
+        WS_DEBUG_PRINTLN("Failed to start single measurement");
+        return false;
+      }
+      delay(510); // Wait for measurement to complete @ 2Hz
+    }
```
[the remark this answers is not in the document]
tyeth: The delay is expected and acceptable for Step Mode only (and sleep step mode), wherein the data request is always fresh and use case demands waiting for result at nearly any time cost.

Improved delay logic to match datasheet now that we have mode control exposed.

tyeth:
Review of existing V1 fastticks / data ready / first read behaviour that may have been missed in v2 during parallel evolution:

<details><summary>Details</summary>
<p>

I've now got the full picture from both sides — v1 main (base driver + update() loop + the evolved drivers) and v2 (migrate-api-v2, the backport branch for PR [#933](https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/933), and api-v2-pins-as-strings). Here's the discussion.

The v1 model, and how it evolved

v1's cadence machinery lives in WipperSnapper_I2C.cpp:1760 (sensorEventRead) and the base class, and there are really five separate ideas layered on over time:

1. Per-metric period + backdated "prv". Every metric has its own _xxxSensorPeriod / _xxxSensorPeriodPrv, and the prv values are initialised to PERIOD_24HRS_AGO_MILLIS (WipperSnapper_I2C_Driver.h:24). That backdating is deliberate: the first pass through update() is always "period elapsed", so the user sees a datapoint immediately after configuring, rather than waiting out one full period.

2. Retry-with-backoff at the loop level. update() wraps everything in a while (sensorsReturningFalse && retries > 0) loop (3 tries). If a getEvent* returns false it retries the whole pass; on the final failed retry it stamps PeriodPrv anyway (WipperSnapper_I2C.cpp:1802), so a sensor that isn't ready gets a couple of quick chances and then backs off a full period instead of being hammered every loop iteration. This is the loop-level answer to "the first read often fails because data isn't ready yet."

3. The ReadSensorData() / HasBeenReadInLastSecond() / IsSensorReady() triad. Because v1 metrics have independent periods, temp/humidity/CO2 for one device can all fire in the same pass, each calling its own getEvent*. The standardised answer (SCD4X, SCD30, SEN5X, SEN6X, D6T1A, MLX90632D, TMP119, STCC4, MAX44009 all carry the same doc-comment "read within last 1s, or is the first read") is:

one physical ReadSensorData() that reads all metrics in a single transaction into cached members,
a _lastRead guard (_lastRead != 0 && millis() - _lastRead < ONE_SECOND_IN_MILLIS) so the second and third getEvent* in the same pass hit the cache — with _lastRead == 0 as the "never read yet" sentinel (MAX44009 zeroes it to force a re-read after a config change),
an IsSensorReady() data-ready poll where the silicon supports it — SCD4X gives it two chances 100 ms apart (WipperSnapper_I2C_Driver_SCD4X.h:86) and returns false rather than reading stale registers,
crucially, failures propagate: getEvent* returns false if ReadSensorData() fails, and the loop's retry/backoff machinery deals with it. Values are only cached on a successful read, and only in-sync samples are ever published.

4. First-sample validity. Several distinct flavours, all serving "don't publish garbage":

Discard-at-begin: HDC302X throws away the first conversion (it once returned −45 °C); NAU7802 flushes stale FIFO datapoints; VL53L4CD/CX restart ranging because the first value was stale.
NAN gating: SEN5X/SEN6X return NAN for the first 10–11 s after power-on; the getEvents check isnan() and return false, letting the backoff handle it.
Sanity checks in the read: SCD4X rejects co2 == 0 as "not a real sample."

5. fastTick(). The newest idea (WipperSnapper_I2C_Driver.h:70, called at WipperSnapper_I2C.cpp:1466): a non-blocking, once-per-loop hook decoupling internal sampling cadence from publish period. SGP30/40/41 need their gas-index algorithms fed at ~1 Hz regardless of whether the user publishes every 5 minutes; SGP41 additionally uses it for the 10-cycle NOx conditioning warm-up. The driver caches results and getEvent* becomes a pure cache accessor.

What v2 does, structurally

v2 inverted the loop: one _sensor_period per driver, and I2cController::update() fires one pass per device — iterate every enabled sensor type through the SensorEventHandlers map, build one I2cDeviceEvent, publish/log, stamp SetSensorPeriodPrv() once. That's genuinely better than v1's per-metric machinery (one message, samples in sync by construction, no 25-way sensorEventRead boilerplate), and it means the reason for the 1-second cache guard largely disappears — all getEvents for a device happen microseconds apart in one pass.

But the v1 lessons got lost around the edges:

No first-read backdating — and worse, UB. _sensor_period_prv is never initialised (not in the drvBase ctor, no member initialiser — drvBase.h line 881). In practice fresh heap is usually zero, so cur_time - 0 makes the first read fire once millis() exceeds the period — meaning a 5-minute period configured at boot publishes nothing for 5 minutes, where v1 published immediately. And formally it's an indeterminate read.
No retry cap / backoff. On a failed pass the controller does continue without stamping PeriodPrv, so a not-ready sensor is retried every single loop iteration, forever — the exact hammering v1's 3-retries-then-stamp was built to prevent. There's also no distinction between "not ready yet" (normal for Sensirion parts) and "actual error."
All-or-nothing device events. One failing metric (read_succeeded = false) drops the entire device event, including the metrics that read fine. This creates a perverse incentive for drivers to lie…
…and the pre-existing v2 drivers do lie. drvScd4x is the "older ideas" fossil: readSensorMeasurements() is called from each of the three getEvents (three blanket delay(100)s and up to three physical reads per pass), its return value is discarded, and each getEvent returns true unconditionally — publishing uninitialised members on the first pass if data wasn't ready, and desynced samples later.
_did_read_send one-shot semantics. SetDidReadSend(true) after publish is only ever cleared by ResetFlags(), which is only called from the loopSleep() paths. In the plain continuous loop() path nothing resets it, which reads like each I2C driver publishes exactly once per boot in non-sleep mode. Presumably testing has been sleep/offline-centric so this hasn't bitten, but it's worth verifying before you build on update() — either the guard should be sleep-mode-only or continuous mode needs a reset. (The period check is redundant with it as written: the flag blocks long before the period ever matters.)
What the backport PR carried, and where it's inconsistent

PR [#933](https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/933) did carry the newest v1 idea — fastTick() is in drvBase and called at the top of the controller loop, before the GetDidReadSend bail-out, which is the right placement (SGP conditioning keeps running after the read/send completes). The custom-settings virtuals came along too.

But the drivers are a snapshot of two eras running in parallel, exactly as you said:

drvSen6x is the good template: single-transaction ReadSensorData(), non-blocking IsSensorReady() gate, _haveData first-read bool (returns false until the first successful sample, so nothing bogus is ever published), NAN-init + isnan() guards, cache untouched when no new data.
drvStcc4 regressed against its own v1 counterpart: the _lastRead/1-second guard was dropped (the doc comment still claims it exists — it lies), ReadSensorData() is called three times per pass with its return value swallowed, every getEvent returns true unconditionally, and the cached members init to 0.0f — so before the first successful sample it publishes CO2 = 0 ppm and temp = 0 °C to IO. The v1 version at least had the read-once guard and honest failure returns.
drvScd4x (pre-existing) is worse still, per above.
What I'd carry forward — into the base/controller, not per-driver

Since every metric shares one period in v2, the standardisation opportunity is better than v1 ever had: the pattern belongs in drvBase + controller, not copy-pasted into 60 drivers.

Give drvBase a virtual bool ReadSensorData() (default return true;) and have the controller call it once per due pass, before the getEvent loop. getEvents become pure cache accessors. That structurally eliminates the per-getEvent re-read problem, the desync problem, and the need for each driver to reinvent HasBeenReadInLastSecond — though keeping a base-class _lastRead guard inside it is cheap insurance against force updates and sleep-cycle double passes.
Standardise _haveData (the SEN6X first-read bool) in the base. ReadSensorData() returns false until the first valid sample; the controller treats that as "not ready," not "error." Kills the publish-zeros class of bug for every driver at once.
Restore backoff semantics in the controller: bounded quick retries (or even just v1's stamp-PeriodPrv-on-give-up), so not-ready sensors wait out a period instead of being polled every loop iteration. Consider letting a partial pass publish the metrics that succeeded rather than dropping the whole event — that removes the incentive for drivers to return true dishonestly.
Initialise _sensor_period_prv backdated (v1's PERIOD_24HRS_AGO_MILLIS trick) in the drvBase ctor — fixes both the UB and "first datapoint appears immediately," which users notice.
Keep IsSensorReady() as an optional non-blocking virtual for the Sensirion-style parts, and keep the begin()-time quirk handling (HDC302X discard, VL53 ranging restart, SGP41 conditioning via fastTick) as per-driver overrides — those are genuinely device-specific.
While you're in there: resolve the _did_read_send continuous-mode question, and fix drvScd4x to the same template rather than leaving the fossil next to the new pattern (it's the driver people copy from).

If you do 1–4 in drvBase/controller first, the STCC4 fix becomes trivial — it's ~15 lines of driver instead of another bespoke copy of the pattern, and every other backported driver inherits the same behaviour. That's the same trajectory v1 took (each idea got added to the base after being proven in one driver), just finished properly this time.

</p>
</details>

tyeth:
Summary of changes incorporated from V1:

<details><summary>Details</summary>
<p>

Committed
Merge of migrate-api-v2 (71 commits, incl. pins-as-strings + SCD43 split) into migrate-api-v2-backport-components (b4f0fca0). Conflict resolution: took upstream's ws_pb_decode_oneof_submsg for the Probe/Add decode (dropped the branch's duplicate local decodeI2cB2DMember), kept your I2C_STEMMA_WIRE1 bus-instance block now calling new I2cHardware(sda.num, scl.num, bus_instance).
In the working tree (uncommitted, for you) — v1 read semantics carried into v2

drvBase.h — virtual bool ReadSensorData() (default no-op true), HasBeenReadInLastSecond(), NoteReadFailure()/NoteReadSuccess() (3-strikes counter mirroring v1's retry loop), protected _last_read/_have_data/_read_fails, and the ctor now inits _sensor_period = 0 and _sensor_period_prv = PERIOD_24HRS_AGO_MILLIS — fixes the uninitialised-member UB and restores v1's immediate-first-publish. No new define names invented: ONE_SECOND_IN_MS is the existing v2 name (with an #ifndef fallback), PERIOD_24HRS_AGO_MILLIS is v1's name verbatim with a docstring noting the sync source. The method names NoteReadFailure/BackoffDriverRead are mine — rename freely.

controller.cpp — update() now calls drv->ReadSensorData() once per due pass before the getEvent loop; a failed pass goes through BackoffDriverRead() (retry in ~1s ×3, then wait a full period — no more every-loop hammering of not-ready sensors); per-metric failures now log a warning and publish the metrics that did read instead of dropping the whole device event; only an all-metrics-failed pass backs off.

Drivers converted to the template: drvStcc4 (regains the 1s guard it lost vs v1, honest returns, NAN-init — supersedes your stash, which I left untouched), drvScd4x (fossil rebuilt: no more delay(100)×3 per pass, non-blocking data-ready gate, co2==0 rejection, no more publishing uninitialised members — and this now serves scd40/41/43), drvSen5x (was 8 full reads per pass with the == NAN always-false bug; now one cached read + isnan gates), drvSen6x/drvScd30/drvDps310 (moved onto base members, local duplicates removed), drvD6t1a/drvHdc302x/drvMlx90632 (1s guard, _have_data, NAN-init for first-pass garbage).

Verified

featheresp32 builds SUCCESS from full clean, and again after the clang-format v20 pass (RAM 16.3% / Flash 83.4%). The repeated "failures" were never our code: the log tail only ever showed the gutted tool-esptoolpy package plus idf_tools.py refusing MSYS — fixed by removing the package via penv python and clearing MSYSTEM* when invoking pio (the metadata-only esptoolpy zip is by design; tools.json fetches the real tool).

Flagged, not touched
_did_read_send is never reset in continuous (non-sleep) loop() — drivers look publish-once-per-boot outside loopSleep(). Systemic across controllers, deserves its own fix/PR against migrate-api-v2.
drvUartBase has the same uninitialised _sensor_period_prv.

</p>
</details>

.vscode/settings.json
```diff
@@ -23,6 +23,6 @@
     "cmake.configureOnOpen": false,
     "workbench.panel.defaultLocation": "right",
     "terminal.integrated.defaultLocation": "view",
-
-
+    "C_Cpp.formatting": "clangFormat",
+    "C_Cpp.clang_format_style": "{ BasedOnStyle: LLVM, IndentCaseLabels: false, ColumnLimit: 80, SpacesBeforeTrailingComments: 1, ReflowComments: true }"
```
brentru: @tyeth  Hm this feels not required? 

How about we add to the `.gitignore` for `.vscode/settings/json` so our local dev configs aren't committed?

docs/COMPONENT_BACKPORT_PLAN.md
```diff
@@ -0,0 +1,144 @@
+# Component Backport Plan: `migrate-api-v2`
```
brentru: I don't think we need this backport plan committed either

src/components/i2c/drivers/drvAs5600.h
```diff
@@ -0,0 +1,389 @@
...
+                The slow filter index from the broker
+                (0=16x, 1=8x, 2=4x, 3=2x).
+      @returns  True if applied successfully, False otherwise.
+  */
+  /*******************************************************************************/
+  bool setSlowFilter(const ws_config_Value &slow_filter) override {
+    if (slow_filter.which_value != ws_config_Value_int_value_tag) {
+      return false;
+    }
+    as5600_slow_filter_t filter;
```
brentru: Assign a filter to this by the `AS5600_SLOW_FILTER_16X` [default](https://github.com/adafruit/Adafruit_AS5600/blob/main/Adafruit_AS5600.h#L85) to avoid calling `setSlowFilter()` with a initialized variable

src/components/i2c/drivers/drvAs5600.h
```diff
@@ -0,0 +1,389 @@
...
+                5=21 LSB, 6=24 LSB, 7=10 LSB).
+      @returns  True if applied successfully, False otherwise.
+  */
+  /*******************************************************************************/
+  bool setFastFilterThreshold(
+      const ws_config_Value &fast_filter_threshold) override {
+    if (fast_filter_threshold.which_value != ws_config_Value_int_value_tag) {
+      return false;
+    }
+    as5600_fast_filter_thresh_t thresh;
```
brentru: Same here, assign the default before dropping into the `switch`:
https://github.com/adafruit/Adafruit_AS5600/blob/main/Adafruit_AS5600.h#L95

src/components/i2c/drivers/drvAs5600.h
```diff
@@ -0,0 +1,389 @@
...
+                The hysteresis index from the broker
+                (0=Off, 1=1 LSB, 2=2 LSB, 3=3 LSB).
+      @returns  True if applied successfully, False otherwise.
+  */
+  /*******************************************************************************/
+  bool setHysteresis(const ws_config_Value &hysteresis) override {
+    if (hysteresis.which_value != ws_config_Value_int_value_tag) {
+      return false;
+    }
+    as5600_hysteresis_t hyst;
```
brentru: Same here
https://github.com/adafruit/Adafruit_AS5600/blob/main/Adafruit_AS5600.h#L55

src/components/i2c/drivers/drvAs5600.h
```diff
@@ -0,0 +1,389 @@
...
+                The zero position from the broker (valid range 0..4095).
+      @returns  True if applied successfully, False otherwise.
+  */
+  /*******************************************************************************/
+  bool setZPosition(const ws_config_Value &z_position) override {
+    if (z_position.which_value != ws_config_Value_int_value_tag) {
+      return false;
+    }
+    int32_t val = z_position.value.int_value;
+    if (val < 0 || val > 4095) {
```
brentru: `4095` is a magic number and reused at least 3 times, #define it.

src/components/i2c/drivers/drvAs7331.h
```diff
@@ -0,0 +1,334 @@
...
+                The gain index from the broker
+                (0=2048x, 1=1024x, 2=512x, 3=256x, 4=128x, 5=64x, 6=32x,
+                7=16x, 8=8x, 9=4x, 10=2x, 11=1x).
+      @returns  True if applied successfully, False otherwise.
+  */
+  /*******************************************************************************/
+  bool setGain(const ws_config_Value &gain) override {
+    if (gain.which_value != ws_config_Value_int_value_tag)
+      return false;
+    as7331_gain_t g;
```
brentru: See my comment on `drvAs5600` about default values for typedefs.

src/components/i2c/drivers/drvAs7331.h
```diff
@@ -0,0 +1,334 @@
...
+                (0=1ms, 1=2ms, 2=4ms, 3=8ms, 4=16ms, 5=32ms, 6=64ms, 7=128ms,
+                8=256ms, 9=512ms, 10=1024ms, 11=2048ms, 12=4096ms, 13=8192ms,
+                14=16384ms).
+      @returns  True if applied successfully, False otherwise.
+  */
+  /*******************************************************************************/
+  bool setIntegrationTime(const ws_config_Value &integration_time) override {
+    if (integration_time.which_value != ws_config_Value_int_value_tag)
+      return false;
+    as7331_time_t t;
```
brentru: See my comment on `drvAs5600` about default values for typedefs.

src/components/i2c/drivers/drvAs7331.h
```diff
@@ -0,0 +1,334 @@
...
+                The measurement mode index from the broker
+                (0=continuous, 1=command, 2=synchronized start,
+                3=synchronized data).
+      @returns  True if applied successfully, False otherwise.
+  */
+  /*******************************************************************************/
+  bool setMode(const ws_config_Value &mode) override {
+    if (mode.which_value != ws_config_Value_int_value_tag)
+      return false;
+    as7331_mode_t m;
```
brentru: See my comment on `drvAs5600` about default values for typedefs.

src/components/i2c/drivers/drvAs7331.h
```diff
@@ -0,0 +1,334 @@
...
+      @param    clock_frequency
+                The clock frequency index from the broker
+                (0=1.024MHz, 1=2.048MHz, 2=4.096MHz, 3=8.192MHz).
+      @returns  True if applied successfully, False otherwise.
+  */
+  /*******************************************************************************/
+  bool setClockFrequency(const ws_config_Value &clock_frequency) override {
+    if (clock_frequency.which_value != ws_config_Value_int_value_tag)
+      return false;
+    as7331_clock_t c;
```
brentru: See my comment on `drvAs5600` about default values for typedefs.

src/components/i2c/drivers/drvStcc4.h
brentru: @tyeth  yeah I spotted this too - was this resolved in 993 or 992?

brentru (changes requested):

PR #936, opened 2026-06-24
Add LilyGo T-Display-S3 support with i8080 ST7789 display (v2)

tyeth-ai-assisted:
## Summary
Adds the standard **LilyGo T-Display-S3** (ESP32-S3 N16R8, 1.9" 170×320 IPS ST7789 driven over an **8-bit i8080 parallel** bus) to the v2 display stack. The v2 proto already defined `ws_display_I8080PinDescriptor`; this fills in the firmware driver that was previously a TODO in `hardware.cpp`.

## Changes
- **New driver** `src/components/display/drivers/dispDrvSt7789I8080.h` — uses Arduino_GFX `Arduino_ESP32LCD8` (i8080 8-bit bus) + `Arduino_ST7789`, guarded to `CONFIG_IDF_TARGET_ESP32S3`. Native 170×320, column offset 35 / row offset 0 (panel centred in the controller's 240×320 RAM), IPS/inverted — matching the [CircuitPython `lilygo_tdisplay_s3` board.c](https://github.com/adafruit/circuitpython/blob/main/ports/espressif/boards/lilygo_tdisplay_s3/board.c). Reuses the existing status-bar/`writeMessage` pattern.
- **Display hardware** — `beginI8080()` added and the `ws_display_InterfaceDescriptor_i8080_tag` case wired up (`hardware.cpp`/`.h`). WR/RD strobes and panel power/backlight are board-fixed (not carried in the i8080 proto), sourced from `I8080_*` macros.
- **Board** — `ARDUINO_LILYGO_T_DISPLAY_S3` block in `ws_boards.h` and a `lilygo_t_display_s3` PlatformIO env. Overrides `ARDUINO_USB_MODE=0` + `USE_TINYUSB` (board default is native USB-JTAG) so the WIPPER provisioning drive enumerates; 16MB tinyuf2 partitions.
- **Provisioning** — added `ARDUINO_LILYGO_T_DISPLAY_S3` to the `Wippersnapper_FS` TinyUSB board allowlist (otherwise the FS symbols don't link).
- **CI** — dedicated `build-lilygo-tdisplays3` job that checks out `ci-arduino` at ref **`ci-wippersnapper-tdisplays3`** (platform key `lilygo_t_display_s3`).

## Testing
`pio run -e lilygo_t_display_s3` builds successfully (Flash 81.1%, RAM 20%). On-device display verification still pending.

## Companion PR
Board definition + magic.json (and `magic_schema.json` i8080 fields) in adafruit/Wippersnapper_Boards, targeting `api-v2`.

> Note: the `build-migrate-api-v2.yml` workflow only triggers on pushes to `migrate-api-v2`, so the new job will first execute once this is merged.


tyeth-ai-assisted:
## ✅ HIL verification — LilyGo T-Display-S3 i8080 ST7789 (real hardware)

Flashed + driven on the `usbip-hil-controller` (device `mcu-lilygo-tdisplay-s3-hil006`, rpi-hil006). Verdict lines from the firmware-bench run:

```
CHECKIN_VERDICT ok=true uid=lilygo-t-display-s321621634 proto=v2
INJECT_VERDICT  published=true kind=display_add_i8080 resp={'status': 'OK'}
INJECT_VERDICT  published=true kind=raw (display Write) resp={'status': 'OK'}
DISPLAY_CAPTURE_VERDICT saved=… exposure_us=6000 cropped=yes
```

Serial confirms the new driver brought up the panel:
```
[display] Adding display: tft-19-i8080 / Driver: ST7789
[display] Backlight ON / ST7789 (i8080) initialized
[display] i8080 ST7789 initialized successfully!
```

**Camera proof** (CSI imx708 on rpi-hil006, autofocus + manual exposure 6000µs + ROI crop) — the WS status bar (cloud/wifi/battery icons) plus a display `Write`:

LilyGo T-Display-S3 panel — WS UI + text

Pipeline: `flash (tinyuf2+WS combined) → power_cycle → write_secrets_msc → power_cycle → verify_checkin (v2) → inject_protobuf(display Add via /api/echo) → inject_protobuf(display Write) → capture_display`. The committed `.github/workflows/hil-lilygo-display.yml` automates exactly this and uploads the camera image as a job artifact (needs `HIL_API_BASE`/`HIL_API_TOKEN` + a bench-reachable runner).

<sub>Posted by automated HIL verification (tyeth-ai-assisted).</sub>

tyeth-ai-assisted:
### 🔆 Brighter camera proof (re-run)

The earlier proof came out very dark. Root cause: the `capture_display` capture
was **6000 µs @ gain 1.0** — far too dim for the self-lit i8080 TFT on an
otherwise dark bench — and the ROI was loose, so the lit panel was a small dim
patch in a mostly-black crop. White-balance was uncontrolled (imx708 green cast).

Tuned on real hardware (csi-rpi-hil006 imx708) by sweeping exposure / autofocus /
white-balance against the live panel:

- **Exposure 32000 µs, gain 3.0** (was 6000 µs / 1.0)
- **Autofocus converged at 1.40 dioptre, then locked to manual** — continuous AF
  drifted during the still grab and blurred the text
- **White-patch white balance** on the crop (off the lit text) — neutralises the
  green cast; the camera-server itself ignores WB query params
- **Tighter ROI** `1270,770,235,135 @2304` (scales to the 4608 full-res frame)

Re-run end-to-end through the HIL pipeline (flash → v2 check-in → inject display
Add + Write → capture):

```
CHECKIN_VERDICT ok=true uid=lilygo-t-display-s321621634 proto=v2
INJECT_VERDICT published=true kind=display_add_i8080 resp={'status': 'OK'}
INJECT_VERDICT published=true kind=raw (display Write) resp={'status': 'OK'}
DISPLAY_CAPTURE_VERDICT saved=... exposure_us=32000 gain=3.0 focus=1.400 wb=yes frame=4608x2592 cropped=yes
```

LilyGo T-Display-S3 i8080 ST7789 — bright HIL camera proof

The panel clearly shows the status bar + `LilyGo T-Display-S3 / WipperSnapper v2 /
i8080 ST7789 / HIL camera proof OK`. Capture defaults updated in
`.github/scripts/hil_lilygo_display.py` and the controller `capture_display` stage.

src/ws_boards.h
```diff
@@ -244,6 +244,18 @@
...
+#elif defined(ARDUINO_LILYGO_T_DISPLAY_S3)
+#define BOARD_ID "lilygo-t-display-s3"
+#define USE_TINYUSB
+#define USE_PSRAM     ///< Board has PSRAM, use it for dynamic memory allocation
+#define BOOT_BUTTON 0 ///< BUTTON_1 (GPIO0) is the boot button
+// On-board ST7789 (170x320) driven over an 8-bit i8080 parallel bus.
+// WR/RD are not carried in the display Add proto, so the i8080 driver sources
+// them (plus panel power/backlight) from these board macros.
+#define I8080_WR LCD_WR             ///< i8080 write-strobe pin (GPIO8)
+#define I8080_RD LCD_RD             ///< i8080 read-strobe pin (GPIO9)
```
tyeth: Need these in the protobuf:
* Backlight pin
* i8080_wr
* i8080_rd
* display_power_pin optional (existing pin proto ideally with inversion but not needed for this one)

platformio.ini
```diff
@@ -306,6 +306,24 @@ lib_deps =
...
+board = lilygo-t-display-s3
+build_flags =
+    -DARDUINO_LILYGO_T_DISPLAY_S3
+    -DBOARD_HAS_PSRAM
+    -DUSE_TINYUSB
+    ; Board default is USB_MODE=1 (HW CDC+JTAG); force the TinyUSB stack so the
+    ; WIPPER provisioning drive enumerates, matching other ESP32-S3 WS boards.
+    -DARDUINO_USB_MODE=0
+    -DARDUINO_USB_CDC_ON_BOOT=1
+board_build.partitions = tinyuf2-partitions-16MB.csv
```
tyeth: Double check CI matches

tyeth (changes requested):

src/components/display/drivers/dispDrvSt7789I8080.h
```diff
@@ -0,0 +1,314 @@
...
+    }
+    // Arduino_GFX does not own/free the data bus, so release it explicitly.
+    if (_bus) {
+      delete _bus;
+      _bus = nullptr;
+    }
+    if (_pin_bl >= 0)
+      digitalWrite(_pin_bl, LOW);
+    if (_pin_power >= 0)
+      digitalWrite(_pin_power, LOW);
```
tyeth: TODO: ideally route through digitalio/pwm controller, so backlight can be controllable.

tyeth-ai-assisted: Recorded the TODO at both backlight touch points (begin + destructor) in 0a3b945e, pointing at the existing proto hook — ws_display_Add.backlight (BacklightConfig, oneof digitalio/pwm Add) — as the intended wiring; the driver keeps direct-drive until that controller routing lands.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/ws_boards.h
tyeth-ai-assisted: Done — every display pin now rides the protobuf, and the `I8080_*` board macros are gone from ws_boards.h:

- `I8080PinDescriptor` gains `pin_wr = 12` / `pin_rd = 13` (RD optional) — proto change in adafruit/Wippersnapper_Protobuf#208, nanopb regenerated here.
- **Backlight** comes through the existing `Add.backlight` (`BacklightConfig`, digitalio/pwm oneof).
- **Panel power** is a new optional `Add.pin_power = 12` (last field), driven high before init.
- `beginI8080` reads all of them from the Add; the boards-repo magic config (`d0–d7/cs/dc/wr/rd/rst/backlight/powerOn`) now maps 1:1 onto the proto.

Verified end-to-end on the HIL bench with the new 123-byte Add: check-in ok (v2), `Backlight ON` + `i8080 ST7789 initialized successfully!` over serial with zero board-hardcoded pins, camera proof captured.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

platformio.ini
tyeth-ai-assisted: Good catch — CI did NOT match. The ci-arduino fqbn (`PartitionScheme=tinyuf2_noota`) was invalid for this board: the upstream `lilygo_t_display_s3` entry in bsp-3-3-10 has no tinyuf2 partition menu (only `rainmaker`) and defaults to `USBMode=hwcdc` (`usb_mode=1`), while this env builds `ARDUINO_USB_MODE=0` + TinyUSB (the WIPPER MSC drive).

Fixed on ci-arduino `ci-wippersnapper-tdisplays3` (1830f30): fqbn is now `espressif:esp32:lilygo_t_display_s3:USBMode=default` (USB-OTG/TinyUSB, matching this env; CDCOnBoot already defaults to 1). Partitions stay the board default for the compile check — the flashable image takes its real partition table from the official tinyuf2 bootloader at merge time — and the workflow's `.uf2` rename is now tolerant since the upstream board entry has no uf2 step. TODO left in ci-arduino: add a tinyuf2 16MB scheme for this board to the Adafruit BSP so CI size-checks against the true 2MB ota_0 slot.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

tyeth-ai-assisted: Follow-up: the TinyUF2 16MB schemes (+ variant bootloader-tinyuf2/tinyuf2 bins) landed on the BSP (adafruit/arduino-esp32 f612cd7f on bsp-3-3-10), so ci-arduino (ed538d9) now builds with `PartitionScheme=tinyuf2_noota,USBMode=default` and the workflow's uf2 rename is strict again (20be9ecc) — same shape as every other WipperSnapper board. Note the platformio env keeps the with-OTA `tinyuf2-partitions-16MB.csv` (2MB app size-check) since the HIL combined image flashes the official tinyuf2 release table (ota_0/ota_1); CI's no-OTA check allows 4MB.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

tyeth-ai-assisted: Exact parity now: the pio env builds with `tinyuf2-partitions-16MB-noota.csv` and the HIL workflow merges **this build's** `partitions.bin` into the combined image (a60aea42) — so pio, the CI fqbn (`PartitionScheme=tinyuf2_noota`), and the flashed layout are all the same single-4MB-app No-OTA table (nvs/otadata/uf2/ffat offsets unchanged). HIL re-verified on hardware: boots via boot_app0→ota_0, secrets persist, v2 check-in, display Add/Write + camera proof.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

tyeth (changes requested):
I expected the power pin to be tracked and then removed by the display controller / hardware during a display Remove message or destructor. If the pin is already enabled previously not by the display (pin in use) and the state (output) and logic level are correct then we can pass through okay but otherwise an error should be raised for power enable pin unavailable during display init etc

tyeth-ai-assisted:
Implemented the requested power-pin lifecycle:

- **Tracked + released on Remove/destructor**: when the pin is free at Add time, the display driver takes ownership — drives the rail to its enable level (honouring `is_inverted`) before init, and releases it (disable level) in its destructor, which runs on display **Remove/replace** (`Handle_Display_Remove` → delete `DisplayHardware` → delete driver).
- **Pin-in-use handling at init**: `beginI8080` first asks the digitalio controller (new read-only `DigitalIOController::QueryPinState()` — `GetPin` stays private). If the pin is already registered: **pass through** only when it's an `OUTPUT` already driving the enable level (the display never touches or releases a pin it doesn't own); otherwise the Add **fails** with `Power-enable pin unavailable — already in use with an incompatible direction or level`.

HIL-verified on the bench (flash → v2 check-in → 127-byte Add with the digitalio-Add power → serial `Backlight ON` / `i8080 ST7789 initialized successfully` / `Display added successfully`). Camera proof pending the bench camera being physically reseated (blind since Jul 05 — sensor-connector timeout).

Known limits (discussed with tyeth): this is deliberately the two-party (display↔digitalio) arbitration from the review — shared rails across display+I²C+NeoPixel, digitalio-Adds that arrive *after* the display owns the pin, and deep-sleep `gpio_hold` for inverted rails need a central refcounted rail registry, which is being scoped as a follow-up rather than grown inside this board PR.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

tyeth-ai-assisted:
Follow-up filed for the shared-rail / refcount ownership + sleep-mode rail handling discussed in review: #952. This PR stays at the two-party display↔digitalio arbitration.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

tyeth:
needs a retest visually confirming the refactor after the power enable pin coming in as a displayAdd.power (ditigalioAdd)

tyeth-ai-assisted:
### ✅ Visual retest — power enable via `displayAdd.power` (digitalio Add)

Fresh end-to-end HIL run at PR HEAD (job `338c6af8`, bench camera repaired + recalibrated): flash → v2 check-in → display Add carrying the panel power rail as a **`ws.digitalio.Add`** (`power`, field 12) → display Write → camera capture.

```
CHECKIN_VERDICT ok=true uid=lilygo-t-display-s321621634 proto=v2
INJECT_VERDICT published=true bytes=127 kind=display_add_i8080 resp={'status': 'OK'}
INJECT_VERDICT published=true bytes=90 kind=raw (display Write) resp={'status': 'OK'}
DISPLAY_CAPTURE_VERDICT exposure_us=32000 gain=2.0 focus=1.742 wb=yes cropped=yes
```

Serial:
```
[display] Adding display: tft-19-i8080
[display] Backlight ON
[display] i8080 ST7789 initialized successfully!
[display] Display added successfully: tft-19-i8080
```

The panel is lit **because** the display driver took ownership of the power rail from the proto (no board macros, no pre-existing digitalio pin — the free-pin ownership path from the lifecycle refactor):

LilyGo T-Display-S3 — power via digitalio Add, visual proof

Capture-side notes: the retest surfaced two camera-server bugs (manual exposure raced the still pipeline; AE left disabled after a manual still) and one `capture_display` weakness (scene-wide AF converging off-panel) — all fixed on the controller side (`e687b35`, `0769292`, `9bf01bf`: exposure baked into the still config, AE handed back, ROI-window autofocus + converged-dioptre lock). ROI/exposure recalibrated for the moved board.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

.github/workflows/build-migrate-api-v2.yml
```diff
@@ -106,6 +106,95 @@ jobs:
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        arduino-platform:
+          [
+            "lilygo_t_display_s3",
+          ]
+    steps:
+      - uses: actions/setup-python@v5
```
tyeth: Update CI-wippersnapper to be inline with 3.3.11 so no custom branch/build-step needed

tyeth (changes requested):
Timestamp proof showing latest test + firmware changes
image


@tyeth-ai-assisted Do not perform the CI-arduino change I have mentioned as a review comment, I will manually perform this task

.github/workflows/build-migrate-api-v2.yml
tyeth-ai-assisted: Not actioned per your review note — you're performing the ci-arduino 3.3.11 alignment manually. Once ci-wippersnapper carries the lilygo entry on 3.3.11, this job's dedicated `ci-wippersnapper-tdisplays3` ref (and the separate build job) can fold back into the standard matrix; happy to do that workflow-side cleanup when you say the branch is ready. Leaving this thread open as your marker.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

tyeth-ai-assisted: Done (88f2e044 + e8a5f599): with your ci-wippersnapper update carrying `lilygo_t_display_s3` on bsp-3-3-11, the board now builds in the standard `build-esp32sx` matrix — the dedicated `ci-wippersnapper-tdisplays3` ref and its separate build job are removed (the extra-library steps were already identical; Arduino_GFX resolves via library.properties depends).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

tyeth:
CI-arduino is updated to latest BSP simplifying the CI job here (now basic S3 matrix), and boards repo includes board in v2 branch.

src/components/display/drivers/dispDrvSt7789I8080.h
```diff
@@ -0,0 +1,338 @@
...
+ */
+#ifndef WS_DISP_DRV_ST7789_I8080_H
+#define WS_DISP_DRV_ST7789_I8080_H
+
+// Arduino_ESP32LCD8 (the i8080 8-bit data bus) requires the ESP32-S3 LCD_CAM
+// peripheral, so this driver only compiles for the ESP32-S3.
+#if defined(ESP32) && defined(CONFIG_IDF_TARGET_ESP32S3)
+
+#include "dispDrvBase.h"
+#include <Arduino_GFX_Library.h>
```
brentru, line 34: isnt this already included by driver base?

brentru (approved):

src/components/display/drivers/dispDrvSt7789I8080.h
tyeth: Not this moon-on-our-nation one no — dispDrvBase.h includes the Adafruit GFX family (via Adafruit_ThinkInk.h), but Arduino_ESP32LCD8/Arduino_ST7789 come from moononournation's separate Arduino_GFX library (Arduino_GFX_Library.h), which nothing else in the chain provides. Same convention as the existing dispDrvRgb666.h driver.

tyeth (approved):

PR #937, opened 2026-06-25
fix(fs): never auto-format on mount failure; gate format on a persistent NVS provisioned-marker

flavio-fernandes:
## What this does

This targets `tyeth/flash-health-format-guard` (the source branch of #931). On-hardware testing of #931 on a MagTag showed secrets.json **was still wiped** on a deep battery drain ([log + analysis](https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/931#issuecomment-4803426537)). Digging into *why* surfaced two issues; this PR fixes both and folds in @brentru's review feedback.

### 1. The flash-health guard in #931 was unreachable (the real bug)

`initFilesystem(force_format=false)` still called `f_mkfs()` on a mount failure:

```cpp
if (force_format || !wipperFatFs.begin(&flash)) {  // <- formats on mount failure
```

So the constructor's **first** `initFilesystem()` call reformatted the corrupted FAT and `createSecretsFile()` halted — all *before* `flashSafeToFormat()` was ever consulted. The guard was effectively dead code, which is why the wipe still happened (`Please edit the secrets.json file`, not the guard's `Flash reads unstable` message).

**Fix:** `initFilesystem()` now returns `false` on a mount failure instead of creating a filesystem. Only the constructor's guarded path may format.

### 2. "Flash is readable" can't tell blank-from-factory apart from brownout-corrupted

Once the battery recovers enough to boot, the partition table and NVS read back fine — so a health probe alone always says "safe to format" on a recovered board. The shallow sags survived; the deep drain (corrupt the FAT, recover, boot) sailed straight through.

**Fix (primary gate):** a persistent `fs_ok` marker in **NVS**, which survives both a brownout and a FAT format. Set once after a successful real mount. On a later mount failure:

- **marker present** → the FAT was corrupted (brownout), never blank → **refuse to format**, halt, conserve battery → secrets.json preserved
- **no marker** → genuine first boot → fall through to the health/blank check, then format

### @brentru's #931 review feedback, folded in

- `flashSafeToFormat()` → `isFlashSafeToFormat()`
- locate NVS via `esp_partition_find_first()` instead of hardcoded `0x8000`/`0x9000`
- inspect the **FAT volume's boot sector** on ESP32 too (not just NVS) — only format if it reads back genuinely erased
- re-validate the flash **JEDEC ID after** the destructive format
- moved the ESP32 includes into `Wippersnapper_FS.h` under an `ARDUINO_ARCH_ESP32` guard

## Behavior

| Scenario | Result |
|---|---|
| Provisioned board, brownout corrupts FAT | marker present → **refuse to format**, halt → secrets safe |
| Genuine first boot (no marker, blank flash) | format ✅ |
| Normal boot, FAT mounts | mounts, marks provisioned, runs |
| Non-ESP32 (no NVS) | falls back to the blank-volume check, as in #931 |

Deliberate tradeoff: a provisioned board that can't mount now **halts asking for a recharge/reset instead of wiping**. Failing safe (preserve data) beats failing destructive. Normal re-provisioning is unaffected — editing `secrets.json` over USB never needs a format; only auto-format-on-mount-failure is blocked. A genuinely (non-brownout) corrupted FS would need an explicit reformat gesture / NVS erase, which could be a nice follow-up.

## Testing

- Builds clean for `adafruit_magtag29_esp32s2` (`pio run -e adafruit_magtag29_esp32s2` → SUCCESS).
- @flavio-fernandes will run the same sacrificed-USB battery-drain reproduction against this build and report back on #931.

Notes for reviewers: the version bump to `1.0.0-pr931.3` is just a test-build marker so the serial log identifies the flashed build — drop it before any merge. NVS namespace `ws_fs` / key `fs_ok`; if a board's partition table has no NVS, `Preferences.begin()` returns false and we degrade safely to the health/blank check.

### DIFF
--- a/.github/workflows/native-tests.yml
+++ b/.github/workflows/native-tests.yml
@@ -1,0 +1,32 @@
+name: Native unit tests
+
+# Host-compiled, hardware-free unit tests for pure decision logic (e.g. the
+# filesystem (re)format policy). These exercise behaviour that the board build
+# matrix cannot: a compile proves the firmware links, not that the gating logic
+# is correct. Runs in seconds with no toolchain download.
+
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  native-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Build & run native tests
+        run: |
+          set -euo pipefail
+          shopt -s nullglob
+          tests=(test/native/*.cpp)
+          if [ ${#tests[@]} -eq 0 ]; then
+            echo "No native tests found under test/native/"; exit 1
+          fi
+          for t in "${tests[@]}"; do
+            echo "==== $t ===="
+            c++ -std=c++17 -Wall -Wextra -I src/provisioning/tinyusb \
+              "$t" -o /tmp/native_test
+            /tmp/native_test
+          done
--- a/library.properties
+++ b/library.properties
@@ -1,5 +1,5 @@
 name=Adafruit WipperSnapper
-version=1.0.0-pr931.4
+version=1.0.0-pr931.5
 author=Adafruit
 maintainer=Adafruit <adafruitio@adafruit.com>
 sentence=Arduino application for Adafruit.io WipperSnapper
--- a/library.properties
+++ b/library.properties
@@ -1,5 +1,5 @@
 name=Adafruit WipperSnapper
-version=1.0.0-beta.130
+version=1.0.0-pr931.3
 author=Adafruit
 maintainer=Adafruit <adafruitio@adafruit.com>
 sentence=Arduino application for Adafruit.io WipperSnapper
--- a/library.properties
+++ b/library.properties
@@ -1,5 +1,5 @@
 name=Adafruit WipperSnapper
-version=1.0.0-pr931.3
+version=1.0.0-pr931.4
 author=Adafruit
 maintainer=Adafruit <adafruitio@adafruit.com>
 sentence=Arduino application for Adafruit.io WipperSnapper
--- a/src/Wippersnapper.cpp
+++ b/src/Wippersnapper.cpp
@@ -2703,64 +2703,62 @@ void Wippersnapper::publish(const char *topic, uint8_t *payload, uint16_t bLen,
 #ifdef ARDUINO_ARCH_ESP32
 /**************************************************************/
 /*!
-    @brief    Prints string reset reason of ESP32
+    @brief    Maps an ESP32 reset-reason code to a human-readable string.
     @param    reason
-              The return code of rtc_get_reset_reason(coreNum)
+              The return code of rtc_get_reset_reason(coreNum).
+    @returns  A static, human-readable name for the reset reason (e.g.
+              "SW_CPU_RESET"), or "NO_MEAN" for an unrecognized code.
+*/
+/**************************************************************/
+// https://github.com/espressif/arduino-esp32/blob/master/libraries/ESP32/examples/ResetReason/ResetReason.ino
+const char *getResetReasonStr(int reason) {
+  switch (reason) {
+  case 1:
+    return "POWERON_RESET"; /**<1,  Vbat power on reset*/
+  case 3:
+    return "SW_RESET"; /**<3,  Software reset digital core*/
+  case 4:
+    return "OWDT_RESET"; /**<4,  Legacy watch dog reset digital core*/
+  case 5:
+    return "DEEPSLEEP_RESET"; /**<5,  Deep Sleep reset digital core*/
+  case 6:
+    return "SDIO_RESET"; /**<6,  Reset by SLC module, reset digital core*/
+  case 7:
+    return "TG0WDT_SYS_RESET"; /**<7,  Timer Group0 Watch dog reset*/
+  case 8:
+    return "TG1WDT_SYS_RESET"; /**<8,  Timer Group1 Watch dog reset*/
+  case 9:
+    return "RTCWDT_SYS_RESET"; /**<9,  RTC Watch dog Reset digital core*/
+  case 10:
+    return "INTRUSION_RESET"; /**<10, Instrusion tested to reset CPU*/
+  case 11:
+    return "TGWDT_CPU_RESET"; /**<11, Time Group reset CPU*/
+  case 12:
+    return "SW_CPU_RESET"; /**<12, Software reset CPU*/
+  case 13:
+    return "RTCWDT_CPU_RESET"; /**<13, RTC Watch dog Reset CPU*/
+  case 14:
+    return "EXT_CPU_RESET"; /**<14, for APP CPU, reset by PRO CPU*/
+  case 15:
+    return "RTCWDT_BROWN_OUT_RESET"; /**<15, Reset when vdd voltage unstable*/
+  case 16:
+    return "RTCWDT_RTC_RESET"; /**<16, RTC Watch dog reset core and rtc module*/
+  default:
+    return "NO_MEAN";
+  }
+}
+
+/**************************************************************/
+/*!
+    @brief    Prints the ESP32 reset reason and flags a brownout-caused reset.
+    @param    reason
+              The return code of rtc_get_reset_reason(coreNum).
 */
 /**************************************************************/
 void print_reset_reason(int reason) {
-  // //
-  // https://github.com/espressif/arduino-esp32/blob/master/libraries/ESP32/examples/ResetReason/ResetReason.ino
-  switch (reason) {
-  case 1:
-    WS_DEBUG_PRINTLN("POWERON_RESET");
-    break; /**<1,  Vbat power on reset*/
-  case 3:
-    WS_DEBUG_PRINTLN("SW_RESET");
-    break; /**<3,  Software reset digital core*/
-  case 4:
-    WS_DEBUG_PRINTLN("OWDT_RESET");
-    break; /**<4,  Legacy watch dog reset digital core*/
-  case 5:
-    WS_DEBUG_PRINTLN("DEEPSLEEP_RESET");
-    break; /**<5,  Deep Sleep reset digital core*/
-  case 6:
-    WS_DEBUG_PRINTLN("SDIO_RESET");
-    break; /**<6,  Reset by SLC module, reset digital core*/
-  case 7:
-    WS_DEBUG_PRINTLN("TG0WDT_SYS_RESET");
-    break; /**<7,  Timer Group0 Watch dog reset digital core*/
-  case 8:
-    WS_DEBUG_PRINTLN("TG1WDT_SYS_RESET");
-    break; /**<8,  Timer Group1 Watch dog reset digital core*/
-  case 9:
-    WS_DEBUG_PRINTLN("RTCWDT_SYS_RESET");
-    break; /**<9,  RTC Watch dog Reset digital core*/
-  case 10:
-    WS_DEBUG_PRINTLN("INTRUSION_RESET");
-    break; /**<10, Instrusion tested to reset CPU*/
-  case 11:
-    WS_DEBUG_PRINTLN("TGWDT_CPU_RESET");
-    break; /**<11, Time Group reset CPU*/
-  case 12:
-    WS_DEBUG_PRINTLN("SW_CPU_RESET");
-    break; /**<12, Software reset CPU*/
-  case 13:
-    WS_DEBUG_PRINTLN("RTCWDT_CPU_RESET");
-    break; /**<13, RTC Watch dog Reset CPU*/
-  case 14:
-    WS_DEBUG_PRINTLN("EXT_CPU_RESET");
-    break; /**<14, for APP CPU, reset by PRO CPU*/
-  case 15:
-    WS_DEBUG_PRINTLN("RTCWDT_BROWN_OUT_RESET");
+  WS_DEBUG_PRINTLNVAR(getResetReasonStr(reason));
+  if (reason == 15) // RTCWDT_BROWN_OUT_RESET
     WS.brownOutCausedReset = true;
-    break; /**<15, Reset when the vdd voltage is not stable*/
-  case 16:
-    WS_DEBUG_PRINTLN("RTCWDT_RTC_RESET");
-    break; /**<16, RTC Watch dog reset digital core and rtc module*/
-  default:
-    WS_DEBUG_PRINTLN("NO_MEAN");
-  }
 }
 
 #if CONFIG_IDF_TARGET_ESP32 // ESP32/PICO-D4
@@ -2790,6 +2788,20 @@ void print_reset_reason(int reason) {
 /**************************************************************************/
 void get_and_print_reset_reason_for_cpu(int cpuCore) {
   print_reset_reason(rtc_get_reset_reason(cpuCore));
+}
+
+/**************************************************************************/
+/*!
+    @brief    Returns the raw reset-reason code for the given CPU core, with no
+              side effects (does not touch brownOutCausedReset or print). Used
+              by the filesystem layer to report/persist the actual boot reason.
+    @param    cpuCore
+              The core number to read the reset reason for.
+    @returns  The rtc_get_reset_reason(cpuCore) code.
+*/
+/**************************************************************************/
+int getResetReasonCode(int cpuCore) {
+  return (int)rtc_get_reset_reason(cpuCore);
 }
 
 // end of ARDUINO_ARCH_ESP32
--- a/src/Wippersnapper.h
+++ b/src/Wippersnapper.h
@@ -167,7 +167,7 @@
 #endif
 
 #define WS_VERSION                                                             \
-  "1.0.0-beta.130" ///< WipperSnapper app. version (semver-formatted)
+  "1.0.0-pr931.3" ///< WipperSnapper app. version (semver-formatted)
 
 // Reserved Adafruit IO MQTT topics
 #define TOPIC_IO_THROTTLE "/throttle" ///< Adafruit IO Throttle MQTT Topic
--- a/src/Wippersnapper.h
+++ b/src/Wippersnapper.h
@@ -167,7 +167,16 @@
 #endif
 
 #define WS_VERSION                                                             \
-  "1.0.0-pr931.4" ///< WipperSnapper app. version (semver-formatted)
+  "1.0.0-pr931.5" ///< WipperSnapper app. version (semver-formatted)
+
+#ifdef ARDUINO_ARCH_ESP32
+// Reset-reason helpers (defined in Wippersnapper.cpp). Exposed so the
+// filesystem layer can surface the actual boot/reset reason in its protective
+// halt and persist it to NVS for diagnosis, instead of only inferring "likely a
+// brownout".
+const char *getResetReasonStr(int reason); ///< Reset-reason code -> string
+int getResetReasonCode(int cpuCore = 0);   ///< rtc_get_reset_reason(cpuCore)
+#endif
 
 // Reserved Adafruit IO MQTT topics
 #define TOPIC_IO_THROTTLE "/throttle" ///< Adafruit IO Throttle MQTT Topic
--- a/src/Wippersnapper.h
+++ b/src/Wippersnapper.h
@@ -167,7 +167,7 @@
 #endif
 
 #define WS_VERSION                                                             \
-  "1.0.0-pr931.3" ///< WipperSnapper app. version (semver-formatted)
+  "1.0.0-pr931.4" ///< WipperSnapper app. version (semver-formatted)
 
 // Reserved Adafruit IO MQTT topics
 #define TOPIC_IO_THROTTLE "/throttle" ///< Adafruit IO Throttle MQTT Topic
--- a/src/provisioning/tinyusb/Wippersnapper_FS.cpp
+++ b/src/provisioning/tinyusb/Wippersnapper_FS.cpp
@@ -95,6 +95,13 @@ bool setVolumeLabel() {
 // fact reliably outlives both a brownout and a FAT format.
 static const char *kWSFsNvsNamespace = "ws_fs";
 static const char *kWSFsProvisionedKey = "fs_ok";
+// Boot-reason diagnostics, persisted in the same NVS namespace (survives a
+// brownout and a FAT format). last_rst is updated every boot; fmt_rst/fmt_cnt
+// capture the reset reason in effect when a (re)format actually ran, so a
+// format that wiped data can be correlated with the reset that preceded it.
+static const char *kWSFsLastResetKey = "last_rst"; ///< reset code this boot
+static const char *kWSFsFormatResetKey = "fmt_rst"; ///< reset code at last format
+static const char *kWSFsFormatCountKey = "fmt_cnt"; ///< number of (re)formats
 #endif
 
 /**************************************************************************/
@@ -138,6 +145,79 @@ static void markFilesystemProvisioned() {
   if (!prefs.getBool(kWSFsProvisionedKey, false))
     prefs.putBool(kWSFsProvisionedKey, true);
   prefs.end();
+#endif
+}
+
+/**************************************************************************/
+/*!
+    @brief    Records the reset reason seen on this boot into NVS (last_rst),
+              so the actual boot reason is available for diagnosis even when the
+              board halts before the usual printDeviceInfo() reset-reason dump.
+              No-op on platforms without NVS / reset-reason support.
+*/
+/**************************************************************************/
+static void recordBootResetReason() {
+#ifdef ARDUINO_ARCH_ESP32
+  Preferences prefs;
+  if (!prefs.begin(kWSFsNvsNamespace, /*readOnly=*/false))
+    return;
+  prefs.putUChar(kWSFsLastResetKey, (uint8_t)getResetReasonCode(0));
+  prefs.end();
+#endif
+}
+
+/**************************************************************************/
+/*!
+    @brief    Records, into NVS, that a (re)format happened this boot and the
+              reset reason in effect when it ran (fmt_rst), plus a running count
+              (fmt_cnt). This is the "write the boot reason only after a format"
+              diagnostic: a format that erased data can later be correlated with
+              the reset that preceded it. No-op without NVS.
+*/
+/**************************************************************************/
+static void recordFormatEvent() {
+#ifdef ARDUINO_ARCH_ESP32
+  Preferences prefs;
+  if (!prefs.begin(kWSFsNvsNamespace, /*readOnly=*/false))
+    return;
+  prefs.putUChar(kWSFsFormatResetKey, (uint8_t)getResetReasonCode(0));
+  prefs.putUInt(kWSFsFormatCountKey, prefs.getUInt(kWSFsFormatCountKey, 0) + 1);
+  prefs.end();
+#endif
+}
+
+/**************************************************************************/
+/*!
+    @brief    Prints the persisted boot/format diagnostics (this boot's reset
+              reason, and - if any format has ever run - the count and the reset
+              reason at the last format) to the serial console. No-op without
+              NVS.
+*/
+/**************************************************************************/
+static void printFsDiagnostics() {
+#ifdef ARDUINO_ARCH_ESP32
+  Preferences prefs;
+  if (!prefs.begin(kWSFsNvsNamespace, /*readOnly=*/true))
+    return;
+  uint8_t lastRst = prefs.getUChar(kWSFsLastResetKey, 0);
+  uint8_t fmtRst = prefs.getUChar(kWSFsFormatResetKey, 0);
+  uint32_t fmtCnt = prefs.getUInt(kWSFsFormatCountKey, 0);
+  prefs.end();
+
+  WS_DEBUG_PRINT("[fs] boot reset reason: ");
+  WS_DEBUG_PRINTVAR((int)lastRst);
+  WS_DEBUG_PRINT(" (");
+  WS_DEBUG_PRINTVAR(getResetReasonStr(lastRst));
+  WS_DEBUG_PRINTLN(")");
+  if (fmtCnt > 0) {
+    WS_DEBUG_PRINT("[fs] filesystem (re)formats so far: ");
+    WS_DEBUG_PRINTLNVAR(fmtCnt);
+    WS_DEBUG_PRINT("[fs] reset reason at last format: ");
+    WS_DEBUG_PRINTVAR((int)fmtRst);
+    WS_DEBUG_PRINT(" (");
+    WS_DEBUG_PRINTVAR(getResetReasonStr(fmtRst));
+    WS_DEBUG_PRINTLN(")");
+  }
 #endif
 }
 
@@ -227,6 +307,11 @@ Wippersnapper_FS::Wippersnapper_FS() {
   // Wait for detach
   delay(500);
 
+  // Persist this boot's reset reason early, before any mount/format decision,
+  // so the actual boot reason is recoverable for diagnosis even if we halt here
+  // (the usual printDeviceInfo() reset-reason dump runs later, in connect()).
+  recordBootResetReason();
+
   // Attempt to mount the existing filesystem WITHOUT formatting. initFilesystem
   // (with force_format=false) no longer creates a filesystem on a mount
   // failure - it returns false and lets the gates below decide - so a
@@ -287,12 +372,19 @@ Wippersnapper_FS::Wippersnapper_FS() {
                    "Recharge/repower the board and reset.");
           }
         }
+        // Persist that a (re)format ran this boot and the reset reason behind
+        // it, so a format that wiped data can be correlated with its cause.
+        recordFormatEvent();
         break;
       case WsFsAction::Mounted:
         break; // unreachable: mount already failed above
       }
     }
   }
+
+  // Report the persisted boot/format diagnostics (this boot's reset reason,
+  // and any prior format + the reset reason that preceded it).
+  printFsDiagnostics();
 
   // Initialize USB-MSD
   initUSBMSC();
@@ -743,9 +835,23 @@ void Wippersnapper_FS::writeToBootOut(PGM_P str) {
 /**************************************************************************/
 void Wippersnapper_FS::fsHalt(String msg) {
   statusLEDSolid(WS_LED_STATUS_FS_WRITE);
+#ifdef ARDUINO_ARCH_ESP32
+  // Capture the actual reset reason once, so the halt reports the observed boot
+  // reason (e.g. "12 (SW_CPU_RESET)") rather than only inferring "likely a
+  // brownout" - a brownout is frequently not reported as a brownout reset.
+  int rstCode = getResetReasonCode(0);
+  const char *rstStr = getResetReasonStr(rstCode);
+#endif
   while (1) {
     WS_DEBUG_PRINTLN("Fatal Error: Halted execution!");
     WS_DEBUG_PRINTLN(msg.c_str());
+#ifdef ARDUINO_ARCH_ESP32
+    WS_DEBUG_PRINT("Boot/reset reason: ");
+    WS_DEBUG_PRINTVAR(rstCode);
+    WS_DEBUG_PRINT(" (");
+    WS_DEBUG_PRINTVAR(rstStr);
+    WS_DEBUG_PRINTLN(")");
+#endif
     delay(1000);
     yield();
   }
--- a/src/provisioning/tinyusb/Wippersnapper_FS.cpp
+++ b/src/provisioning/tinyusb/Wippersnapper_FS.cpp
@@ -485,7 +485,32 @@ bool Wippersnapper_FS::initFilesystem(bool force_format) {
 
   // Check if secrets.json file already exists
   if (!configFileExists()) {
-    // Create new secrets.json file and halt
+    // configFileExists() returns false when secrets.json is absent OR when its
+    // first byte reads back 0x00/0xFF. That blank first byte is the same
+    // failed-flash-read a brownout produces (the SPI rail sags during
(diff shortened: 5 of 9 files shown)


flavio-fernandes:
Added a hardware-free regression test (`217b00db`), since the board build matrix only proves the firmware compiles — it can't catch a logic regression like the one this branch fixes (the old flash-health guard was unreachable dead code, and it compiled fine).

- Extracted the (re)format decision into a pure `decideFsAction()` in `Wippersnapper_FS_format_policy.h` — no Arduino/flash/NVS includes — and the constructor now calls it (also makes that gate easier to read).
- `test/native/test_fs_format_policy.cpp` walks the full truth table on a host compiler, including the exact case that wiped secrets.json: **previously provisioned + unmountable FAT + healthy-looking flash → refuse to format**.
- `.github/workflows/native-tests.yml` compiles and runs every `test/native/*.cpp` on push/PR (seconds, no toolchain download).

Run it locally:
```
c++ -std=c++17 -I src/provisioning/tinyusb \
  test/native/test_fs_format_policy.cpp -o /tmp/fs_policy_test && /tmp/fs_policy_test
```
```
decideFsAction() truth table
  ok   : mounted (fresh board)
  ok   : mounted (provisioned board)
  ok   : provisioned + unmountable + flash looks healthy -> refuse to format
  ok   : provisioned + unmountable + flash unhealthy -> refuse to format
  ok   : first boot + blank healthy flash -> format
  ok   : first boot + unhealthy/not-blank flash -> refuse to format
PASS: all cases
```

Easy to extend: drop another `test/native/*.cpp` and CI picks it up. Entirely separable from the fix if you'd rather take the firmware change alone — happy to split it out.

flavio-fernandes:
## Status update: on-hardware results + an open design question for you both

@tyeth @brentru — quick summary of where this PR stands after a round of real MagTag testing, plus one design call I'd like your input on.

### What's in the PR now

1. **`initFilesystem()` no longer auto-formats on a mount failure.** This was the actual reason #931's guard didn't protect the board: `initFilesystem(force_format=false)` still ran `f_mkfs()` on a failed mount, so the constructor's first call reformatted the corrupted FAT *before* the flash-health guard was ever reached. It was dead code. It now returns `false` and lets the constructor's guarded path decide.
2. **NVS "provisioned" marker as the primary gate.** Once a working FS has mounted, a `fs_ok` flag is set in NVS (survives both a brownout and a FAT format). On a later mount failure: marker present → it's corruption, **refuse to format**; no marker → genuine first boot → fall through to the health/blank check, then format.
3. **brentru's review notes, folded in:** `isFlashSafeToFormat()` rename, NVS located via `esp_partition_find_first()` (no hardcoded `0x8000`/`0x9000`), FAT boot-sector inspected on ESP32 too, post-format JEDEC re-check, ESP32 includes moved to the header.
4. **Host unit test + CI** (`test/native/` + `.github/workflows/native-tests.yml`): the format decision is extracted into a pure `decideFsAction()` and the truth table is unit-tested on a host compiler — the build matrix can't catch a logic regression like the dead-code guard above.
5. **Protective halt now surfaces its reason over USB serial** (added after the test below).

Builds clean for `adafruit_magtag29_esp32s2`; host test passes. Current test build is `1.0.0-pr931.4`.

### On-hardware results (MagTag, sacrificed no-VBUS cable, battery-only)

- **Round 1:** survived many shallow sag/recover cycles incl. a 27-min dead stretch — secrets intact every boot, no wipe. But every mount *succeeded*, so this only proved "no spurious wipe," not that the refusal path fires.
- **Round 2 — the real one:** a deeper drain genuinely corrupted the FAT. On the next boot the device found the FAT unmountable, saw the NVS marker, and **refused to reformat** — halting instead of wiping. Confirmed by the new serial message:
  ```
  Filesystem unreadable but this board was previously provisioned - likely a
  brownout. Recharge and reset; refusing to reformat to protect your secrets.json.
  ```
  On beta.130 / pr931.2 this exact moment is where `secrets.json` got wiped. Here it was preserved on flash (the partition was never reformatted).

**Verified:** the device did not reformat (halted with the protective message, not "Please edit the secrets.json file"). **Pending:** a flash-dump byte-check that the `secrets.json` contents themselves are intact (vs. damaged by the same brownout) — I'll post that once I've run it.

### The open design question

Round 2 also exposed a gap: when a FAT corrupts **permanently** (won't remount even on stable power), the fix protects the data but the board is stuck needing esptool to recover. Fine at a desk, rough for a board in the field.

Proposed direction (would like your steer before I build it):
- **Expose the drive read-only in the protective halt** (`usb_msc.setWritable(false)`): the host can read/copy `secrets.json` off for rescue, but cannot reformat/wipe it even if macOS prompts to "initialize." Solves diagnosis + data rescue with zero wipe risk.
- Optionally pair with a **button-hold-at-boot = deliberate reformat** escape hatch for users who just want a fresh start.

Also flagging an earlier note of yours, @tyeth — the deep-sleep-and-retry idea for the slow-solar-recovery case is complementary to all of this and probably still worth doing.

Happy to keep iterating and re-testing on hardware. What's your read on the marker-as-primary-gate approach, and on the read-only-rescue halt?

flavio-fernandes:
### Supporting logs (MagTag, SSIDs redacted)

Two exhibits backing the summary above.

**Exhibit A — secrets survived aggressive cycling (`pr931.3`).** Many disconnect/reconnect cycles including a ~27-minute dead stretch; every boot that ran came back with the **full original config** and never once printed `Please edit the secrets.json file` or `Invalid IO credentials`:

```
---------- 2026-06-25 16:37:02 -0400: Disconnected ----------     ← deep drain begins
---------- 2026-06-25 17:04:25 -0400: Connected ----------        (~27 min dead)
Connected to WiFi!
Registering hardware with WipperSnapper...
GOT Registration Response Message:
        GPIO Pins: 18
        Analog Pins: 5
        Reference voltage: 2.60v
Completed registration process, configuration next!
...
Configured digital input pin on D15 / D14 / D12 / D11 / D0
Configured digital output pin on D21 / D16
Initial Pin Configuration Complete!
Hardware configured successfully!
Registration and configuration complete!
                                        Running application...
[display] Writing message to display:  Flavio Testing  /  Magtag Adafruit
```

The device resumed its real configuration — secrets intact — after a 27-minute outage. (Every mount here *succeeded*, so this proves "no spurious wipe under cycling," not the refusal path. That's Exhibit B.)

**Exhibit B — the refuse-to-format gate firing on a genuinely corrupted FAT (`pr931.4`).** A deeper drain corrupted the FAT; on the next boot the device refused to reformat and halted, now visible over serial (the no-USB black box from before is fixed). The status NeoPixels were solid yellow and no `WIPPER` drive mounted:

```
---------- 2026-06-25 22:25:20 -0400: Logging Started ----------
Filesystem unreadable but this board was previously provisioned - likely a
brownout. Recharge and reset; refusing to reformat to protect your secrets.json.
Filesystem unreadable but this board was previously provisioned - likely a
brownout. Recharge and reset; refusing to reformat to protect your secrets.json.
   ... (repeats once/second) ...
```

For contrast, the same hardware on `beta.130` / `pr931.2` reached this point and printed `ERROR: Please edit the secrets.json file` — i.e. it had already run `f_mkfs()` and wiped `secrets.json` (full log earlier on #931). The fix turns that destructive moment into a non-destructive, recoverable halt.

I had to step away before dumping the `ffat` partition to byte-verify the `secrets.json` contents survived intact — I'll post that read-back when I'm back at the bench. Happy to capture any other specific serial output you'd find useful.

tyeth:
Interesting, thanks for this.
It's a heatwave this week so I'll take a proper look next week instead.

brentru:
> initFilesystem() no longer auto-formats on a mount failure

Interesting. The FAT filesystem corruption after a full drain is concerning. I am going to take a look at beta.130 / pr931.2 and diff it with this PR, maybe we're doing something differently there.

brentru:
> Round 2 — the real one: a deeper drain genuinely corrupted the FAT

@flavio-fernandes  This responds with "likely a brownout" but are you able to see the boot reason?

tyeth:
There should be no corruption, merely the misinterpretation of a failed
flash operation as a corrupted state / file / partition.
 Apart from avoiding the format if unnecessary hypothetically, the previous
PR also verified the existing file content compared to the planned new file
content before writing.
 Looking at the battery levels from before it's at about 2.87 the
predefined ESP brownout voltage, but the boot reason will vary between
those brownout reboots (Not always being brown out).
  I should have stated more clearly that the most recent PR of mine that was agent generated was not verified as to its
utility or correctness, merely a suggestion at a possible verification
method implemented by an ineffective LLM.
  The most promising option is probably to write out the boot reason to a
file called boot reason (or nvm) only after a format has occurred on that
boot cycle. Then we can more easily diagnose the potential fault states.
  I'll sling something together next week if the two of you don't tackle it
first

flavio-fernandes:
Thank you both. I pulled the flash off the halted board, and it explains the whole sequence — including, finally, *what made it unbootable*. (Private values redacted; it's my own board.)

**The damage is a single 4 KB sector — the FAT boot sector — erased to all-`0xFF`. Everything else is intact:** the partition table reads perfectly, the FAT tables and root directory are untouched (the root dir still lists `secrets.json`, 207 B), and the file data is all there. The volume won't mount *only* because the boot sector's `0x55AA`/BPB is blank.

**So what wrote the `0xFF`? Nothing *wrote* it — a flash *erase* blanked it.** A clean 4 KB-aligned block of `0xFF` is the erased state of flash; it can't come from bit-rot or a bad read. And the only thing that erases that particular sector is the format step, `f_mkfs()` (normal file writes never touch the boot-sector page). Since the filesystem underneath is still the original one with `secrets.json` in it, the format had *only just started*. The chain:

> dying battery → mount fails → **old auto-format-on-mount-failure calls `f_mkfs()`** → brownout resets the chip mid-format, right after the boot-sector erase, before any rewrite → boot sector left blank, the rest of the FS never touched.

That's why a full recharge doesn't fix it (the sector is genuinely erased, not a transient read), and it's the *same* `f_mkfs()` that would have wiped `secrets.json` had it run to completion. **This PR removes exactly that call** — a provisioned board never starts the format, so it never reaches this state. pr931.4 confirms it: it can't mount, sees the NVS marker, and halts instead of formatting (the loop the board was stuck in). The erase here predates pr931.4 — it can't format a provisioned board, and the on-flash boot-out history still shows older `beta.13x` boots.

**@brentru — byte-verify:** `secrets.json` survived 100% intact; I recovered it as valid JSON straight off the "dead" board.

**@tyeth** — this is exactly your "failed flash op misread as corruption": no corruption, just an interrupted erase. The one refinement is that it needs a recovery gesture, since it won't self-heal on recharge.

**@brentru on the beta.130 diff:** the only behavioral change is `initFilesystem()` no longer running that `f_mkfs()` on a failed mount. Nothing new corrupts anything.

To close the loop I recovered and resuscitated this board from the dump (carved out `secrets.json`, let the firmware rebuild a clean FAT, dropped the file back) — it's running again on its original secrets. I'll also add the boot-reason instrumentation (`rtc_get_reset_reason()` in the halt path + persist to NVS on format, per @tyeth) so future cases report the reset code instead of guessing.

Redacted forensic summary + exact commands in the next comment. No rush — stay cool. 🌡️

flavio-fernandes:
### Flash forensics + recovery commands

Full 4 MB dump taken read-only; all private values (IO creds, SSID, MAC) redacted.

**What the dump shows**

```text
[1] CHIP IS HEALTHY. Partition table @0x8000 reads perfectly; all 4 MB read back
    clean on stable USB power. => a "flash health" probe ALWAYS passes on a
    recovered board, so health alone can't gate format (validates the NVS-marker gate).

[2] EXACTLY ONE 4 KB SECTOR IS DAMAGED — the FAT boot sector.
      ffat partition @ 0x310000 (960 KB):
        sector 0  (0x310000–0x310FFF) ......... ERASED to 0xFF   <-- only damage
        FAT tables + root dir (next sectors) .. INTACT
        file data ............................. INTACT
    Boot-sector signature 0x55AA reads 0xFFFF -> mount fails. Nothing else wrong.

[3] ROOT DIRECTORY STILL REFERENCES THE FILES:
        SECR~138.JSO  -> secrets.json        size=207   cluster=11
        WIPPER~1.TXT  -> wipper_boot_out.txt size=19296 cluster=14
        FSEVEN~1      -> .fseventsd/

[4] secrets.json SURVIVED BYTE-FOR-BYTE — recovered as valid JSON:
        { "io_username": "****", "io_key": "aio_****",
          "network_type_wifi": { "network_ssid": "****",
                                 "network_password": "****" },
          "status_pixel_brightness": 0.2 }

[5] NVS "provisioned" marker present (ws_fs / fs_ok) -> the guard fired correctly.

[6] On-flash boot-out history still shows older beta.13x boots -> the erase
    predates this firmware (an earlier auto-formatting build).
```

An all-`0xFF`, 4 KB-aligned boot sector over an otherwise-pristine filesystem is the
fingerprint of an **interrupted `f_mkfs()`** — a flash erase that a brownout cut off
before the rewrite — not bit-rot and not a chip fault.

**How the dump was taken**

```bash
# Board into ROM download mode: hold BOOT, tap RESET, release BOOT
esptool --chip esp32s2 --port <port> --before no_reset --after no_reset \
        read_flash 0x0 0x400000 magtag_fullflash.bin      # full 4 MB, read-only
```

**Resuscitation — recover secrets + rebuild FS, no data loss**

```bash
# 1) Back up everything first (read-only)
esptool --chip esp32s2 --port <port> --before no_reset --after no_reset \
        read_flash 0x0 0x400000 magtag_fullflash.bin

# 2) Carve secrets.json out of the ffat region (data was intact)
strings magtag_fullflash.bin | grep -A6 io_username      # save as recovered_secrets.json

# 3) Clear NVS marker + the damaged ffat so firmware does a clean first-boot format
esptool --chip esp32s2 --port <port> --before no_reset --after no_reset erase_region 0x9000   0x5000    # nvs (fs_ok)
esptool --chip esp32s2 --port <port> --before no_reset --after no_reset erase_region 0x310000 0xF0000   # ffat

# 4) Press RESET. Resident firmware finds a clean slate -> builds a fresh FAT ->
#    WIPPER mounts (and halts asking to edit secrets.json — expected first boot).

# 5) Restore the recovered file onto the drive
cp recovered_secrets.json /Volumes/WIPPER/secrets.json
sync && diskutil eject /Volumes/WIPPER

# 6) Press RESET -> board boots with the original secrets and reconnects to AIO. Alive.
```

Because only the boot sector was gone, the *data* was always recoverable. A future
firmware could automate steps 4–5 as a **read-only MSC rescue** (`usb_msc.setWritable(false)`)
so a user just drags their `secrets.json` off the drive — no esptool needed.

flavio-fernandes:
Hi @tyeth and @brentru — fresh bench results, and I think they finally pin down what's going on. I ran a full battery-drain cycle on the instrumented build (pr931.5, which prints the real reset reason in the halt path), dumped the flash to see exactly what survived, then recovered the board. Here's the whole thing in order.

## 1. Battery good → board running fine

The MagTag was up and healthy on battery: WiFi scan, `Connected to WiFi!`, `Connecting to AIO MQTT`, `WiFi Status: 20`. This was a freshly-formatted filesystem with a valid `secrets.json` — a known-good starting point.

## 2. Battery depleted → board powers off

The serial log shows a clean `Disconnected` the moment the battery gave out. Full power loss, not a flicker.

## 3. Power returns → protective halt, and now we can see the reset reason

On the next boot the board can't mount the filesystem and enters the guard's halt loop. The new instrumentation prints the **actual** reset reason instead of guessing:

```
Filesystem unreadable but this board was previously provisioned - likely a brownout.
Recharge and reset; refusing to reformat to protect your secrets.json.
Boot/reset reason: 1 (POWERON_RESET)
```

@brentru — this answers your "are you able to see the boot reason?" question: yes, now we can. And @tyeth, it backs up your point that the reason varies. It came back as **POWERON_RESET (1), not brownout (15)** — which makes sense: a full drain is a clean power-off followed by a cold boot, so the brownout detector never trips. The halt message's hard-coded "likely a brownout" wording is actually wrong for this case, so I'll soften it to "likely a brownout or power-loss recovery."

## 4. Reset on USB power → still halts

I pressed reset with the board on USB. It halted again. That's an important clue: on USB the flash chip gets a stable 3.3V, so if the stored bytes were fine, the mount would succeed. It didn't — so this is **not** a low-voltage read that clears up on recharge. The filesystem is genuinely unreadable. Recharging won't fix it, because recharging never rewrites flash.

## 5. Flash dump → exactly what survived

I dumped the `ffat` partition and looked at it byte by byte:

```
esptool read-flash 0x310000 0xF0000 ffat.bin
```

| Region | State |
|---|---|
| FAT boot sector (first 4 KB at `0x310000`) | **all `0xFF` — erased** |
| Root directory | intact (file entries present) |
| `secrets.json` | **intact — 206 bytes, parses as valid JSON, all fields present** |
| `wipper_boot_out.txt` | intact |

So **only the 4 KB boot sector was wiped. Everything else, including `secrets.json`, is byte-perfect.** The guard did its job — it refused to reformat, and the credentials survived untouched.

@brentru, on diffing beta.130 / pr931.2: this is the difference in one picture. The old code formatted on a failed mount, which would have erased that perfect `secrets.json`. This PR returns `false` and lets the guard protect it.

## 6. So what erased the boot sector?

I went through every flash-writing call in the boot path to be sure we hadn't missed one:

- `eraseBootFile()` / `createBootFile()` only touch `wipper_boot_out.txt` — that's the FAT table, root directory, and data clusters, never sector 0. And `createBootFile()` already compares old vs. new content and skips the write when nothing changed (the diff-before-write idea from the earlier PR is already in here), plus it bails out on a brownout reset.
- The **only** code that writes the boot sector is `f_mkfs()` + `f_setlabel("WIPPER")` — and those run only during a format.

This cycle the guard blocked the format, and the reset reason confirms no format ran. So no firmware path wrote those `0xFF` bytes.

That leaves the power-down itself. A read can't erase flash. But as the rail collapses during a drain, commands on the SPI bus can get garbled, and a corrupted command can turn into a write-enable + sector-erase that wipes the 4 KB block at `0x310000`. That's a known failure mode for SPI flash on an unclean power-down, and it matches the evidence exactly: one sector gone, everything else perfect, no format in the logs.

@tyeth, this lines up with your read of it: there's **no file corruption** (the file is perfect) and it isn't even an interrupted firmware write — it looks like a stray flash erase as the power dies. The catch is that it does **not** heal itself on recharge; the board needs an active repair.

## 7. Recovery (the exact steps)

The board is bricked until something rewrites a valid boot sector. Since the data is intact, recovery keeps the credentials. **Important: no full `erase_flash`** — only targeted region erases, so we don't lose anything we don't have to.

```
# 1. Save the intact secrets.json out of the dump (one-time, for restoring)
# 2. Targeted erase of just the filesystem + the provisioned marker:
esptool erase-region 0x310000 0xF0000     # ffat
esptool erase-region 0x9000  0x5000       # nvs (clears the "provisioned" flag)
# 3. Reset -> firmware sees an unprovisioned, blank FS and writes a fresh,
#    valid filesystem (this is the only step that legitimately writes sector 0)
# 4. Copy the saved secrets.json back onto the WIPPER drive, eject, reset
```

After that the board boots normally and reconnects. (One gotcha worth noting: an `erase-region` immediately followed by a read can hit a transient "port busy" while the flasher stub releases — just retry the read. I got bitten by it and it briefly looked like the erase had failed.)

## 8. What this means for the PR

- The fix does the right thing: a previously-provisioned board with an unreadable filesystem refuses to reformat, and the dump proves the credentials survive. That's the part this PR set out to do, and it holds up.
- The real open problem is upstream of the firmware: the boot sector gets erased as the power dies. Recovering the board afterward (a read-only rescue drive, a reformat gesture, etc.) is a band-aid over that, not a fix — so I don't want to commit to one here. Better to keep the focus on the core issue and not paper over it.
- Worth chewing on for the core issue: whether there's anything we can do to make the boot-sector window smaller or the flash less likely to take a stray erase as the rail collapses, and whether the intact FAT/root/data could be used to repair just the lost sector instead of a full reformat. No conclusions yet — flagging it as the thing actually worth solving.

One small thing I'd like to fold into this PR: fixing the halt message wording, since "likely a brownout" is wrong for a full drain — I'll change it to "brownout or power-loss recovery." Thank you both for the careful reads on this one.

flavio-fernandes:
Heads-up on a follow-on that builds directly on this PR: **#942 (boot-region rescue)**.

This PR (#937) stops a power-loss-erased board from being **reformatted**, so the data is never lost — but the board still can't boot; it halts and waits for a human. #942 is the other half: it lets that same board **repair itself and come back up automatically**.

**What #942 does.** From the flash dumps here, the damage on a hard drain is always the same shape: exactly one 4 KB block at the start of the `ffat` partition (`0x310000`) gets erased to `0xFF`. That block holds the **boot sector + both FAT tables**; the root directory and file data (including a byte-perfect `secrets.json`) survive just past it. FAT's own redundancy doesn't help — its two FAT copies both live in that one erased block — and a FAT12 super-floppy has no backup boot sector. So #942 keeps our own: after a clean mount it snapshots that boot region into NVS (separate flash, survives a format), and on a later boot, if the region reads back entirely `0xFF`, it restores it and re-mounts. It's heavily guarded — it only writes when the region is *entirely* blank and the backup passes a CRC32 + boot-signature check — so it can only help, never make things worse.

**Why it needs this PR.** #942 can't stand on its own. On a `main` base, `initFilesystem()` reformats on a failed mount *before* the restore could ever run, so the boot region is gone before we look at it. This PR's change — return `false` instead of auto-formatting — is exactly the precondition the rescue needs. Together they're complete: **restore the boot region if we can; if we can't, this PR's guard refuses to reformat so the secrets are still safe.**

**Testing.** Because of that dependency, #942 can only be validated *on top of* this PR. I put the two together on `test/fs-937-plus-942-rescue` and tested on a MagTag. You can reproduce the exact failure in seconds without a battery drain:

```
esptool erase-region 0x310000 0x1000   # wipe just the boot sector + FATs
# reset
```

On `main`+#942 that wipes the FS (reformat-before-restore). On the combined branch the board restores the boot region from NVS and boots straight back with `secrets.json` intact. A full battery-drain run is in progress to confirm the real-world path too.

No action needed here — just flagging the relationship so the review of #942 has the full picture. The cleanest merge order is this PR first, then #942 rebased on top.

flavio-fernandes:
**On-hardware end-to-end result: a real battery drain triggered the rescue, and the board recovered itself — `secrets.json` survived, no reformat.**

Following up on @brentru's "are you able to see the boot reason?" and @tyeth's "no corruption, just a misread of a failed flash op" — I took the combined branch (`test/fs-937-plus-942-rescue`, this PR's no-auto-format change plus the boot-region rescue in #942) to the bench, drained a provisioned MagTag flat, and powered it back on over USB. It came back online on its own. Here is exactly what the firmware recorded, straight from the serial log:

```
[fs] firmware: 1.0.0-rescue-test.6
[fs] boot reset reason: 1 (POWERON_RESET)
[fs] boot-region rescue this boot: RESTORED from backup
[fs] boot-region rescues so far: 1 (evaluated 1 time(s))
[fs] last rescue eval: reset reason 1 (POWERON_RESET), region 4096 bytes, 0 non-0xFF byte(s)
...
Connected to WiFi!
Registering hardware with WipperSnapper...
Registering hardware with IO...
```

So the full chain happened by itself: power-on after the drain → boot region came back blank → rescue restored it from the NVS backup → FAT mounted → the board re-registered with Adafruit IO using its stored credentials. No `f_mkfs()`, no lost provisioning.

**@brentru — yes, we can see the boot reason now, and it's durable.** It's read at boot and persisted to NVS, so it's readable on the recovery boot *and* on any later boot (the `[fs]` block above heartbeats in the run loop). Worth noting this recovery boot logged `POWERON_RESET (1)`, not a brownout code — which lines up with @tyeth's point that the reset reason varies between these reboots. That's exactly why the rescue gates on *blank-region + valid-backup*, not on the reset code: the reason is for diagnosis, not the decision.

**@tyeth — your "no corruption, just a misread failed flash op" read holds up, with one refinement.** The `0 non-0xFF byte(s)` over a 4096-byte region is the key measurement: the drain left the FAT boot region *cleanly erased to all-`0xFF`*, not scrambled with random garbage. The stored file bytes past that region were never touched. So the mount genuinely fails (the boot sector + both FAT tables live in that erased 4 KB — see the #942 writeup), but nothing is *corrupt* in the destructive sense; it's a recoverable erase. The old code's mistake was treating that failed mount as "reformat me." This run restores instead.

**Cross-referencing #942's "blank enough" finding.** #942 documents an *earlier* drain where the erased region wasn't perfectly clean — it had a single stray `0x7F` byte at `0x310C4C`, the fingerprint of an erase interrupted as Vcc collapsed. That's why the rescue rule is "blank *enough*" (no `0x55AA` boot signature + at most `len/64` non-`0xFF` stragglers) rather than "entirely `0xFF`." This latest drain is the other end of that same spectrum — a *fully clean* erase, `0` stragglers — and the same gate handled it. Between the two drains we've now seen both the interrupted-erase case (#942) and the clean-erase case (here), and the rescue restored in both without ever being at risk of matching live filesystem data (which carries the boot signature and thousands of non-`0xFF` bytes).

**Byte-verify of the survivor.** I dumped the `ffat` partition (`0x310000`, 0xF0000) off-chip after the recovery and checked the bytes directly:
- The restored boot region reads back as a valid FAT boot sector — `EB FE 90 "MSDOS5.0"` BPB at offset 0, and the `55 AA` boot signature at offset `0x1FE` — i.e. the previously-blank 4 KB is whole again, not reformatted.
- `secrets.json` is intact with all five credential fields present and well-formed: `io_username`, `io_key`, `network_ssid`, `network_password`, `network_type_wifi`. (Values redacted here; I only verified the keys parse.)

So this isn't "we skipped the format and hoped" — the survivor's bytes are confirmed on the flash.

Net: this PR's change (don't auto-format a provisioned board on mount failure) is the precondition, and #942's rescue turns "halt and wait for a human" into "repair and boot." The drain → recover → reconnect cycle now runs hands-off. Logs have SSID/credentials redacted; happy to share more of the capture if useful.

flavio-fernandes:
**A second drain just exposed a real data-loss bug this PR doesn't cover — I've moved #937 to draft. We need a lot more testing before I'd call this verified.**

Quick context: the *last* drain recovered perfectly (boot-region rescue fired, `secrets.json` survived, board reconnected on its own). So I drained the same board again. This time it did **not** recover — and the reason is important, because it's a failure mode neither #937 nor #942 addresses, and it's the same class of bug @tyeth has been pointing at all along.

### What happened

Every boot **mounted cleanly** — the FAT boot region was intact, the rescue was never even needed:

```
[fs] boot-region rescue this boot: not evaluated (mounted cleanly)
```

But the board halted in a loop on:

```
ERROR: Invalid IO credentials in secrets.json! TO FIX: Please change io_username and io_key ...
```

I pulled the `ffat` partition off-chip to see the actual bytes. Two findings:

1. **The real `secrets.json` is still physically on the flash, byte-intact** — the genuine `io_username` / `io_key` / SSID are sitting in an **orphaned cluster**, fully readable. Nothing corrupted or erased the data.
2. The **live** `secrets.json` directory entry now points at a freshly-written **placeholder template** (`YOUR_IO_USERNAME_HERE`, default brightness). The good file was *unlinked and replaced*, not damaged.

### Root cause

In `parseSecrets()`, an intact `secrets.json` read back **empty** during the power-loss-recovery boot (SPI rail sags → bytes come back `0x00`/`0xFF` → `deserializeJson()` returns `EmptyInput`). The existing code treats `EmptyInput` as "genuinely empty file → delete and recreate from template," guarded **only** by `WS.brownOutCausedReset`:

```cpp
if (error == DeserializationError::EmptyInput) {
  if (WS.brownOutCausedReset) {
    fsHalt(...);                       // protected
  } else {
    wipperFatFs.remove("/secrets.json");
    createSecretsFile();               // ← wrote the placeholder over my creds
  }
}
```

The recovery boot reported **`POWERON_RESET (1)`, not brownout (15)**, so `WS.brownOutCausedReset` was false and the destructive branch ran. **We deleted a working file because a single read failed.**

This is precisely:
- **@tyeth's "misinterpretation of a failed flash operation as a corrupted state"** — exactly that, just in the secrets path rather than the FAT path.
- **@tyeth's "the boot reason varies between those reboots (not always brownout)"** — the guard *depends* on the brownout flag, and the brownout-induced bad read surfaced on a POWERON boot. The reset reason is not a reliable safety gate. (This is the same lesson #937 already applied to the *format* decision by using a persistent NVS provisioned-marker instead of the reset reason.)

### Fix (pushed to this branch as a draft for review)

Commit `f8bb6090` gates the recreate on the same `wasFilesystemProvisioned()` NVS marker #937 uses for formatting, not on the reset reason: a previously-provisioned board that reads `secrets.json` back empty now **halts-and-protects** (recharge & reset) instead of recreating it. A never-provisioned board still bootstraps a template (correct first-boot behavior). MagTag build green, native test still passes.

### Why I'm marking this draft and asking for more testing

This is the part I want to be candid about. We had **one** successful drain and I was ready to call the rescue validated — then the very next drain found a *different* destructive path that our instrumentation only caught because we now dump the filesystem and read it back byte-for-byte. That tells me:

- **Brownout failures are not one bug, they're a family.** The boot region erase (#942), the auto-format-on-mount-failure (#937), and now the empty-read recreate are three distinct paths to the same "provisioned board loses its identity" outcome. There are very likely more (e.g. `createBootFile()` writes, the `configFileExists()` → `createSecretsFile()` path in `initFilesystem()`, any other "file looks wrong → rewrite it" logic), each needing the same provisioned-marker discipline.
- **Pass/fail on a single drain proves nothing.** The outcome depends on *which* 4 KB block the collapsing rail happens to disturb and *when* in the boot the read lands. We need many drains, and ideally fault-injection (forcing empty reads / single-block erases at specific points), not just battery cycles.
- **The reset reason being unreliable is now confirmed twice** — it can't gate any protective decision; only persistent off-FAT state (NVS markers, the boot-region backup, and ideally a `secrets.json` backup) can.

So I'd rather hold #937 (and #942) in draft until we've run a proper matrix: repeated drains, fault injection on each "rewrite" path, and a byte-level survivor check after each. I'll keep posting dumps. Grateful for the earlier skepticism — it's exactly what's bearing out here. 🙏

(All values redacted; the orphaned real credentials were verified locally only and never logged.)

flavio-fernandes:
Added a [`todo.md`](https://github.com/flavio-fernandes/Adafruit_Wippersnapper_Arduino/blob/flavio/fs-nvs-provisioned-guard/todo.md) to this branch (commit 3d99d776) — a running, **request-for-comment** checklist of the brownout/power-loss data-loss hazards turned up while validating this PR and #942 on hardware. Several are still unfixed, and I'd genuinely rather have more eyes on the scope than carry it solo.

**Update:** since I first posted this, another battery drain walked straight through item #2 below (the `configFileExists()` twin) — live confirmation it was a real, not theoretical, door — so I've now **fixed it** (commit d5913680, ordered right before the `todo.md` commit). That leaves item #1 as the remaining must-fix-before-undraft.

The one I'd now flag as *must-fix before undraft*:
1. **`WS.brownOutCausedReset` is set too late to matter** — it's set in `printDeviceInfo()` during `connect()`, but the whole FS constructor + `parseSecrets()` run earlier in `provision()`, so every `if (WS.brownOutCausedReset)` guard in the filesystem path is effectively always-false / dead. **And the empirical kicker:** that flag is set only on reset code `15` (brownout), but across **~2,400 boots over every drain cycle logged, code 15 appeared exactly 0 times** — every power-loss recovery comes back as `POWERON_RESET`. So gating FS-destructive ops on the reset reason protects ~0% of real drains even with the timing fixed. (Full distribution table in `todo.md` #1.) This is the measurement behind @tyeth's "the boot reason varies (not always brownout)" — on this hardware it's *never* brownout.

**Now fixed — item #2 (`configFileExists()` → `createSecretsFile()` twin).** This was the un-fixed twin of the recreate bug patched in `parseSecrets()`: a single first-byte misread (`0x00`/`0xFF`) on a provisioned board triggered a placeholder-template overwrite, and it wasn't gated on the NVS provisioned-marker — and because it runs earlier (in the constructor, before `parseSecrets()`), the `parseSecrets()` guard never got a chance. A drain reproduced it exactly: the board mounted cleanly, peeked `secrets.json` blank on a `POWERON_RESET` boot (not brownout), recreated the template, and halted on *"Please edit the secrets.json file"*; an off-chip `ffat` dump showed the real credentials byte-intact in an orphaned cluster next to the fresh template. Fixed by gating the recreate on the same persistent provisioned-marker — a provisioned board that can't read `secrets.json` now halts-and-protects instead of overwriting.

**`todo.md` #8 (rescue window vs. stochastic erase).** Repeated drains of the same board show the power-loss erase lands on a *different* 4 KB block each time: drain #1 hit the boot region (the #942 rescue's window → it self-healed ✅), but drains #4/#7/#9 hit **metadata** (the root directory) while the boot region stayed intact — so the rescue correctly stood down and #937 correctly halted, but the damage was entirely outside the 4 KB backup window. `secrets.json`'s bytes survived but were orphaned. Drain #9 sharpened this further: the casualty was a **single 32-byte directory entry** — the volume stayed fully mountable (valid boot sector, clean FAT) yet simply no longer listed `secrets.json`, with the creds still byte-intact at their cluster. So a power-loss can drop one live dir record, not just a whole root-dir sector. Drain #10 then showed the flip side: the *same* first-root-dir-sector erase as #7 but a **mountable** volume with `secrets.json` orphaned — so a given erase location doesn't map to a fixed outcome (unmountable vs mountable-with-orphan is itself variable). A region/boot-sector backup can't cover any of it. Takeaway: a fixed-region backup can't guarantee "survive any single-block erase"; the most robust fix is to back up **`secrets.json` content** to NVS and re-materialize it, independent of which block dies.

**`todo.md` #12 (new — #942 restore trigger is too narrow).** A later drain turned up a failure mode *inside* the rescue's own window that it still didn't catch: the boot region (`ffat 0x310000`, boot sector + FAT) came back as **high-entropy garbage** (3767/4096 bytes non-`0xFF`, no valid `0x55AA`), while the rest of the volume — including a byte-valid `secrets.json` with the real creds — was completely intact, and a **valid boot-region backup was sitting in NVS**. The rescue still stood down, because `restoreBootRegionIfBlank()` only restores a *blank* region (no `0x55AA` **and** ≤~`len/64` stragglers); a *corrupted* region has thousands of stragglers, so it fails the test exactly when restoring the known-good backup would have made the volume mountable again. The backup and the window were both right — only the trigger was wrong. Cheap, high-value fix: restore whenever the live boot sector is **structurally invalid / unmountable** (no valid `0x55AA`), not only when it's blank.

**`todo.md` #11 (recoverability).** Follow-on from drains #4 and #7: once provisioned + damaged beyond the rescue's reach, the board is **deadlocked** in the protective halt with *no in-firmware way out* — re-flashing the app doesn't help (the damage is in `ffat`, the marker in `nvs`), and recovery required a host with `esptool` to `erase_region` both. A normal user with only the USB drive would have a bricked board. The halt is the right floor, but it needs a deliberate user-reachable override (button-hold-to-reformat, or a read-only recovery MSC after N failed boots).

@tyeth @brentru — when the heatwave lets up, would love your read on whether the proposed direction (gate *all* FS-destructive ops on the persistent provisioned-marker rather than the reset reason; back up secrets.json content rather than only a fixed flash region; and add a user-reachable reformat override) is the right call, and which items you'd consider blocking vs follow-up. No rush at all — just laying out the full map so it's not a surprise later. 🙏

tyeth:
Mentally filing that we could be checking the validity of boot reason in boot log txt file (as should never be blank if file exists), and that could be done if json was empty/corrupt. 
We should also attempt multiple reads and compare, for both config.json and bootlog, before considering a format necessary. Either being read okay twice meaning no format, or if possibly needing format then maybe we attempt to write an 8kb file first (meaning two blocks of erase/write cycle) and see if that errors (it won't affect anything) before doing the format (i.e. verify we have a stable enough power supply for a file/block erase+write cycle before doing the same many times in a format).

flavio-fernandes:
> Mentally filing that we could be checking the validity of boot reason in boot log txt file (as should never be blank if file exists), and that could be done if json was empty/corrupt. We should also attempt multiple reads and compare, for both config.json and bootlog, before considering a format necessary. Either being read okay twice meaning no format, or if possibly needing format then maybe we attempt to write an 8kb file first (meaning two blocks of erase/write cycle) and see if that errors (it won't affect anything) before doing the format (i.e. verify we have a stable enough power supply for a file/block erase+write cycle before doing the same many times in a format).

@tyeth I like the multiple-read idea as another sanity check, but I'm a bit worried about using an 8 KB write as the final "power is safe" test.

One thing the later #937 drain testing taught us is that flash operations themselves become unsafe in this voltage region. We saw two related failure modes:

- an intact secrets.json could read back empty during the sag, after which our code destructively recreated it;
- other drains actually erased/corrupted 4 KB flash regions as power collapsed, with the affected block varying between runs.

So two matching reads would definitely be better than trusting one read, but I don't think they prove that the supply is safe. And an 8 KB test file isn't really isolated on FAT -- creating it also updates FAT allocation and directory metadata. If power collapses during that test, we're deliberately doing erase/write operations at exactly the point we've found them to be unsafe.

I'm wondering if the better primary fix is further upstream: don't allow flash operations once the supply enters this region at all.

Your older Espressif #10445 testing seems particularly relevant here. You reproduced the corruption at roughly 3.0 V and were already experimenting with the brownout threshold / shutting flash and WiFi down before things became unstable. The ESP32-S2 BOD levels include approximately 2.98 V and 3.19 V, so perhaps we should investigate raising the trip point enough to trigger a brownout reset before reaching the flash-danger region.

Then the multiple reads, provisioned marker, no-auto-format rule, etc. remain valuable defense-in-depth, but ideally we never execute the questionable read/write/format paths at an unsafe voltage in the first place.

That feels safer to me than trying increasingly hard to determine whether a flash operation is trustworthy after we've already entered the marginal-power region.

Am I remembering your earlier brownout work correctly?

Reference discord conversation: [https://discord.com/channels/327254708534116352/355063529302392832/1522389909485125743](https://discord.com/channels/327254708534116352/355063529302392832/1522389909485125743)

Issue #898, opened 2026-03-25
Add recovery fallback for secrets.json file


brentru:
Sometimes the `secrets.json` is either deleted (accidentally) or corrupted. Instead of this, @tyeth  and I were discussing a way to store, on boot, the `secrets.json`'s content in-memory and persist it through boot cycles. 

On boot, the device would:
1) check for the existence of this content, recover it and place into a new secrets.json file on the FS
2) create a new `secrets.json` file (current operation)

We'd need to:
* Profile the secret.json max. possible size
* Consider using a struct or just a string depending on serialization/deserialization difficulty.

PR #942, opened 2026-06-27
feat(fs): auto-repair a power-loss-erased FAT boot region from NVS backup

flavio-fernandes:
> [!IMPORTANT]
> **Depends on #937.** On its own this PR cannot recover anything: on a plain `main` base, `initFilesystem()` reformats the volume the moment a mount fails — *before* the restore hook can run — so the boot region (and `secrets.json`) is already gone. #937 is what changes `initFilesystem()` to return `false` instead of auto-formatting, which is the precondition this rescue needs. The two are partners: #937 stops a power-loss-erased board from being reformatted (no data loss); this PR lets that same board **repair itself and boot** instead of only halting. Full end-to-end validation lives on the combined branch `test/fs-937-plus-942-rescue` (see Testing).

## What this does

Adds an automatic, self-healing recovery for the "MagTag loses `secrets.json` after the battery dies" failure (#621, #898). After a clean mount, the firmware keeps a backup of the FAT **boot region** (boot sector + reserved sectors + both FAT tables) in NVS. On a later boot, if that region comes back blank, it restores it from the backup and re-mounts — recovering the whole filesystem, untouched, with no reformat.

## Why — what actually breaks

On a full battery drain the on-board SPI flash can misread a command as Vcc collapses and erase a single 4 KB block. We reproduced this on a MagTag and dumped the flash to see exactly what happened:

- Exactly the **first 4 KB** of the `ffat` partition (`0x310000`) came back blank.
- Everything past it was **byte-perfect** — the root directory, and a valid 206-byte `secrets.json`.

The catch is what lives in that first 4 KB. It is not just the 512-byte boot sector — it is the **boot sector + both FAT tables**. The root directory and file data start right after, at `0x1000`. So the volume won't mount even though every file is physically intact.

Two bits of built-in FAT redundancy don't save us here:

- FAT keeps **two FAT copies** — but a single 4 KB erase takes *both* at once.
- FAT32 keeps a **backup boot sector** (sector 6) — but this volume is FAT12 (`f_mkfs(FM_FAT | FM_SFD)`) and has none.

So we add our own backup, the same way a filesystem copies its superblock.

## How it works

- **Backup** (`backupBootRegion`): after a healthy mount, read the boot sector, parse the BPB to size the region (`reserved + numFATs * FATSz`, rounded up to the 4 KB erase block), read those bytes, and store them in NVS (namespace `ws_boot_bk`) with their length and a CRC32. NVS lives in a separate flash area that survives a format and is very unlikely to be hit by the same erase. The write is skipped when the stored copy already matches, so it costs nothing on a normal boot.
- **Restore** (`restoreBootRegionIfBlank`): on a failed mount, if the live region is blank and a good backup exists, write the backup back and let the constructor re-mount.

In the constructor the order is: try mount → if it failed, try the boot-region restore and re-mount → only then fall through to #937's careful format/halt gates.

## "Blank enough", not "entirely 0xFF" — a finding from a real drain

The first cut of this rescue restored only when the live region was **entirely** `0xFF`. On the bench that always worked, because forcing the failure with `esptool erase-region 0x310000 0x1000` produces a *perfectly clean* all-`0xFF` block. A genuine battery drain does **not** erase that cleanly.

After a real full drain, the dump showed the 4 KB region almost entirely `0xFF` — but with **one stray `0x7F` byte** at `0x310C4C`. That single byte is the fingerprint of an **erase interrupted partway**: as Vcc collapsed mid-erase-cycle, one byte didn't finish erasing (top bit still programmed → `0111 1111`). The strict "entirely `0xFF`" rule (correctly, per its own wording) refused that case, so the board sat in #937's protective halt loop instead of self-healing.

So the rule is now **blank *enough***: restore when the live region has **no valid `0x55AA` boot signature** and **at most `len/64` non-`0xFF` stragglers** (64 bytes for a 4 KB region). A populated boot region carries a boot signature and *thousands* of non-`0xFF` bytes (FAT free entries are stored as `0x00`), so this can never match live filesystem data — it stays paranoid, just no longer fooled by an incomplete erase.

## Paranoid by design

It will only ever write to flash when all of these hold, and is a complete no-op otherwise — so it can help but never make a bad situation worse:

1. The live boot region carries **no valid `0x55AA` boot signature** and is **blank enough** — at most `len/64` bytes differ from `0xFF`. If it looks structured (a signature, or many non-`0xFF` bytes), the FS is in some state we don't understand, and we do **not** touch it.
2. A backup must exist in NVS with a matching stored length.
3. The backup must pass a **CRC32** check and carry a valid **`0x55AA` boot signature**.
4. The flash chip must report a healthy **JEDEC ID** before any write.
5. The BPB geometry must be the expected 512-byte-sector layout and fit a small size cap; anything odd → skip.

Scope/footprint: ESP32-only (uses NVS via `Preferences`). The region buffers are heap-allocated, not stack, to stay off the small constructor stack. No change to the format decision or to any non-ESP32 board.

Honest limitation: the backup can be at most one write stale — if a write changes the FAT and power is lost before the next backup refresh, a restore brings back the previous good region (board mounts, `secrets.json` intact; you'd lose only the very last `wipper_boot_out.txt` rewrite). That's an acceptable trade for a rescue path.

## How it was tested

**Build:** `pio run -e adafruit_magtag29_esp32s2` → SUCCESS.

> [!NOTE]
> This rescue **cannot be tested on a `main`-only base** — without #937, the volume is reformatted on the failed mount before the restore can run, so you'd only ever see a wiped FS. All on-hardware testing below was done on `test/fs-937-plus-942-rescue` (this branch rebased on top of #937).

**On hardware — the real failure (the decisive test).** A provisioned MagTag was drained fully, then powered back. Before this change it stayed in the halt loop for hours; a flash dump confirmed the first 4 KB was blank (the `0x7F` straggler described above) with `secrets.json` intact past it. We then reflashed the app **only** — leaving that genuine drain damage and the NVS backup on the flash — and reset. The board **restored the boot region from NVS on the next boot and came straight back up**: WiFi connected, MQTT pinging, `secrets.json` intact, no reformat. The *same flash* had kept the strict "entirely `0xFF`" build halting; the blank-enough build recovered it.

**On hardware — forcing it without a battery drain.** This reproduces a *clean* version of the failure in seconds. With the board in ROM download mode:

```
# wipe the first 4 KB of the ffat partition (boot sector + FATs):
esptool erase-region 0x310000 0x1000
# reset the board
```

To exercise the **incomplete-erase** path specifically (the real-world shape), leave a straggler behind after erasing, e.g. write a single non-`0xFF` byte into the region (one `0x7F` at `0x310C4C`) before resetting.

Expected serial on the next boot:

```
[fs] boot region blank - restoring from backup...
[fs] boot region restored (4096 bytes). Re-mounting...
```

(To confirm the backup exists first, do a normal boot once after flashing; that prints `[fs] boot-region backup updated: 4096 bytes saved to NVS`.)

## Notes for reviewers

- The 4 KB erase size matches the SPI-NOR erase granularity, which is why the damage is always at least one 4 KB block aligned to `0x310000`. The backup size is derived from the BPB and rounded up to that block, so it adapts if a board's FAT geometry differs (with a small cap so the blob always fits the nvs partition).
- The `len/64` straggler tolerance is intentionally generous against incomplete erases yet far below real data: even a freshly-formatted region has thousands of non-`0xFF` bytes, versus the single straggler we observed. Combined with the no-signature requirement, it cannot match a structured boot region.
- Because of the dependency above, the cleanest path is to land #937 first and rebase this on top of it (which is what `test/fs-937-plus-942-rescue` already is). Happy to retarget this PR onto #937's branch if that's preferred.

### Known limitation — backup-write failures are not surfaced (deliberate, for simplicity)

`backupBootRegion()` is intentionally best-effort and **silent on failure**. If any step can't complete — `flash.begin()`/read fails, the BPB isn't recognized, the heap alloc fails, or the NVS `Preferences` write can't open or tears partway — it simply returns and leaves whatever backup was already in NVS untouched. It does **not** check the `putBytes`/`putUInt` return values or record a "backup failed/stale" breadcrumb. The consequence is a *diagnostics* gap, not a data-safety one:

- **It is still fail-safe.** A torn or inconsistent backup is caught on the *restore* side, which validates length, CRC32, and the `55 AA` signature before trusting the blob. A backup that fails those checks is rejected (`kRescueBackupBad` / `kRescueNoBackup`) and the board falls through to #937's protective halt (recharge & reset, refuse to reformat) rather than writing garbage over the boot region. A failed backup can never cause data loss — at worst it means "no auto-repair this time."
- **What you lose is early warning.** You won't learn the backup is missing or stale until a later restore needs it and reports it. Combined with the one-write-stale window noted above (the backup only catches up on a clean boot, so a re-provision immediately followed by a drain could restore a slightly-older boot region), the board could decline to self-heal a case it "should" have covered — and you'd only find out after the fact.

This was left as-is on purpose to keep the rescue path small and obvious. The additive fix would be to check the write result and persist a `last_backup_ok` (and/or backup age) breadcrumb into the existing `[fs]` diagnostics block, so a board self-reports a stale/missing backup *before* a drain needs it.

**Question for reviewers (@tyeth, @brentru):** is documenting this as a known, fail-safe limitation acceptable for now, or would you consider surfacing the backup result in diagnostics a must-have for this PR? Happy to fold the breadcrumb in if you'd prefer it land here rather than as a follow-up.

### DIFF
--- a/src/provisioning/tinyusb/Wippersnapper_FS.cpp
+++ b/src/provisioning/tinyusb/Wippersnapper_FS.cpp
@@ -33,6 +33,10 @@
 
 #include "Wippersnapper_FS.h"
 #include "print_dependencies.h"
+#ifdef ARDUINO_ARCH_ESP32
+#include <Preferences.h> // NVS-backed store for the boot-region rescue backup
+#include <stdlib.h>      // malloc/free for the (heap, not stack) region buffers
+#endif
 // On-board external flash (QSPI or SPI) macros should already
 // defined in your board variant if supported
 // - EXTERNAL_FLASH_USE_QSPI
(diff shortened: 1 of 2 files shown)


flavio-fernandes:
**End-to-end on-hardware validation: the rescue fired on a real drain and recovered the board.**

Companion data point to the "blank enough" finding in the PR description (full writeup on #937). I drained a provisioned MagTag flat on the combined branch (`test/fs-937-plus-942-rescue`) and powered it back on over USB — it recovered on its own:

```
[fs] boot reset reason: 1 (POWERON_RESET)
[fs] boot-region rescue this boot: RESTORED from backup
[fs] boot-region rescues so far: 1 (evaluated 1 time(s))
[fs] last rescue eval: reset reason 1 (POWERON_RESET), region 4096 bytes, 0 non-0xFF byte(s)
```

Where the drain documented in the description above was an *interrupted* erase (one stray `0x7F` at `0x310C4C`), this one is the **clean-erase** end of the same spectrum: `0` non-`0xFF` stragglers across the whole 4 KB region. The "blank enough" rule (no `0x55AA` signature + ≤ `len/64` stragglers) covered both. After the restore the FAT mounted, the board reconnected to Adafruit IO, and an off-chip `ffat` dump confirms the boot region is whole again (valid `MSDOS5.0` BPB + `55 AA` signature) with `secrets.json` byte-intact. No reformat, no human in the loop.

This is the piece this PR adds on top of #937: #937 stops the wipe, and this turns the protective halt into a self-repair + boot.

flavio-fernandes:
## Field note: recovering a board by hand-restoring the boot region from this PR's NVS backup — and what it says about the restore trigger

A deep battery drain on a board running this rescue (combined `1.0.0-rescue-test.8` build) produced a case this PR **almost** handles automatically but currently doesn't — and recovering it by hand turned into a nice end-to-end validation that the backup itself is sound. Writing up the procedure and the logic, both as a recovery recipe and as motivation for one small change.

### What the drain did

The power-loss corrupted the **boot region** (`ffat 0x310000`, boot sector + FAT, first 4 KB) into **high-entropy garbage** — *not* a clean `0xFF` erase:

- boot signature gone: offset `0x1FE` read `f7 fb`, not `55 AA`; sector 0 started `fb fe 97 cf …` instead of a FAT `EB ..`
- 3767 of 4096 bytes non-`0xFF`

**Everything else was intact** — the root directory parsed cleanly and `secrets.json` (cluster 13) still held valid credentials, byte-complete. So the volume was unmountable purely because its boot sector + FAT were scrambled.

### Why the rescue stood down (the gap)

`restoreBootRegionIfBlank()` only restores when the live region is **"blank enough"** — no `0x55AA` **and** ≤ ~`len/64` (≈64) non-`0xFF` stragglers. A drain that *corrupts* rather than *erases* leaves **thousands** of non-`0xFF` bytes, so `3767 ≫ 64` fails the test and the rescue hands off — even though a **valid backup of exactly that region was sitting in NVS** (`ws_boot_bk`) and the damage was squarely inside the backup's window. #937 then (correctly) halted: unmountable + provisioned → refuse to reformat.

So: backup right, window right, only the **trigger** was too narrow. The heuristic was tuned for the clean-erase case and is blind to in-window corruption.

### The hand-recovery (recipe)

Board in ROM download mode (`/dev/cu.usbmodem01`); `esptool` = `~/.platformio/packages/tool-esptoolpy/esptool.py --chip esp32s2 --before no_reset --after no_reset`.

**1. Dump ffat + nvs (read-only):**
```
esptool read_flash 0x310000 0xF0000 ffat.bin    # ffat partition
esptool read_flash 0x9000  0x5000  nvs.bin       # nvs partition (holds the backup)
```

**2. Confirm the diagnosis** from `ffat.bin`: boot sector has no `55AA` at `0x1FE`; root dir (`0x1000+`) and the `secrets.json` dir entry + data cluster are intact. (If the *data* were gone, this recipe wouldn't apply — you'd reprovision instead.)

**3. Extract this PR's backup from `nvs.bin`.** The backup lives in NVS namespace `ws_boot_bk`, keys `blob` (the bytes), `len`, `crc` (CRC32, poly `0xEDB88420`). Parsing gotchas that bit me:
- Honor the **entry-state bitmap** (32 bytes at page offset `0x20`, 2 bits/entry: `0b10`=written, `0b00`=erased). Skip non-written entries or you'll pick up superseded copies.
- A 4 KB blob is split into **BLOB_DATA chunks** (here 2) plus a **BLOB_IDX**; concatenate chunk data in chunk-index order.

**4. Validate by filesystem consistency, NOT by the backup's own CRC.** This is the important subtlety: the NVS copy here was **torn across GC generations** (`chunk0@seq39 + chunk1@seq40`), and the stored `crc` key matched *no* assembly. But the assembled 4 KB was still byte-correct — proven by checking the reconstructed **FAT chains against the live directory**:
```
wipper_boot_out.txt: dir cluster 3, 13698 B -> 27 clusters; FAT chain [3,4,14,15,16,17,...] -> EOF  ✓
secrets.json:        dir cluster 13, 207 B  ->  1 cluster;  FAT chain [13] -> EOF                   ✓
```
The chain correctly skips the secrets clusters — a real fragmented allocation that exactly matches the on-disk layout. (Entry-header CRCs were all off by a *constant* XOR, which actually confirms byte-alignment was perfect — just a different CRC init than `zlib`.)

**5. Write the 4 KB back (only the boot region; data untouched):**
```
esptool write_flash 0x310000 bootregion.bin      # auto-erases the 4 KB sector first
esptool read_flash  0x310000 0x1000 verify.bin   # confirm read-back == bootregion.bin, 55AA present
```

**6. Reset.** Board mounts cleanly, `parseSecrets()` reads the original credentials, registers + connects — **no erase, no re-typing secrets.** On boot the rescue sees a valid `55AA` and leaves it alone (and re-backs-up the healthy region).

### What this argues for in this PR

1. **Broaden the restore trigger** from "blank enough" to **"structurally invalid / unmountable"** — restore the backup whenever the live boot sector lacks a valid `0x55AA` (or won't parse as FAT), whether the region is blank *or* garbage. An unmountable boot region has nothing worth preserving, so a known-good restore can only help. This single change would have made the board above self-heal with zero human involvement.
2. **Don't gate the restore solely on the backup's stored CRC.** The NVS copy can be torn mid-GC (stale `crc`/`len` vs. live `blob` chunks). A belt-and-suspenders check — does the restored FAT chain the existing root-dir entries? — is what actually proves the bytes are usable, and is cheap to do in firmware before committing the write. (Related: the one-write-stale limitation already noted on this PR.)

The protective halt (#937) remains the correct floor — no data was destroyed — but with (1) this class of in-window corruption becomes an automatic recovery instead of a bench session with `esptool`.

_(All offsets/dumps here are from real hardware; credentials were inspected only to confirm byte-integrity locally and are not reproduced.)_

PR #943, opened 2026-07-02
api-v2: use string pin names for i2c bus and uart rx/tx pins

tyeth:
## Summary

Companion to adafruit/Wippersnapper_Protobuf#207 — the i2c `AddressSpace.pin_scl`/`pin_sda` and uart `Descriptor.pin_rx`/`pin_tx` protobuf fields are now **strings** instead of `uint32`, so bus scans and components can reference named pins like `"SDA"`/`"SCL"` (as the CircuitPython/Blinka firmware does) and stay consistent with every other pin field in the API (`"D22"`, `"A4"`, `"EXP_0x48_0"`, ...).

## Changes

- **Regenerated** `src/protos/i2c.pb.*` and `uart.pb.*` with nanopb 0.4.8 from the updated protos (pin fields are now `char[32]`)
- **New `WsPinNameToNum()` helper** (`src/helpers/ws_helper_pins.h`): resolves `"22"`, `"D22"`, `"GPIO22"`, `"SDA"`/`"SCL"` (and `"SDA1"`/`"SCL1"` where a second Wire port exists) to pin numbers, matching the existing `atoi(name + 1)` convention used by the other components
- **I2cController** resolves pin names once at the bus layer, so `"SCL"` and its GPIO number dedupe to the same physical bus; `GetOrCreateI2cBus`/`GetI2cBus`/`findOrCreateBus` now take pin-name strings (display/expander/GPS call sites pass through unchanged)
- **Drivers echo what IO sent**: `drvBase` stores the original pin-name strings and I2C device events reproduce them verbatim in the event descriptor
- **UART**: `UARTHardware`/`drvUartBase` resolve the RX/TX names (port identity remains the resolved RX pin number), and UART input events echo the original RX pin name
- **Offline (SD card) config**: `ParseI2cDeviceAddReplace` now actually reads `i2cBusScl`/`i2cBusSda` from the JSON (the previous code assigned in the wrong direction, leaving the pins zeroed), accepting either strings or legacy integer pins and defaulting to `"SCL"`/`"SDA"`
- **ESP8266 GCC compatibility**: replaced `msg = ws_..._init_zero;` re-assignments with `memset` in the error handler / signal encoder — the older GCC rejects braced-list assignment once the initializer contains string-literal char-array members (declarations are unaffected)

## Build verification

| env | result |
|---|---|
| `adafruit_feather_esp32s3` | ✅ SUCCESS |
| `huzzah8266` | ✅ SUCCESS |
| `raspberrypi_picow` | ⚠️ compiles, fails at link on a **pre-existing** duplicate NAU7802 `lib_deps` entry (registry + git URL both listed in `[env]`) — unrelated to this change, will be fixed separately |


src/helpers/ws_helper_pins.h
```diff
@@ -0,0 +1,64 @@
...
+ * BSD license, all text here must be included in any redistribution.
+ *
+ */
+#ifndef WS_HELPER_PINS_H
+#define WS_HELPER_PINS_H
+#include <Arduino.h>
+#include <ctype.h>
+
+/*!
+    @brief  Resolves a pin name string to a pin number. Accepts numeric
```
tyeth, line 44: Curious where expanders will fit into this if ever. Also if this is the route we want then all other `atoi` with string + 1 will need replacing/augmenting too.

tyeth (commented):
OK, possibly needs a rethink, or expanding to cover the other classes (new PR?)

tyeth:
@brentru can I grab your eyes on this one please

src/components/error/handler.cpp
```diff
@@ -38,7 +38,13 @@ static bool encode_string_callback(pb_ostream_t *stream,
 /*!
     @brief  ErrorHandler constructor
 */
-ErrorHandler::ErrorHandler() { _d2b_msg = ws_error_D2B_init_zero; }
+ErrorHandler::ErrorHandler() {
+  // NOTE: memset, not `_d2b_msg = ws_error_D2B_init_zero;` - ESP8266's GCC
+  // rejects braced-list re-assignment now that the initializer contains
+  // string-literal char-array members (the pin name fields). memset is
+  // bit-identical to init_zero for this all-static struct.
```
brentru, line 45: The error handler fixes should be in another PR against `migrate-api-v2`...

src/components/i2c/drivers/drvBase.h
```diff
@@ -861,8 +863,8 @@ class drvBase {
   uint32_t _i2c_mux_addr;    ///< The I2C MUX address, if applicable.
   uint32_t _i2c_mux_channel; ///< The I2C MUX channel, if applicable.
   char _name[15];            ///< The device's name.
-  uint32_t _pin_scl;         ///< The SCL pin number for this driver's bus.
-  uint32_t _pin_sda;         ///< The SDA pin number for this driver's bus.
+  char _pin_scl[32];         ///< The SCL pin name for this driver's bus.
+  char _pin_sda[32];         ///< The SDA pin name for this driver's bus.
```
brentru: The size of these char's and (`name`) should be `#define`'d at the top of the file.

It should exactly match what's in the `i2c.pb.h` file. For example, `name` here should reflect,
`char name[16]; /* * The I2c device's name. * */`

src/components/i2c/drivers/drvBase.h
```diff
@@ -861,8 +863,8 @@ class drvBase {
   uint32_t _i2c_mux_addr;    ///< The I2C MUX address, if applicable.
   uint32_t _i2c_mux_channel; ///< The I2C MUX channel, if applicable.
   char _name[15];            ///< The device's name.
-  uint32_t _pin_scl;         ///< The SCL pin number for this driver's bus.
-  uint32_t _pin_sda;         ///< The SDA pin number for this driver's bus.
+  char _pin_scl[32];         ///< The SCL pin name for this driver's bus.
```
brentru: `32` chars is pretty large, even for identifiers `PIN_SERIAL1`, `PIN_WIRE_SCL`, .... Maybe we could drop this (and in the `pb`) to 16?

src/components/i2c/controller.cpp
```diff
@@ -1034,22 +1034,31 @@ TwoWire *I2cController::GetI2cBusByIndex(size_t index) {
...
     @returns  Pointer to the I2cHardware bus, or nullptr if initialization
               failed.
 */
-I2cHardware *I2cController::findOrCreateBus(uint32_t pin_scl,
-                                            uint32_t pin_sda) {
+I2cHardware *I2cController::findOrCreateBus(const char *pin_scl,
+                                            const char *pin_sda) {
+  // Resolve the pin names to pin numbers
+  uint32_t scl_num, sda_num;
+  if (!WsPinNameToNum(pin_scl, scl_num) || !WsPinNameToNum(pin_sda, sda_num)) {
```
brentru: `scl_num` and `sda_num` are uninitialized before going into `WsPinNameToNum()`. There's no guarantee they get set.

I'm not 100% on this pattern as a result.

What about something like `inline int32_t WsPinNameToNum(const char *pin_name)`, which returns `-1` or a defined magic on failure.

Then the call would assign and validate:
```
uint32_t scl_num = WsPinNameToNum(pin_scl)
if (scl_num == -1)
  failure!
```

src/components/uart/drivers/drvUartBase.h
```diff
@@ -297,7 +305,21 @@ class drvUartBase {
   SoftwareSerial *_sw_serial; ///< Pointer to a SoftwareSerial instance
 #endif                        // HAS_SW_SERIAL
   char _name[32];             ///< The device's name
-  uint32_t _port_num = 0;     ///< The port number for the UART device
+  char _port_name[32] = {0};  ///< The RX pin name for the UART device
```
brentru: ensure this matches the PB

src/helpers/ws_helper_pins.h
brentru: We'd probably want to add a check at the front for `EXP_`

brentru (changes requested):
@tyeth I'm looking at a few areas that have a higher impact, first.

tyeth:
Pushed 4 commits fixing bugs found while bringing up i2c over the string-pin protos on real hardware (LilyGO T-Dongle C5, ESP32-C5):

- **fix(i2c): decode Probe payload as standalone submessage** — `address_spaces`/`addresses` are nanopb callback fields inside the `ws_i2c_B2D` `payload` oneof. nanopb zeroes the oneof member when it first sets the union tag, wiping callbacks pre-set on `b2d.payload.probe`, so a probe decoded to **0 address spaces** (silent empty scan, no error). Adds a shared `ws_pb_decode_oneof_submsg()` helper that finds the field and decodes it as a standalone message (callbacks survive).
- **fix(i2c): decode Add payload as standalone submessage** — same oneof-callback wipe on the settings map; device settings were silently dropped.
- **fix(expander): decode Add payload as standalone submessage** — same issue on the nested i2c cfg.
- **chore(i2c): warn when probe address list is empty** — `ProbeAddresses` only iterates caller-supplied addresses; the "empty = scan all 0x08–0x77" semantic isn't implemented, so an empty list silently finds nothing. Logs a warning until scan-all is added.

HIL-verified: after the probe fix, a full-range scan finds the attached HTU31D at 0x40 (SDA=GPIO11/SCL=GPIO12). checkin already uses a message-level `cb_payload` callback so it was unaffected.

src/helpers/ws_helper_pins.h
```diff
@@ -0,0 +1,64 @@
...
+  }
+#if defined(PIN_WIRE1_SDA) && defined(PIN_WIRE1_SCL)
+  if (strcmp(pin_name, "SDA1") == 0) {
+    pin_num = (uint32_t)PIN_WIRE1_SDA;
+    return true;
+  }
+  if (strcmp(pin_name, "SCL1") == 0) {
+    pin_num = (uint32_t)PIN_WIRE1_SCL;
+    return true;
+  }
```
tyeth: should probably duplicate for `if defined(SDA1) && defined(SCL1)`

src/helpers/ws_helper_pins.h
```diff
@@ -0,0 +1,64 @@
...
+#endif
```
tyeth: `#elif ` begins EXP, pass off to Expander controller / hardware to resolve, similar to the QueryPinState added in the i8080 display PR and related shared pin issue #952

src/components/error/handler.cpp
tyeth-ai-assisted: Split out as #954 (against `migrate-api-v2`) — byte-identical lines, so this PR will rebase over it cleanly once merged.

src/helpers/ws_helper_pins.h
tyeth: Added in #953

tyeth:
> OK, possibly needs a rethink, or expanding to cover the other classes (new PR?)

Covered in #953

src/helpers/ws_helper_pins.h
tyeth: covered in #953

src/components/digitalIO/hardware.cpp
```diff
@@ -133,6 +141,15 @@ bool DigitalIOHardware::ReadValue() {
...
 
+  // First read after add: _prv_value was only a guess (initial_value), so we
+  // cannot infer a real transition. Seed the baseline from this read and report
+  // it as the pin's initial state rather than comparing against the guess.
+  if (_first_read) {
+    _first_read = false;
+    _prv_value = _value;
+    return true;
+  }
+
```
tyeth: I'd prefer if we had the last state from the proto pinAdd in an initial write for an input pin that would be used as the starting previous value for the input on change trigger

tyeth: Raising as a nice to have feature for v2.1 instead.

tyeth-ai-assisted: Raised for v2.1: adafruit/Wippersnapper_Protobuf#213 (proto semantics, `v2.1` label) + #955 (firmware side). Nothing left blocking this thread for merge.

src/components/i2c/drivers/drvBase.h
tyeth-ai-assisted: Done in #953 (merged): `DRV_BASE_NAME_LEN`/`DRV_BASE_PIN_NAME_LEN` defined at the top of the file as `sizeof()` of the `ws_i2c_Add.name` / `ws_i2c_AddressSpace.pin_scl` pb fields — exactly matches i2c.pb.h and can't drift on regen (fixed `_name[15]` vs pb `name[16]` too).

src/components/i2c/drivers/drvBase.h
tyeth-ai-assisted: Raised as adafruit/Wippersnapper_Protobuf#214 — firmware buffers are sized via `sizeof()` of the pb fields since #953, so the regen with 16 propagates automatically.

src/components/i2c/controller.cpp
tyeth-ai-assisted: Done in #953 (merged): `int32_t WsPinNameToNum(const char*)` returning `WS_PIN_INVALID` (-1), call sites assign-and-validate exactly as sketched.

src/components/uart/drivers/drvUartBase.h
tyeth-ai-assisted: Done in #953 (merged): `_name`/`_port_name` sized via `sizeof()` of `ws_uart_Add.name` / `ws_uart_Descriptor.pin_rx`.

src/components/digitalIO/hardware.cpp
```diff
@@ -63,16 +66,19 @@ bool DigitalIOHardware::SetMode() {
   bool has_expander = (_expander_drv != nullptr);
 
   if (_direction == ws_digitalio_Direction_D_OUTPUT) {
+    // _value is the logical value; drive the electrical level (inverted for
+    // active-low pins).
+    bool electrical = _is_inverted ? !_value : _value;
```
brentru: Why are we converting `value` to `electrical` and then using a ternary operator within digital write?

src/components/digitalIO/hardware.cpp
```diff
@@ -118,11 +127,10 @@ void DigitalIOHardware::Write(bool value) {
 bool DigitalIOHardware::ReadValue() {
   bool has_expander = (_expander_drv != nullptr);
 
-  if (!has_expander) {
-    _value = digitalRead(_name);
-  } else {
-    _value = _expander_drv->digitalRead(_name);
-  }
+  bool raw =
```
brentru: Prefer not to use ternary ops.

src/components/digitalIO/hardware.cpp
```diff
@@ -133,6 +141,15 @@ bool DigitalIOHardware::ReadValue() {
 bool DigitalIOHardware::CheckEvent() {
   ReadValue();
 
+  // First read after add: _prv_value was only a guess (initial_value), so we
```
brentru: Why are we changing existing functionality here?

src/components/digitalIO/hardware.h
```diff
@@ -60,6 +61,8 @@ class DigitalIOHardware {
   ws_digitalio_SampleMode _sample_mode;
   bool _value;
   bool _prv_value;
+  bool _first_read; ///< True until the first CheckEvent() read seeds _prv_value
```
brentru: Why do we need a separate., unset?, value here instead of using `value` or `prv_value`

src/components/expander/controller.cpp
```diff
@@ -130,6 +132,33 @@ ExpanderHardware *ExpanderController::GetDriver(uint8_t addr) {
...
+ * @param pin_num      Out: the pin number — for expander pins this is the
+ *                     pin (channel) index on that expander.
+ * @param expander_drv Out: the owning expander driver, or nullptr for a
+ *                     native MCU pin.
+ * @return True if resolved, False if the pin name is malformed or no
+ *         expander is registered at the referenced address.
+ */
+bool ExpanderController::ResolvePinName(const char *pin_name, uint8_t &pin_num,
+                                        ExpanderHardware **expander_drv) {
+  *expander_drv = nullptr;
```
brentru: Why?

src/components/expander/controller.cpp
```diff
@@ -130,6 +132,33 @@ ExpanderHardware *ExpanderController::GetDriver(uint8_t addr) {
...
+  if (!ExpanderHardware::ParsePinNum(pin_name, pin_num))
```
brentru: Why `::ParsePinNum` notation here?

src/components/expander/controller.cpp
```diff
@@ -130,6 +132,33 @@ ExpanderHardware *ExpanderController::GetDriver(uint8_t addr) {
...
+    return false;
+  int32_t addr = WsPinNameToExpanderAddr(pin_name);
+  if (addr == WS_PIN_INVALID)
+    return true; // Native MCU pin, no expander driver
```
brentru: ..shouldn't it fail, then?

src/components/i2c/drivers/drvBase.h
```diff
@@ -88,28 +95,30 @@ class drvBase {
...
-                The SCL pin number.
+                The SCL pin name.
       @param    pin_sda
-                The SDA pin number.
+                The SDA pin name.
   */
-  void SetPins(uint32_t pin_scl, uint32_t pin_sda) {
-    _pin_scl = pin_scl;
-    _pin_sda = pin_sda;
+  void SetPins(const char *pin_scl, const char *pin_sda) {
```
brentru: Needs `nullptr` check

brentru: and then, it needs to assert or return..

src/components/i2c/drivers/drvBase.h
```diff
@@ -88,28 +95,30 @@ class drvBase {
...
+    strncpy(_pin_scl, pin_scl, sizeof(_pin_scl) - 1);
```
brentru: strncpy commands will always truncate. you should validate the size (`sizeof()`) of _pin_scl and pin_scl beofre strncpy commands

src/components/i2c/controller.cpp
```diff
@@ -1034,22 +1043,32 @@ TwoWire *I2cController::GetI2cBusByIndex(size_t index) {
...
+              The SDA pin name, e.g. "D21", "SDA".
     @returns  Pointer to the I2cHardware bus, or nullptr if initialization
               failed.
 */
-I2cHardware *I2cController::findOrCreateBus(uint32_t pin_scl,
-                                            uint32_t pin_sda) {
+I2cHardware *I2cController::findOrCreateBus(const char *pin_scl,
+                                            const char *pin_sda) {
+  // Resolve the pin names to pin numbers
+  int32_t scl_num = WsPinNameToNum(pin_scl);
```
brentru: If we are going to be: 
1) Parsing out the string pin
2) Storing the string pin
2) Resolving the string whenever we use it

Why don't we just have a `struct` to store the pins..?

brentru (changes requested):
Not fully done reviewing, these are the glaring issues and discussion points so far that we should discuss.

src/components/digitalIO/hardware.cpp
tyeth: We cannot infer previous value safely at boot, so depending on logic level (and/or inversion) we assume no event at wake up potentially incorrectly. 
 This sends the initial value at boot no matter what and updates prv_value accordingly to current state.

In an ideal world we would pass the previous logic level (stored in IO feed) through the protobuf at the time of digitialioAdd (see https://github.com/adafruit/Wippersnapper_Protobuf/issues/213) so the prv_value can be initialised correctly and we avoid unnecessary events being published.

src/components/i2c/controller.cpp
tyeth-ai-assisted: Done in #959: added a `WsPinName` struct (`ws_helper_pins.h`) pairing the name string with its resolved number — validated + resolved once via `Set()`, then both travel together. Adopted by `drvBase`, `UARTHardware`, `drvUartBase` and this bus lookup, so nothing re-resolves strings at point of use anymore.

src/components/i2c/drivers/drvBase.h
tyeth-ai-assisted: Done in #959: `WsPinName::Set()` checks `strlen(pin_name) >= sizeof(name)` and fails instead of truncating — `strncpy` only runs once the name is known to fit.

src/components/i2c/drivers/drvBase.h
tyeth-ai-assisted: Done in #959: `Set()` rejects `nullptr`, and `SetPins()` now returns bool — the i2c controller treats false as a failed Add (publishes "Invalid I2C bus pin name(s)!" and deletes the driver), so it returns rather than continuing with bad state.

src/components/expander/controller.cpp
tyeth-ai-assisted: Restructured in #959 with a comment explaining it: the out-param must be deterministic for the native-pin path — callers pass it straight into `GetPin(pin_num, expander_drv)` where `nullptr` means "native pin", so it's initialized before any early return.

src/components/expander/controller.cpp
tyeth-ai-assisted: Gone in #959 — `ExpanderHardware::ParsePinNum` (a static, hence the `::`) is removed outright; the resolver now uses the `ws_helper_pins.h` functions directly.

src/components/expander/controller.cpp
tyeth-ai-assisted: It does fail for anything actually wrong — malformed name, out-of-range pin, or an `EXP_` name whose expander isn't registered. A native pin name resolving successfully with `expander_drv` left `nullptr` is the deliberate success path: digitalio/analogio/pwm make this one call for every pin they handle, native or expander-backed. #959 restructures the body into explicit expander-vs-native paths and documents exactly what fails.

src/components/digitalIO/hardware.cpp
tyeth-ai-assisted: That's the `is_inverted` (active-low) support from the digitalio proto: IO always speaks logical values, so an inverted pin must drive the opposite electrical level. In #959 the conversion is a named, documented `GetElectricalLevel()` helper and the ternaries are gone.

src/components/digitalIO/hardware.cpp
tyeth-ai-assisted: Done in #959 — all ternary ops in hardware.cpp replaced with if/else.

src/components/digitalIO/hardware.cpp
tyeth-ai-assisted: It fixes a bug found during HIL bring-up (T-Dongle C5): `_prv_value` starts as the Add's `initial_value` guess, and comparing the first real read against that guess either suppressed the pin's initial report entirely (guess happened to match, IO never learns the pin state) or fabricated a transition that never happened. First read now seeds the baseline and reports the initial state. #959 expands the code comment with this rationale. Related v2.1 follow-up: seeding from a broker-supplied last-known state instead (adafruit/Wippersnapper_Protobuf#213 / #955).

src/components/digitalIO/hardware.h
tyeth-ai-assisted: Because `_value`/`_prv_value` are bools and can't encode "not yet read" — before the first read they only hold the Add's `initial_value` guess, which must not be compared against as if it were an observed state. The flag is the tri-state marker. #959 expands the doc comment; adafruit/Wippersnapper_Protobuf#213 tracks replacing the guess with a broker-supplied last-known state in v2.1.

brentru (approved):
@tyeth  Robot addressed a lot of my comments which is cool. I would like you to take a _manual_ pass/review over `src/helpers/ws_helper_pins.h` to see if it can be cleaned up, simplified, refactored, or if there are any residual bugs I may have missed during review.

src/helpers/ws_helper_pins.h
```diff
@@ -0,0 +1,130 @@
...
+    return (int32_t)PIN_WIRE1_SDA;
+  if (strcmp(pin_name, "SCL1") == 0)
+    return (int32_t)PIN_WIRE1_SCL;
+#elif defined(SDA1) && defined(SCL1)
+  if (strcmp(pin_name, "SDA1") == 0)
+    return (int32_t)SDA1;
+  if (strcmp(pin_name, "SCL1") == 0)
+    return (int32_t)SCL1;
+#endif
+  // Skip a non-numeric prefix, e.g. "D22", "A4", "GPIO22"
```
tyeth, line 71: This allows us to move away from A/D22 to GPIO22 in board defs and still function as expected. That wouldn't be backwards compatible with v1.

src/helpers/ws_helper_pins.h
```diff
@@ -0,0 +1,130 @@
...
+               truncation, and resolves to a pin number; False otherwise
+               (fields are reset to empty / WS_PIN_INVALID).
+  */
+  bool Set(const char *pin_name) {
+    name[0] = '\0';
+    num = WS_PIN_INVALID;
+    if (pin_name == nullptr)
+      return false;
+    if (strlen(pin_name) >= sizeof(name))
+      return false; // would truncate
```
tyeth, line 106: Double check a truncation would be bubbled up to error msg proto, or be semi quick to establish the issue as no custom err msg defining failure reason.

src/nanopb/ws_pb_helpers.cpp
```diff
@@ -58,4 +58,56 @@ bool ws_pb_encode(pb_ostream_t *stream, const pb_msgdesc_t *fields,
...
+      if (!pb_close_string_substream(stream, &substream) || !ok) {
+        if (!ok) {
+          WS_DEBUG_PRINT("Protobuf decode error: ");
+          WS_DEBUG_PRINTLNVAR(PB_GET_ERROR(&substream));
+        }
+        return false;
+      }
+      decoded = true;
+    } else if (!pb_skip_field(stream, wire_type)) {
+      return false;
```
tyeth, line 109: be nice if we reported both return false to ws_debug or err channel

tyeth (commented):
k, two QoL things to verify, but LGTM

src/nanopb/ws_pb_helpers.cpp
brentru: I'm not sure we want to report when the protobuf decode fails to IO through the `error.proto` API - any thoughts @lorennorman ?

src/helpers/ws_helper_pins.h
```diff
@@ -0,0 +1,130 @@
...
+#if defined(PIN_WIRE1_SDA) && defined(PIN_WIRE1_SCL)
+  if (strcmp(pin_name, "SDA1") == 0)
+    return (int32_t)PIN_WIRE1_SDA;
+  if (strcmp(pin_name, "SCL1") == 0)
+    return (int32_t)PIN_WIRE1_SCL;
+#elif defined(SDA1) && defined(SCL1)
+  if (strcmp(pin_name, "SDA1") == 0)
+    return (int32_t)SDA1;
+  if (strcmp(pin_name, "SCL1") == 0)
+    return (int32_t)SCL1;
```
brentru: Why would we want to support this style, in Arduino? This is only for CircuitPython or Blinka.

src/helpers/ws_helper_pins.h
brentru: See my comment below please.

src/helpers/ws_helper_pins.h
```diff
@@ -0,0 +1,130 @@
...
+
+  /*!
+      @brief  Validates a pin name and resolves it to a pin number.
+      @param  pin_name
+              The pin name string to store and resolve.
+      @returns True if pin_name is non-null, fits the buffer without
+               truncation, and resolves to a pin number; False otherwise
+               (fields are reset to empty / WS_PIN_INVALID).
+  */
+  bool Set(const char *pin_name) {
```
brentru, line 100: How does this differ from `WsPinNameToNum`? Why do we need a separate function here?

brentru (changes requested):

src/nanopb/ws_pb_helpers.cpp
tyeth: @lorennorman ping, we are wondering about error handling, and specifically here after failed proto decodes.

src/helpers/ws_helper_pins.h
tyeth: Mostly I've been using this for unsupported boards, and ensuring the same pin conventions work on both Py + Ardu. Could drop it, no major loss/win either way.

src/helpers/ws_helper_pins.h
tyeth: This allows a boolean check if a pin is valid, effectively a shortcut of ` return WsPinNameToNum(name) != WS_PIN_INVALID`

src/nanopb/ws_pb_helpers.cpp
lorennorman: for posterity: we talked this out in Slack and decided:
- yes, we want to capture errors as often as possible, at least to begin with
- for this one we need to be careful because a proto error could also mean we'll fail to send an error proto
- we won't address the error sending in this PR, but rather start a new one that can focus on errors

Issue #332, opened 2022-09-22
eps8266


bakty04:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Issue #215, opened 2022-02-01
Creating an SCD30 sensor with "Send Every" interval of < 5 minutes causes SCD30 to not initialize


brentru:
**Describe the bug**
Creating an SCD30 component with a "send every" interval of "every 1 minute", "every 30 seconds" causes the device to not receive the I2C Device Init Request and initialize the SCD30.
![IO_-_WipperSnapper_🔊](https://user-images.githubusercontent.com/322428/152055329-61353010-c2ce-4be7-9202-ccb17f47cd67.png)

This does not occur when < 3 sensors on this component are initialized OR if the interval is greater than every 1 minute. I.e.:
![IO_-_WipperSnapper_1](https://user-images.githubusercontent.com/322428/152055388-15a5773c-635f-40f7-b723-024cd394390c.png)

***Arduino board***
Adafruit Feather ESP32-S2 TFT

***WipperSnapper Version***
v1.0.0-beta.23

**Expected behavior**
I2C topic callback should be executed, the I2C Device Init Request decoded, and the SCD30 initialized.

**Which components are connected to your device**
Adafruit SCD30

**Additional context**
Log Output,

BAD SCD30 Init, 1 minute period
```
6:20:07.208 -> Packet Length:	127
16:20:07.208 -> Read data:		  [0x00], T [0x54], b [0x62], r [0x72], e [0x65], n [0x6E], t [0x74], r [0x72], 
16:20:07.208 -> 	u [0x75], b [0x62], e [0x65], l [0x6C], l [0x6C], / [0x2F], w [0x77], p [0x70], 
16:20:07.208 -> 	r [0x72], s [0x73], n [0x6E], p [0x70], r [0x72], / [0x2F], i [0x69], o [0x6F], 
16:20:07.208 -> 	- [0x2D], w [0x77], i [0x69], p [0x70], p [0x70], e [0x65], r [0x72], - [0x2D], 
16:20:07.208 -> 	a [0x61], d [0x64], a [0x61], f [0x66], r [0x72], u [0x75], i [0x69], t [0x74], 
16:20:07.208 -> 	- [0x2D], f [0x66], e [0x65], a [0x61], t [0x74], h [0x68], e [0x65], r [0x72], 
16:20:07.208 -> 	- [0x2D], e [0x65], s [0x73], p [0x70], 3 [0x33], 2 [0x32], s [0x73], 2 [0x32], 
[...]

Issue #204, opened 2022-01-07
Beta.21 continuous reboot on SAMD51


brentru:
SAMD51 platforms continuously reboot while they attempt to connect to WiFi.

brentru:
Fixed in #205

PR #946, opened 2026-07-10
feat(telemetry): RSSI + boot reason telemetry components

tyeth:
## Summary

Adds a new **telemetry** component to the API v2 firmware, implementing the telemetry.proto API. Each telemetry metric is its own component instance that reports a device/host reading up to the broker as a `ws.telemetry.Event`.

This PR implements the two metrics requested — **RSSI** and **boot reason** — and is aligned with the metric contract in [Wippersnapper_Components#332](https://github.com/adafruit/Wippersnapper_Components/pull/332).

> **Draft** for @tyeth to review.

## Metric identity is a `ws.telemetry.Type` enum

Per review feedback, a telemetry metric is identified on the wire by the **`ws.telemetry.Type` enum**, not a string name — so the metric is defined by the component library, filled by the broker, and mapped straight to a reader on the device with no name→kind resolution or per-instance name storage.

The enum ([Wippersnapper_Protobuf#215](https://github.com/adafruit/Wippersnapper_Protobuf/pull/215)) is banded by category so it can grow without renumbering:

```proto
enum Type {
  TM_UNSPECIFIED = 0;
  TM_BOOT_REASON = 1;   // boot / reset (1-9)
  TM_BOOT_COUNT  = 2;
  TM_RSSI        = 10;  // connectivity / latency (10-19)
  TM_LATENCY_MS  = 11;
  // memory / resource metrics begin at 20
}
```

## Dependencies

- **Protobuf** — [#204](https://github.com/adafruit/Wippersnapper_Protobuf/pull/204) *(merged into `api-v2`)* added `telemetry.proto` (B2D `Add`/`Remove`, D2B `Event`), the signal B2D tag 41 / D2B tag 22, and checkin `ComponentAdds.telemetry_adds` tag 11. [#215](https://github.com/adafruit/Wippersnapper_Protobuf/pull/215) then replaced the string `name` field with the `Type` enum above.
- **Components** — [#332](https://github.com/adafruit/Wippersnapper_Components/pull/332) defines the metric contract (valueKind → `ws.sensor.Event`: float → `T_RAW`, bytes → `T_BYTES`).
- ⚠️ The generated wrappers in `src/protos/telemetry.pb.{c,h}` were produced from the adafruit/Wippersnapper_Protobuf#215 proto (via the Protobuf repo's nanopb generator). **Once it merges, regenerate them from the merged CI output.**

## Metrics

| Metric | `Type` | Value | sensor.Event | Cadence |
|---|---|---|---|---|
| Wi-Fi RSSI | `TM_RSSI` | signal strength (dBm) | `float_value` / `T_RAW` | periodic (default 300s) |
| Boot reason | `TM_BOOT_REASON` | reset/boot reason string | `bytes_value` / `T_BYTES` | report-once (period 0) |

`TM_BOOT_COUNT` and `TM_LATENCY_MS` are recognized but **not yet sourced** by this firmware — `TelemetryTypeIsSupported()` returns false, so they're rejected with a component error without aborting checkin (forward-compat).

## Boot reason (per-CPU-core, multi-arch)

`boot_reason` reports a best-effort reset reason across all supported architectures. On multi-core ESP32 it reports **one CPU core per line**:

```
CPU0: SW_CPU_RESET
CPU1: EXT_CPU_RESET
```

| Arch | Source |
|---|---|
| ESP32 (all variants) | `rtc_get_reset_reason(core)` per core, looped over `ESP.getChipCores()` |
| RP2040 / RP2350 | `rp2040.getResetReason()` (chip-level; Earle Philhower core) |
| SAMD51 | `RSTC->RCAUSE` reset cause register |
| ESP8266 | `ESP.getResetReason()` |

## Implementation

`src/components/telemetry/{controller,model,hardware}.{h,cpp}`, mirroring the existing component pattern:

- **`TelemetryHardware`** is constructed directly from a `ws_telemetry_Type` (no name/kind resolution). A short name for logs / offline SD entries / component errors is *derived* from the type via `TelemetryTypeName()`.
- **`TelemetryController`** — `Handle_TelemetryAdd` rejects unsupported types *before* allocating, dedups an existing metric via `updatePeriod(type, period)`, and null/identity-checks the new instance before tracking it; `Handle_TelemetryRemove` matches by type. `Publish(type, value_type, value)` fills the model, then encodes and publishes.
- Routed through `routeBrokerToDevice` (B2D tag 41) and `PublishD2b` (D2B tag 22); added at checkin via `cbTelemetryAdds`.
- Polled from `loop()` and the sleep-mode `loopSleep()` readiness loop; participates in `ResetAllControllerFlags()`.
- **Offline/SD mode**: each reading is logged to the SD card via `ws_sdcard::LogTelemetryEventToSD()` instead of publishing.

## Review feedback addressed

- Metric identity is now the `Type` enum, not a string (`Kind` → `Type`, no device-side resolution/struct).
- `updatePeriod()` extracted; `PublishTelemetry()` → `Publish(type, value_type, value)` that fills the model.
- Unsupported types checked before `new`; instance null/identity-checked before `push_back`.
- `ESP.getChipCores()` typed as `uint8_t`; boot-reason line/buffer sizes matched to worst case.
- `if`/`else` instead of ternaries throughout.
- Offline mode logs telemetry readings to SD (no longer skipped).

## Build / verification

The CI matrix builds across all targets — ESP32 variants, ESP8266, SAMD, RP2040/RP2350, and Clang & Doxygen — with one exception introduced by the added component: the `esp32c6_debug` (verbose, `min_spiffs`) target overflowed its 1.9 MB app partition. Resolved by enlarging that debug target's app partition in `ci-arduino@ci-wippersnapper`. Locally, `adafruit_feather_esp32s3` builds SUCCESS.


src/components/telemetry/controller.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+  // Telemetry is a network concept - if we're offline there's nowhere to
+  // report to, so mark everything complete so sleep-mode readiness isn't
+  // blocked and bail out.
+  if (Ws._sdCardV2->isModeOffline()) {
+    for (size_t i = 0; i < _telemetry_metrics.size(); i++) {
+      _telemetry_metrics[i]->did_read_send = true;
+    }
+    return;
+  }
```
tyeth: Might want to log to sd here

src/components/telemetry/controller.cpp
```diff
@@ -0,0 +1,240 @@
...
+    case TELEMETRY_KIND_RSSI:
+      _telemetry_model->SetValueFloat(ws_sensor_Type_T_RAW,
+                                      (float)metric.ReadRSSI());
+      break;
+    case TELEMETRY_KIND_BOOT_REASON:
+      _telemetry_model->SetValueString(ws_sensor_Type_T_BYTES,
+                                       metric.ReadBootReason());
+      break;
+    default:
+      // Shouldn't happen - unknown metrics are never registered
```
tyeth: log here, if not already logged during TelemetryAdd

src/components/telemetry/controller.cpp
tyeth-ai-assisted: Done in 847d7bd9 — the offline branch no longer skips telemetry. `update()` now logs each reading to the SD card in offline mode via new `ws_sdcard::LogTelemetryEventToSD()` overloads (float for `rssi`, string for `boot_reason`), mirroring `LogDS18xSensorEventToSD`/`LogI2cDeviceEvent` (timestamp + `telemetry` name + value + si_unit). Report-once/period gating is preserved. Let me know if you'd prefer a different JSON shape for the offline log entries.

src/components/telemetry/controller.cpp
tyeth-ai-assisted: Added in 847d7bd9 — the `default:` case now logs `[telemetry] WARNING: No reader for metric: <name>`. This shouldn't be reachable (unknown metric names are rejected in `Handle_TelemetryAdd` with a component error), but it's logged here as a safety net in case one ever slips through.

tyeth:

src/components/telemetry/controller.cpp
```diff
@@ -0,0 +1,260 @@
...
+  // If a metric with this name already exists, update its period instead of
+  // registering a duplicate.
+  for (size_t i = 0; i < _telemetry_metrics.size(); i++) {
+    if (strcmp(_telemetry_metrics[i]->GetName(), msg->name) == 0) {
+      _telemetry_metrics[i]->SetPeriod(msg->period);
+      WS_DEBUG_PRINTLN("[telemetry] Updated existing metric's period");
+      return true;
+    }
+  }
+
```
brentru: Refactor this out into an `updatePeriod()` call that takes in the component name

src/components/telemetry/controller.cpp
```diff
@@ -0,0 +1,260 @@
...
+  TelemetryHardware *new_metric = new TelemetryHardware(msg->name, msg->period);
+
+  // Gracefully ignore metrics this firmware doesn't know how to source, so a
+  // single unsupported metric doesn't abort the rest of the checkin.
+  if (new_metric->GetKind() == TELEMETRY_KIND_UNKNOWN) {
```
brentru: Would prefer if `Kind`->`Type` to standardize across APIs

brentru: Could this check execute prior to the `new` creation?

src/components/telemetry/controller.cpp
```diff
@@ -0,0 +1,260 @@
...
+    WS_DEBUG_PRINTLN("[telemetry] WARNING: Unknown telemetry metric, ignoring");
+    Ws.error_handler->publishComponentError(msg->name,
+                                            "Unknown telemetry metric");
+    delete new_metric;
+    return true;
+  }
+
+  _telemetry_metrics.push_back(new_metric);
```
brentru, line 133: Hrm.. before we `push_back` the instance, we should check if:
1) was the object created? check if `new_metric` == NULL
2) did it succeed? check period getter or name getter to ensure a match

src/components/telemetry/controller.cpp
```diff
@@ -0,0 +1,260 @@
...
+                                                     "found");
+  return false;
+}
+
+/*!
+    @brief  Encodes and publishes the telemetry model's current Event to the
+            broker.
+    @return True if the Event was encoded and published, False otherwise.
+*/
+bool TelemetryController::PublishTelemetry() {
```
brentru: `Publish`

src/components/telemetry/hardware.cpp
```diff
@@ -0,0 +1,266 @@
...
+            Philhower core's chip-level reset reason; SAMD51 reads the RSTC
+            reset cause; ESP8266 uses ESP.getResetReason().
+    @returns A short human-readable reset reason. May contain multiple
+             newline-separated lines on multi-core targets.
+*/
+const char *TelemetryHardware::ReadBootReason() {
+  _boot_reason[0] = '\0';
+#if defined(ARDUINO_ARCH_ESP32)
+  // ESP32 exposes a reset reason per CPU core - report one CPU per line.
+  int num_cores = ESP.getChipCores();
```
brentru: This is a `uint8_t`:
https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/Esp.cpp#L338

Please match

brentru: https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/misc_system_api.html#_CPPv4N15esp_chip_info_t5coresE

src/components/telemetry/hardware.h
```diff
@@ -0,0 +1,57 @@
...
+  int32_t ReadRSSI();
+  const char *ReadBootReason();
+  bool did_read_send; ///< True if the last read was sent to IO, False otherwise
+private:
+  char _name[24];        ///< Telemetry component name (e.g. "rssi")
+  TelemetryKind _kind;   ///< Metric kind resolved from _name
+  float _period;         ///< Desired reporting interval, in milliseconds
+  float _prv_period;     ///< Last time the metric was reported, in milliseconds
+  bool _reported_once;   ///< True once a report-once (period 0) metric has sent
+  char _boot_reason[96]; ///< Backing store for the assembled boot reason string
```
brentru: Worst case for this func is `"CPU0: RTCWDT_BROWN_OUT_RESET\nCPU1: RTCWDT_BROWN_OUT_RESET" `, `56` bytes w/terminator. Maybe allocate `64` bytes here instead.

src/components/telemetry/hardware.cpp
```diff
@@ -0,0 +1,266 @@
...
+  if (num_cores < 1)
+    num_cores = 1;
+  for (int core = 0; core < num_cores; core++) {
+    char line[48];
```
brentru: why does `line[48]` not match `_boot_reason[96]`?

src/components/telemetry/hardware.h
```diff
@@ -0,0 +1,57 @@
...
+
+/*!
+    @brief  The metric a telemetry component instance reports. Resolved from
+            the instance's name (e.g. "rssi", "boot_reason").
+*/
+typedef enum {
+  TELEMETRY_KIND_UNKNOWN = 0, ///< Unrecognized telemetry metric name
+  TELEMETRY_KIND_RSSI,        ///< WiFi signal strength (RSSI), in dBm
+  TELEMETRY_KIND_BOOT_REASON, ///< Reason for the last device reset/boot
+} TelemetryKind;
```
brentru: Wouldn't this be best as an enum within `telemetry.proto` so it can be defined by the component, filled by the broker, and require no resolution or extra struct in the hardware?

src/components/telemetry/controller.cpp
```diff
@@ -0,0 +1,260 @@
...
+            If true, forces a report on all metrics regardless of period.
+*/
+void TelemetryController::update(bool force) {
+  // Bail out if there are no metrics to report
+  if (_telemetry_metrics.empty())
+    return;
+
+  // When online we publish to the broker; when running offline we log the
+  // reading to the SD card instead.
+  bool is_offline = Ws._sdCardV2->isModeOffline();
```
brentru: We only reference `is_offline` 2x in the loop, can just do the sd card call instead.

src/components/telemetry/controller.cpp
```diff
@@ -0,0 +1,260 @@
...
+    switch (metric.GetKind()) {
+    case TELEMETRY_KIND_RSSI: {
+      float value = (float)metric.ReadRSSI();
+      if (is_offline) {
+        delivered = Ws._sdCardV2->LogTelemetryEventToSD(metric.GetName(), value,
+                                                        ws_sensor_Type_T_RAW);
+      } else {
+        _telemetry_model->InitEventMsg(metric.GetName());
+        _telemetry_model->SetValueFloat(ws_sensor_Type_T_RAW, value);
+        delivered = PublishTelemetry();
```
brentru: `PublishTelemetry` should fill the model (what youa re doing above) so `delivered = PublishTelemetry(name, value type, value)`

brentru (changes requested):
@tyeth  Requesting changes

tyeth-ai-assisted:
@tyeth — pushed the enum refactor and the follow-up review fixes. CI is green across the full matrix. ✅

**Latest changes (`0d080bcd`, `317baa2e`):**
- Telemetry now keys on the **`ws.telemetry.Type` enum** ([Protobuf#215](https://github.com/adafruit/Wippersnapper_Protobuf/pull/215)) instead of a string name — no name→kind resolution, no per-instance name storage. The human-readable name for logs / SD entries / component errors is derived from the type via `TelemetryTypeName()`.
- Regenerated `src/protos/telemetry.pb.{c,h}` from the adafruit/Wippersnapper_Protobuf#215 proto.
- Addressed @brentru's review comments: `Kind`→`Type`; `updatePeriod()` extracted; `PublishTelemetry()`→`Publish(type, value_type, value)` that fills the model; unsupported types rejected before `new`; instance null/identity-checked before `push_back`; `ESP.getChipCores()` typed `uint8_t`; boot-reason line/buffer sizes matched to worst case; ternaries replaced with `if`/`else`.
- `TM_BOOT_COUNT` / `TM_LATENCY_MS` are recognized but not yet sourced here (rejected via `TelemetryTypeIsSupported()` without aborting checkin).

**Heads up on two things:**
1. The `wippersnapper_feather_esp32c6_debug` (verbose, `min_spiffs`) target overflowed its 1.9 MB app partition once the component was added — fixed by your partition bump to `huge_app` for only the debug build target in `ci-arduino@ci-wippersnapper`, which the re-run confirmed green.
2. `src/protos/telemetry.pb.{c,h}` were generated from the adafruit/Wippersnapper_Protobuf#215 branch — **once it merges, they should be regenerated from the merged CI output**.

The PR description is updated to match the enum design.

tyeth (commented):
LGTM, all resolved.
Needs https://github.com/adafruit/Wippersnapper_Protobuf/pull/215 merging as existing telemetry proto is merged to wippersnapper_protobuf:api-v2

brentru:
@tyeth  Has this been tested on hardware against either IO or ProtoMQ (with test scripts)?

tyeth-ai-assisted:
HIL update — **RSSI telemetry proven end-to-end** on a QT Py ESP32-S3 over protomq (proto ref `telemetry-enum-type`): `TM_RSSI` Add → device created the metric → periodic `T_RAW`/`float_value` D2B every 2s, decoded cleanly by the broker.

One caveat surfaced: the reported value is **0 dBm** because `wippersnapper::getRSSI()` on the base global `Ws` is an unimplemented stub (the real `getRSSI()→WiFi.RSSI()` lives on the `ws_adapter_wifi` object, unreachable from components) — a **pre-existing v2 gap** that also affects the display's RSSI, not a telemetry-code issue. Tracked in adafruit/Adafruit_Wippersnapper_Arduino#960, to be fixed in a **separate PR stacked on top of this one** (it touches base networking). The `TM_RSSI` component here is correct as-is; only the value source needs wiring.

(boot_reason was already proven on hardware: `CPU0: POWERON_RESET\nCPU1: POWERON_RESET`.)

tyeth-ai-assisted:
Correction on tracking: the RSSI value=0 gap is the **existing** "two WS globals" issue adafruit/Adafruit_Wippersnapper_Arduino#830 (fix drafted in adafruit/Adafruit_Wippersnapper_Arduino#829 — switch the global to the `Wippersnapper_WiFi` type), not a new bug. I closed my duplicate #960 and moved the RSSI/ESP32-S3 HIL evidence onto #829. #829 landing (or a rebase of it on top of this) will fix the RSSI telemetry value + display RSSI.

tyeth:
@brentru yes tested on hardware, boot reasons passed up for an S3, will chuck more boards at it once home but previously checked compilation and flashing worked.

brentru:
@tyeth  Ok - tag me here when it is fully tested and ready for review

tyeth (commented):
Retested with latest telemetry proto, confirmed report_once based Adds can be resent and the expected response is received (e.g. ask for boot reason a second time without removing the existing boot reason telemetry component first - effectively a period update and resend).

tyeth:
@brentru ready for re-review after being fully tested boot reason and RSSI resolved via the linking fix etc to finish testing.

I don't know how we're doing boot count yet, presumably nvm tied into sleep mode (vague memory of it existing there already), so that'll be a new PR.

After this it seems worth immediately adding the python latency (instead of rssi or optionally both - I'd naively assumed ping only for SBC), and probably things like free space / mem.

brentru:
> I don't know how we're doing boot count yet, presumably nvm tied into sleep mode (vague memory of it existing there already), so that'll be a new PR.


Let's remove it for now from the PB api then, I made a comment over there

src/components/telemetry/controller.h
```diff
@@ -0,0 +1,52 @@
...
+  void ResetFlags();
+
+private:
+  bool updatePeriod(ws_telemetry_Type type, float period);
+  bool encodeAndPublish();
+  bool Publish(ws_telemetry_Type type, ws_sensor_Type value_type, float value);
+  bool Publish(ws_telemetry_Type type, ws_sensor_Type value_type,
+               const char *value);
+  TelemetryModel *_telemetry_model; ///< telemetry model
+  std::vector<TelemetryHardware *> _telemetry_metrics;
```
brentru, line 49: I feel the `vector` is useful for I2c where we have a lot of different types, but the telemetry API only has RSSI and Boot Reason. We'll probably increase the # of types down the line, but consider switching to a fixed array with a `#define`'d max count.

src/components/telemetry/hardware.h
```diff
@@ -0,0 +1,66 @@
...
+  const char *ReadBootReason();
+  bool did_read_send; ///< True if the last read was sent to IO, False otherwise
+private:
+  ws_telemetry_Type _type; ///< The metric this instance reports
+  float _period;           ///< Desired reporting interval, in milliseconds
+  float _prv_period;       ///< Last time the metric was reported, in ms
+  bool _report_once;       ///< True if the metric should be reported only once
+  bool _has_reported_once; ///< True once a report-once (period 0) metric has
+                           ///< sent
+  char _boot_reason[64]; ///< Backing store for the assembled boot reason string
```
brentru: There may be more boot reasons as espressif, rpi and other mcu vendors add chips - switch `64` to a `#define`

src/components/telemetry/model.h
```diff
@@ -0,0 +1,48 @@
...
+  bool EncodeEvent();
+  ws_telemetry_D2B *GetD2B();
+
+private:
+  static bool encode_string_cb(pb_ostream_t *stream, const pb_field_t *field,
+                               void *const *arg);
+  ws_telemetry_Add _msg_Add;       ///< Add message
+  ws_telemetry_Remove _msg_Remove; ///< Remove message
+  ws_telemetry_D2B _msg_D2B;       ///< D2B wrapper holding the Event
+  char _str_value[96]; ///< Backing store for a string value (bytes_value)
```
brentru: Why is this 96 bytes when the largest string you're holding within `controller` is 64 bytes? Where did this value come from?

brentru (changes requested):
@tyeth  Looking better! Why is this not merging into `migrate-api-v2` though?

src/components/telemetry/model.h
tyeth: There are two boot reasons (48 char max) if it's a dual cpu. I went with 64 for above so we can keep a larger reason for a single cpu if one ever is added.

brentru:
I'd also like screenshots of IO with each component (rssi, boot reason) on a device before we merge this in.

tyeth:
> I'd also like screenshots of IO with each component (rssi, boot reason) on a device before we merge this in.

Yep cool, currently **blocked** by lack of v2 component import which @lorennorman is working on this week.

tyeth:
@brent ~~worth review of the new guards, think that satisfies the comments.~~
Won't merge before fixing base and seeing online.

@tyeth make sure to rebase https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/tree/claude/telemetry-protupdate 
from migrate-api-v2-backport-components to migrate-api-v2 (dropping 15 sensor backport/fix commits).

src/components/telemetry/controller.h
tyeth: Good point, mentally filed, for now leaving as is.

brentru (approved):
@tyeth  Approved and understood that you're blocked on the component importer

tyeth-ai-assisted:
Superseded by adafruit/Adafruit_Wippersnapper_Arduino#978 (merged into `migrate-api-v2` at 0ab77503).

The telemetry component was ported there directly: `src/components/telemetry/` (verified zero-diff against this branch), checkin `cbTelemetryAdds`, all `wippersnapper.{h,cpp}` wiring including the `PublishD2b` D2B telemetry case, `ws_sdcard::LogTelemetryEventToSD`, and the telemetry proto wrappers (byte-identical to the Wippersnapper_Protobuf api-v2 CI output, which #978 adopted wholesale).

Everything else on this branch is an older snapshot of work that lives in current form elsewhere: the driver backports + HIL scripts/workflows are on #933's branch, and the Ws-pointer refactor (#974) landed via #978. Nothing unique remains — recommend closing this PR.

tyeth-ai-assisted:
Correction to the above: please keep this PR **open** for now — @tyeth wants to prove the ported telemetry end-to-end against Adafruit IO before closing (the staging broker isn't v2-ready yet, so online verification is pending). Close only after that proof.

Issue #952, opened 2026-07-17
v2: refcounted shared power-rail/pin ownership + sleep-mode rail handling


tyeth-ai-assisted:
### Context

adafruit/Adafruit_Wippersnapper_Arduino#936 added a panel **power-enable rail** to the display Add (`ws.display.Add.power`, a prerequisite `ws.digitalio.Add` — see adafruit/Wippersnapper_Protobuf#208). The PR implements deliberately **two-party** arbitration, per review:

- pin free at Add → the display driver *owns* the rail (enable level before init, released to disable level in its destructor, i.e. display Remove/replace);
- pin already registered with the digitalio controller → **pass through** only if it's an `OUTPUT` already at the enable level, else the Add fails ("Power-enable pin unavailable").

That's correct for the direct display↔digitalio conflict, but it does not generalize. This issue captures the gaps so the design discussion has a home.

### Failure scenarios (current behavior, verified against the code)

Assume one rail (e.g. GPIO15) powers the display **and** I²C sensors **and** NeoPixels, with components added/removed in arbitrary order at check-in or runtime:

| Order of events | What happens today |
|---|---|
| digitalio Add (OUTPUT@enable) → display Add | ✅ pass-through works |
| display Add → **digitalio Add, same pin** | ❌ digitalio never queries display ownership — it reconfigures the pin per its own Add (possibly INPUT or LOW) and kills the panel |
| display owns rail → **display Remove** | ❌ driver dtor drives the rail to disable — I²C sensors / NeoPixels on the same rail are power-cut silently |
| passed-through digitalio pin → **digitalio Remove** | ❌ `DigitalIOHardware` dtor does `LOW + pinMode(INPUT)` — the rail floats and the panel dies with no notification path |
| two displays sharing one rail | ❌ ownership is private to each driver instance; the second display sees the pin as "free", both own it, first Remove powers down the second |
| I²C bus init when `PIN_I2C_POWER` == the rail | ❌ `i2c/hardware.cpp` drives the macro pin directly — it even flips it through `INPUT` to sniff rest polarity (momentary float/brownout), invisible to any check |
| NeoPixel power pin | v2 `pixels` doesn't model a power pin at all |

Root cause: **ownership is pairwise, one-directional, and implicit** — there is no shared registry, no refcount, and time-of-Add-only checking.

### Deep sleep / light sleep (v2 sleep component)

[...]

PR #953, opened 2026-07-17
api-v2: pin-name review fixes + expander pin resolution utility (#943 follow-up)

tyeth:
## Summary

Follow-up to #943, addressing the review comments there, plus the expander pin-resolution utility discussed in the `ws_helper_pins.h` thread (refs #936's `QueryPinState` pattern and the shared-pin issue #952).

## Review comments addressed

- **`WsPinNameToNum` signature** (@brentru): now `int32_t WsPinNameToNum(const char *pin_name)` returning `WS_PIN_INVALID` (-1) on failure — no more out-param that could be left uninitialized. All i2c/uart call sites updated to assign-and-validate.
- **`EXP_` check at the front** (@brentru): `"EXP_0xNN_P"` resolves to `P`, the pin (**channel**) index on the expander at address `0xNN` — matching the 16-channel solenoid-driver paradigm where 0–7 are the A-channel transistor-backed GPIO and 8–15 the spare B-channel GPIO. A companion `WsPinNameToExpanderAddr()` returns the expander's I2C address (also serving as the "is this an expander pin" test).
- **Pass off to the expander controller** (@tyeth, analogous to `QueryPinState` from #936): new `ExpanderController::ResolvePinName(pin_name, pin_num, &expander_drv)` resolves any pin name to its number and, for expander names, the owning `ExpanderHardware*` (nullptr for native pins). The eight duplicated `ParsePinNum` + `strtoul` + `GetDriver` blocks in digitalio/analogio/pwm now route through it, and `ExpanderHardware::ParsePinNum` delegates to `WsPinNameToNum` — malformed expander names now fail instead of silently parsing as pin 0.
- **Buffer sizes `#define`'d to exactly match the pb** (@brentru): `drvBase.h` defines `DRV_BASE_NAME_LEN` / `DRV_BASE_PIN_NAME_LEN` as `sizeof()` of the `ws_i2c_Add.name` / `ws_i2c_AddressSpace.pin_scl` pb fields (fixes `_name[15]` vs the pb's `name[16]`); `drvUartBase.h` likewise sizes `_name` / `_port_name` from `ws_uart_Add.name` / `ws_uart_Descriptor.pin_rx`. Tied to the pb via `sizeof`, they can't drift when the protos are regenerated.
- **SDA1/SCL1** (@tyeth): added the `#elif defined(SDA1) && defined(SCL1)` fallback for cores that define the bare macros rather than `PIN_WIRE1_*`.

## Expander pin paradigm (why the channel index)

For an `EXP_` pin the "pin number" is the channel index scoped to that expander, exactly as the existing digitalio/analogio/pwm expander support treats it — `(expander_drv, pin_num)` pairs are the identity everywhere. A hypothetical board that routes header pins through an onboard expander (e.g. an Arduino Nesso-style port) would surface those pins as `EXP_<addr>_<n>` names and resolve through the same two calls: `WsPinNameToNum()` for the channel, `WsPinNameToExpanderAddr()` → `ExpanderController::GetDriver()` (or just `ResolvePinName()`) for the owning driver — no per-component parsing required.

## Deliberately out of scope

- **Shrinking the pb pin fields 32 → 16 chars** (@brentru): needs the field width changed in adafruit/Wippersnapper_Protobuf and a nanopb regen; the `sizeof`-based defines here will pick it up automatically when that lands.
- **Error-handler / signal-encoder fixes in #943** (@brentru asked for a separate PR against `migrate-api-v2`): untouched here.
- Refcounted shared pin/rail ownership remains tracked in #952.

## Build verification

CI on this PR; local `adafruit_feather_esp32s3` build was still in flight when pushed (will follow up with the result).

### DIFF
--- a/src/components/analogIn/controller.cpp
+++ b/src/components/analogIn/controller.cpp
@@ -131,8 +131,10 @@ AnalogInHardware *AnalogInController::GetPin(uint8_t pin_num,
 bool AnalogInController::Handle_AnalogInAdd(ws_analogin_Add *msg) {
   WS_DEBUG_PRINTLN("[analogin] Handle_AnalogInAdd MESSAGE...");
   uint8_t pin_num = 0;
-  if (!ExpanderHardware::ParsePinNum(msg->pin_name, pin_num)) {
-    WS_DEBUG_PRINTLN("[analogin] ERROR: Malformed expander pin name!");
+  ExpanderHardware *expander_drv = nullptr;
+  if (!Ws._expander_controller->ResolvePinName(msg->pin_name, pin_num,
+                                               &expander_drv)) {
+    WS_DEBUG_PRINTLN("[analogin] ERROR: Unable to resolve pin name!");
     return false;
   }
 
@@ -147,17 +149,6 @@ bool AnalogInController::Handle_AnalogInAdd(ws_analogin_Add *msg) {
       msg->sample_mode != ws_analogin_SampleMode_SM_EVENT) {
     WS_DEBUG_PRINTLN("[analogin] ERROR: Invalid sample mode in message!");
     return false;
-  }
-
-  // Resolve the expander driver if this is an expander pin
-  ExpanderHardware *expander_drv = nullptr;
-  if (strncmp(msg->pin_name, "EXP_", 4) == 0) {
-    uint8_t i2c_addr = (uint8_t)strtoul(msg->pin_name + 4, nullptr, 16);
-    expander_drv = Ws._expander_controller->GetDriver(i2c_addr);
-    if (!expander_drv) {
-      WS_DEBUG_PRINTLN("[analogin] ERROR: Expander not found for address!");
-      return false;
-    }
   }
 
   // If pin is being updated, remove the existing pin first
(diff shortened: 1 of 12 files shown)


tyeth-ai-assisted:
@tyeth ready for your review — all 38 CI checks pass. Addresses the #943 review comments (return-value `WsPinNameToNum`, `EXP_` channel resolution + `WsPinNameToExpanderAddr()`, pb-`sizeof` buffer defines, SDA1/SCL1 fallback) and adds the QueryPinState-style `ExpanderController::ResolvePinName()` pass-off, consolidating the eight duplicated resolve blocks in digitalio/analogio/pwm.

tyeth:
I half expected the WS pin helper to gain an extra arguement that would return the expander driver or hardware as appropriate. Although I'm not entirely sure I like that, because it's then going to require switching code branch everywhere you do some digital IO, so ideally it will be abstracted away to be digital reads or digitalio hardware interfaces, And the hardware drivers themselves can be oblivious to the underlying pin interface

brentru:
@tyeth  This work is becoming difficult for me to follow.

. Could you please clarify, in **simple terms**, what the difference between this Pr and https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/943 is? 

 Also, the PB migration to string pins got merged in [three weeks ago](https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/943) without a matching Arduino PR at the time of merge to resolve the pins.  This meant my feature branch, which works off the api, had a lot of debugging to find this issue. In the future - I'd like to avoid merging in breaking updates to PB branches without a matching Arduino branch.

tyeth:
> @tyeth This work is becoming difficult for me to follow.
> 
> . Could you please clarify, in **simple terms**, what the difference between this Pr and #943 is?

* Refactors the private/public interfaces for getting string pins converted into pin nums or expander pins, for use with displays + i2c + expander pins + digitialio + uart.
* Supports SDA1 better
* Adds sizeof based length defines to match protos automatically
* Returns -1 for pins that fail to resolve

> 
> Also, the PB migration to string pins got merged in [three weeks ago](https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/943) without a matching Arduino PR at the time of merge to resolve the pins. This meant my feature branch, which works off the api, had a lot of debugging to find this issue. In the future - I'd like to avoid merging in breaking updates to PB branches without a matching Arduino branch.

Two/three weeks ago when the proto was merged I had the Arduino PR ready the day before in review, and we had been discussing it already, but the review blocked and python had been broken due to string pins for a long time and was blocking other PRs + work so the proto merge felt urgent (probably wasn't -  although was a stacked PR with many others based on it).

The need in arduino initially looked smaller to get a passing build reserving future expander pin work (although in the end was a larger refactor), so the most basic need was done in #943.

Then subsequently with the i8080 display power pin from adding the T-DisplayS3, and then during string pin testing i2c (mcp9808) uart (pm2.5) and expander pins (4x1 arcade button w/LED stemmaQT seesaw) while adding a new board (T-Dongle C5) that the extra issues were discovered (pin ownership and resolution).

PR #954, opened 2026-07-21
fix: memset nanopb message re-inits instead of init_zero assignment (ESP8266 GCC)

tyeth:
## Summary

Splits the error-handler / signal-encoder fixes out of #943, per review there ("The error handler fixes should be in another PR against `migrate-api-v2`...").

## Changes

Replaces `_d2b_msg = ws_error_D2B_init_zero;` re-assignments in `ErrorHandler` (ctor + the four `publishComponentError` overloads) and the redundant `*msg = ws_signal_DeviceToBroker_init_zero;` after `memset` in `wippersnapper::PublishD2b` with plain `memset`.

**Why:** ESP8266's older GCC rejects braced-list *re-assignment* (as opposed to declaration-initialization) once the `init_zero` initializer contains string-literal char-array members — which happens as soon as the i2c/uart pin fields become `char[]` strings (#943 / adafruit/Wippersnapper_Protobuf#207). `memset` is bit-identical to `init_zero` for these all-static structs, so this is a no-op on `migrate-api-v2` today and unblocks the huzzah8266 build for #943.

The changes are byte-identical to the corresponding lines in #943, so once this merges, #943 rebases/merges over it without conflicts and its diff shrinks accordingly.

### DIFF
--- a/src/components/error/handler.cpp
+++ b/src/components/error/handler.cpp
@@ -38,7 +38,13 @@ static bool encode_string_callback(pb_ostream_t *stream,
 /*!
     @brief  ErrorHandler constructor
 */
-ErrorHandler::ErrorHandler() { memset(&_d2b_msg, 0, sizeof(_d2b_msg)); }
+ErrorHandler::ErrorHandler() {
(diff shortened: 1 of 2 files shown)


brentru (approved):
Resolves issue with ESP8266's GCC

PR #959, opened 2026-07-29
api-v2: address second review round on #943 (WsPinName struct, resolver semantics, digitalio cleanups)

tyeth:
## Summary

Addresses the 2026-07-23 review round on #943. Each inline thread there gets a reply pointing at the change here.

## Changes

- **`WsPinName` struct** (`ws_helper_pins.h`): pairs the pin-name string with its resolved number — validated and resolved once via `Set()`, then both representations travel together instead of re-resolving the string at every use. `Set()` rejects `nullptr` and any name that would truncate (`strlen >= sizeof` fails; `strncpy` can no longer silently truncate). Adopted by:
  - `drvBase` (`_pin_scl`/`_pin_sda`, `static_assert`'d to hold the pb pin-name fields; `SetPins()` returns bool and the i2c controller fails the Add with a component error on invalid names)
  - `UARTHardware` (`_pin_rx`/`_pin_tx`) and `drvUartBase` (`_port`)
  - the i2c bus lookup (`findOrCreateBus`/`GetI2cBus`)
- **`ExpanderController::ResolvePinName` restructured**: explicit expander-vs-native paths. An expander pin name whose expander isn't registered **fails**; a native name succeeds with `expander_drv` left `nullptr` — documented as the point of the API (one call serves both pin domains for digitalio/analogio/pwm). `ExpanderHardware::ParsePinNum` is removed outright, so no more `::` static call.
- **digitalio**: the `is_inverted` (active-low) logical→electrical conversion is now a documented `GetElectricalLevel()` helper; all ternary ops in `hardware.cpp` replaced with if/else. The `_first_read` seeding is kept (it fixes a HIL-verified bug where comparing against the Add's `initial_value` guess either suppressed the pin's initial report or fabricated a transition) with the reasoning now spelled out in the comments — two bools can't encode "not yet read".

## Build verification

Local `adafruit_feather_esp32s3` ✅ and `huzzah8266` ✅; full matrix via CI on this PR.

### DIFF
--- a/src/components/digitalIO/hardware.cpp
+++ b/src/components/digitalIO/hardware.cpp
@@ -56,6 +56,23 @@ DigitalIOHardware::~DigitalIOHardware() {
 }
 
 /*!
+    @brief  Converts a logical pin value to the electrical level to drive.
+            The digitalio.proto `is_inverted` field marks active-low pins
+            (e.g. an LED wired to VCC): IO always speaks in logical values,
+            so inverted pins must drive the opposite electrical level.
+    @param  logical_value  The logical value from the broker's perspective.
+    @return HIGH or LOW, the electrical level to apply to the pin.
+*/
+uint8_t DigitalIOHardware::GetElectricalLevel(bool logical_value) {
+  bool level = logical_value;
+  if (_is_inverted)
+    level = !logical_value;
+  if (level)
+    return HIGH;
+  return LOW;
+}
+
+/*!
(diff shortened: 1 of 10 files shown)


tyeth-ai-assisted:
@tyeth ready for your review — addresses all 11 comments from the 2026-07-23 review round on #943 (each thread there now has a reply pointing at the specific change). Local esp32s3 + huzzah8266 builds pass; CI matrix running.

PR #969, opened 2026-08-11
feat. Adafruit IO Marquee

brentru:
## Summary

Adds support for new feature - marquee - for ThinkInk EPDs. Marquee allows a broker to push a full-frame BMP in chunks over MQTT. The WipperSnapper device reassembles the BMP, draws it, and ACKs back to the broker.

This pull request also fixes a handful of high and medium severity bugs within the deep-sleep feature, blocking it from working as-expected.

## Files to look at
The majority of the files in this PR are regenerated protocol buffer files, the key files to look at are:
1. `display/controller.*`
2. `wippersnapper.cpp` 
3. `dispDrvThinkInk.h`

## Marquee

How it works:
- New `ws_display_Canvas` (checksum / size / chunk_id / chunk_total / chunk_data) and a `WriteComplete` D2B ack. `DisplayController` files chunks by id, concatenates when the count is met, verifies against the declared size, draws, then acks.
- State tracking via`marquee_state_t { IDLE, AWAITING, STREAMING, DRAWING }` — anything but `IDLE` is blocking for the loop (prevents sleep - keeps device awake during RX or writing)
- **MQTT read path widened.** - The `processPackets(10ms)` short-reads on multi-KB payloads and desyncs the stream. When a large message is pending - it drains via `readSubscription(500ms)` inside a 2s window instead.
- **Display Adds now handled during check-in**
- **Condensed 6 ThinkInk headers → 1**
- **Sleep mode is aware of displays now**

## Deep sleep

Sleep didn't work on this branch's base. There were 4 separate bugs:

1) **Duration was 10x too long.** 
2) **Sleep never started if enabled at runtime.** 
3) **No Goodnight before sleep.** - `loopSleep()` now publishes it 
4) **ext0 wakeup was lost on a net-FSM failure.** -  ext0 pin/level/pull now persists in `RTC_DATA_ATTR` and gets re-armed.
5) `loopSleep()` was restructured - the deadline check used to sit after an early `return` and could be skipped. Now `all_controllers_complete` and `has_deadline_expired` are computed up front and there's one teardown path.


src/components/display/hardware.cpp
```diff
@@ -350,110 +349,99 @@ bool DisplayHardware::beginSpiEpd(ws_display_Add *msg) {
...
-
   _drvDisp->setWidth(config->properties.width);
   _drvDisp->setHeight(config->properties.height);
+  _drvDisp->setRotation(config->properties.rotation);
   if (config->properties.text_size > 0)
     _drvDisp->setTextSize(config->properties.text_size);
 
-  if (!_drvDisp->begin(epd_mode)) {
+    // If this is not a cold boot, and the device is waking from sleep, tell the
+    // driver to skip the panel refresh since it already holds an image.
```
tyeth: two bools for this probably instead of an ever growing list of arch ifdefs

PR #971, opened 2026-08-19
add(component): Add SCD43 and split SCD40 / SCD41

tyeth:
Adds SCD43 and SCD41 as separate components to v1, matching https://github.com/adafruit/Wippersnapper_Components/pull/341

> Adds the SCD43.
> Adds descriptions to the existing shared SCD40\41 component, to specify the differing measurement ranges, splitting them out and leaving separate SCD40 SCD41 and SCD43 components.
> 
> The publish is false for the 43 and 41, but 41 is already supported in the arduino firmware so can be published true(~ = only true for v2).
> 
> Older user components were originally labelled SCD40/41 and will now be SCD40.


brentru (approved):

brentru (approved):
LGTM

PR #972, opened 2026-08-20
Telemetry with linking v2 fix

tyeth:
Fixes the issue of telemetry report-once metrics not re-reporting on re-adding.

Fixes the linking issue for v2 where the Ws global is pointing at the base class, causing failure to access the wifi adapter specific superclass methods.


tyeth:
fixed in https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/974

PR #973, opened 2026-08-20
fix(linking): Update global Ws to be the platform specific version (matches #829 which was v1)

tyeth:
Fixes the linking issue for v2 where the Ws global is pointing at the base class, causing failure to access the wifi adapter specific superclass methods.

Matches v1 #829 and fixes the RSSI returning 0 and logging network adapter not defined (https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/946)

### DIFF
--- a/src/components/analogIn/controller.h
+++ b/src/components/analogIn/controller.h
@@ -54,5 +54,4 @@ private:
   float _mcu_ref_voltage; ///< MCU/board reference voltage, used for native pins
                           ///< (expander pins use their own message ref_voltage)
 };
-extern wippersnapper Ws; ///< Wippersnapper V2 instance
 #endif                   // WS_ANALOGIN_CONTROLLER_H
--- a/src/components/analogIn/controller.h
+++ b/src/components/analogIn/controller.h
@@ -54,4 +54,4 @@ private:
   float _mcu_ref_voltage; ///< MCU/board reference voltage, used for native pins
                           ///< (expander pins use their own message ref_voltage)
 };
-#endif                   // WS_ANALOGIN_CONTROLLER_H
+#endif // WS_ANALOGIN_CONTROLLER_H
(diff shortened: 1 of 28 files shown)


tyeth:
@brentru can you have a look at this one please, it's the WS global issue where calling rssi etc from other controllers returned Network interface not defined and an RSSI of 0 (base class hit instead of platform specific super-class).

There are a few related issues, but this breaks it out into a clear single PR against v2 so other PRs can benefit/rebase.

brentru:
> Fixes the linking issue for v2 where the Ws global is pointing at the base class, causing failure to access the wifi adapter specific superclass methods.

@tyeth In the future, could the PR body be more explanatory? 

As I understand, this PR fixes the wifi adapter access by:
* Removing `extern wippersnapper` forward decls. from all header files
* Adding specific forward decls. to each network interface
* Redefining the global as `ws_platforms` which imports the network interface decl (`esp32_wifi`, etc..)

1) Is my understanding here correct? 
2) To avoid a circular dependency - could you please check that all the headers files within `ws_platforms.h` do **not** include `wippersnapper.h`? Let me know and then we can merge, I don't see any warnings in the CI compile log..

tyeth:
> > Fixes the linking issue for v2 where the Ws global is pointing at the base class, causing failure to access the wifi adapter specific superclass methods.
> 
> @tyeth In the future, could the PR body be more explanatory?
> 
> As I understand, this PR fixes the wifi adapter access by:
> 
> * Removing `extern wippersnapper` forward decls. from all header files
> * Adding specific forward decls. to each network interface
> * Redefining the global as `ws_platforms` which imports the network interface decl (`esp32_wifi`, etc..)
> 
> 1. Is my understanding here correct?
> 2. To avoid a circular dependency - could you please check that all the headers files within `ws_platforms.h` do **not** include `wippersnapper.h`? Let me know and then we can merge, I don't see any warnings in the CI compile log..

@brentru I've realised my plan was less than ideal, so have instead gone with the more correct coverting Ws to a pointer allowing the polymorphism to work correctly. See replacement PR #974

tyeth:
Closing in favour of #974

PR #974, opened 2026-08-20
fix(ws): Update Ws to pointer allowing polymorphism

tyeth:
Instead of #973 it should instead be the case that the Ws global is a pointer, allowing the polymorphism between wifi/offline adapter classes.

It's a much more exhaustive change, but minor really (externs and Ws become pointers, Ws. becomes Ws-> and a lot of formatting).

* Wippersnapper.cpp now defines the pointer Ws instead of concrete object.
* Adapter classes now set Ws pointer inside contructors
* All uses of ws updated to pointers and reformatted

The work is not really worth reviewing except the wippersnapper.cpp first change (pointer def) and the wippersnapper.h end of file. Apart from that it's find and replace, and then formatting to pass CI.


tyeth:
@brentru Having re-reviewed #973 after your comments, and dug out the v1 related comments, I see it's better to convert Ws to a pointer and correctly support the polymorphism. I've ditched #973 in favour of this PR, which is ready for review.

src/nanopb/pb_decode.c
```diff
@@ -571,7 +571,7 @@ static bool checkreturn allocate_field(pb_istream_t *stream, void *pData, size_t
     }
 #endif
 
-    /* Check for multiplication overflows.
+    /* Check for multiplication overfloWs->
```
brentru: Weird comment prob bc of the Ws-> change

src/components/analogIn/controller.cpp
```diff
@@ -153,7 +153,7 @@ bool AnalogInController::Handle_AnalogInAdd(ws_analogin_Add *msg) {
   ExpanderHardware *expander_drv = nullptr;
   if (strncmp(msg->pin_name, "EXP_", 4) == 0) {
     uint8_t i2c_addr = (uint8_t)strtoul(msg->pin_name + 4, nullptr, 16);
-    expander_drv = Ws._expander_controller->GetDriver(i2c_addr);
+    expander_drv = Ws->_expander_controller->GetDriver(i2c_addr);
```
brentru, line 156: The overall refactor is OK but we'll have to watch out now that `Ws` is a global ptr, so it's unassigned when the code starts.

As a result, calls like `Ws->` have no `nullptr` guard. 

Some things:
1) We need to ensure multiple `Ws` objects can't exist.
2) We need to ensure `Ws` is **created and initialized** before it is used by **anything** to avoid a null deref. This includes the `platforms/`,  `provisioning/`, and `wippersnapper.cpp` app.

src/wippersnapper.cpp
```diff
@@ -231,10 +231,10 @@ void wippersnapper::provision() {
   _littleFSV2->GetSDCSPin();
 #elif defined(OFFLINE_MODE_WOKWI)
   WS_DEBUG_PRINTLN("Wokwi offline mode detected, setting SD CS pin to 15");
-  Ws.pin_sd_cs = 15;
+  Ws->pin_sd_cs = 15;
```
brentru, line 282: Like, here, we're assuming `Ws` is a valid pointer **before** we provision

src/wippersnapper.cpp
```diff
@@ -270,24 +270,24 @@ void wippersnapper::provision() {
 bool handleCheckinResponse(pb_istream_t *stream) {
   // Decode the Checkin Response message
   WS_DEBUG_PRINTLN("[app] Decoding Checkin Response message");
-  if (!Ws.CheckInModel->ProcessResponse(stream)) {
+  if (!Ws->CheckInModel->ProcessResponse(stream)) {
```
brentru, line 321: Check global functions like this one and a few others for potential dereferences.

src/platforms/esp32_wifi.h
```diff
@@ -39,6 +39,7 @@ class esp32_wifi : public wippersnapper {
   @brief  Initializes the Adafruit IO class for ESP32 devices.
   */
   esp32_wifi() : wippersnapper() {
+    Ws = this;
```
brentru: Yeah the assignment in each of these is what I'm concerned most about. I'd like the code to guarantee this happens only in one location, this ctor is fine for each of the `platforms/`, but it should be dedicated, guarantee initialization, and happen before anything else.

src/wippersnapper.cpp
```diff
@@ -33,7 +33,7 @@
 
 #include "wippersnapper.h"
 
-wippersnapper Ws;
+wippersnapper *Ws;
```
brentru, line 37: Maybe something like `static inline returnType requireWsGlobal()` and then assert `Ws != nullptr` within it.

brentru (changes requested):
@tyeth  Overall I think this pattern is OK but the possibility of not guaranteeing an object and then potentially dereferencing it makes me iffy. I think it's good but I would like you to implement a few safety guarantees.

src/wippersnapper.cpp
tyeth: This will be a concrete instance after the constructor which is before provision() [this method].

src/components/analogIn/controller.cpp
tyeth: The base constructor is the only issue where a null ws could exist, but I prefer that than silently failing in the base class methods unwittingly.

Moving the ws=this to the base ctor makes some sense if the base ctor did very little else, but currently the base wippersnapper constructor unnecessarily does a lot (could be a first step of provision instead).
Having the ws=this in the base ctor and leaving the controller newing up there etc will mean we still run the risk of controllers unknowingly interacting with the base methods instead of platform specific ones.

src/nanopb/pb_decode.c
tyeth: Good spot, upset I missed those (ff.h also got one replacement for windows.h include 😂 )

tyeth:
okay, I've patched the disconnect to the superclasses, because calling disconnect from the base dtor happens after the platform dtor has finished (no disconnect possible) so replaced with a warning comment. 
Pushed it along with the fixed esp32 no-fs ctor so the PR is complete for the current/original idea (but wrong). 

I'll next refactor to the base class as suggested

tyeth:
@brentru  Okay, happy with this now, hopefully you will be too 😄 
Can I suggest you look up to the one commit change since your review, and see the complete older picture (so the comment above makes sense at a point in time). Then I've cleaned up the implementation (review whole PR which should be reduced mostly to the last two commits only).

* wippersnapper::_init() — new private method in wippersnapper.cpp holding everything that was in the base ctor body (WDT, sensor model [do we use this???], all controllers), with Ws = this; as the first action and an idempotence guard (if (Ws == this) return;). 
 Since it's called from provision() — the first thing every sketch does — the object is fully constructed with the platform vtable in place before Ws is ever published or any controller exists. No ctor-window issues in either direction.

* The base ctor is now just the member-init list (everything nulled).
* `Ws = this;` removed from all platform ctors

> On nullptr checks: no, you don't need any. The invariant is now structural rather than guarded: Ws is zero-initialized before static init, nothing in the codebase touches it until provision() runs (debug macros go straight to Serial, no globals dereference it at static-init time — I verified both), and every example calls provision() before connect()/run().
 The only "check" that exists is the Ws == this early-return in _init(), and that's re-init protection (double provision() won't leak controllers), not a null guard. Multiple instances also stop being a hazard under this design: even if someone declared two adapters, whichever one you provision() owns Ws — deterministic, no crash. And re-newing can't happen since the adapter is a never-destroyed global.

While building platformio for SAMD, I ran into issues of <SPI.h> not being included for tinyusb (fine for arduino but not pio ldf), possibly worth PRing (appreciate your thoughts as I know Limor would probably tolerate the PR, but maybe say unnecessary as little love for platformio and the framework works in arduino-cli): 
The reason is our requirement in lib_deps pulled in a copy later than bsp version so is used (and doesn't depend on spi directly), and that copy from the pio registry didn't have library.json (our repo does) which means the ldf doesn't include framework/bsp searching for it.
 Pulling the git copy of tinyusb instead of by name in libdeps means we get library.json and spi.h gets found (first whack-a-mole resolved). 
 Next servo.cpp (and other bits) in the BSP get a tinyusb import that fails to find adafruit_tinyusb.h (under PlatformIO framework-bundled libraries never receive libdeps include paths, so the unguarded include hard-fails every SAMD build there).
I proposed a fix which works to update the adafruit-samd framework to has_include guard the includes like so (which works):
(libraries/SPI/SPI.cpp, libraries/Wire/Wire.cpp, libraries/I2S/src/I2S.cpp, libraries/SAMD_AnalogCorrection/src/SAMD_AnalogCorrection.cpp, libraries/Servo/src/samd/Servo.cpp):

#ifdef USE_TINYUSB
// For Serial when selecting TinyUSB (also causes the Arduino builder to link
// the TinyUSB library)
#if !defined(__has_include) || __has_include(<Adafruit_TinyUSB.h>)
#include <Adafruit_TinyUSB.h>
#endif
#endif

brentru:
> wippersnapper::_init() — new private me

@tyeth Yeah this works, just ensure user code **always calls provision() first**, in all examples. I still would like the other Ws-> ptr review comments addressed too.

tyeth:
@tyeth double check the debug example calling provision before serial works okay (immaterial to this PR)

brentru:
@tyeth Watching this but unsure of status, tag me for a re-review whenever you're ready here

tyeth:
> @tyeth Watching this but unsure of status, tag me for a re-review whenever you're ready here

@brentru yeah rereview please, unable to see the option even after clicking resolve on things

brentru (approved):

Issue #830, opened 2025-10-15
Component fails to confirm - Linking Issue for Feather ESP32V2 causing "ERROR: Please define a network interface!"


tyeth:
Found during display testing of the E-ink featherwing (a magtag screen with 3buttons).
Component didn't confirm after add, due to the network issue.
```
[display] Successfully initialized epd display!␍␊
[display] Display added or replaced successfully!␍␊
ERROR: Please define a network interface!␍␊
Establishing network connection...␍␊
Performing a WiFi scan for SSID...ERROR: Please define a network interface!␍␊
ERROR [WDT RESET IN 25]: ␍␊
ERROR: Unable to find WiFi network, rebooting soon...␍␊
```

Digging a bit deeper it looks like we end up with two different versions of the WS global. This might be down to an older version of gcc for the esp32 versus esp32s3/s2 etc where the link-time-optimisiation is better and correctly hooks the Wippersnapper_WiFi overloads of the Wippersnapper base class.

Switching the global to be of type Wippersnapper_WiFi resolves the linking issue, but does require a small refactor. The user should see no difference, and we just have a minor tweak to each cpp file using WS.

Associated draft PR for testing - https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/pull/829

These changes solve the issue for display confirm and runNetFSM from displays/controller.cpp 
https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/compare/7857c9d07cc0e2f022731d8d2003268d9cb5c53a...3d601a92aeea619ca4a0fd98696060bc5de9ae81

PR #978, opened 2026-09-02
fix(ws): Ws global to pointer for polymorphism (PR #974 extraction) + ESP flash watchdog-reset

tyeth:
## Summary

Extracts the **Ws global-to-pointer polymorphism fix** from adafruit/Adafruit_Wippersnapper_Arduino#974 onto `migrate-api-v2`, so it can land in v2 independently of the telemetry work.

**Why an extraction instead of retargeting #974**: #974 merged into the telemetry branch (`claude/telemetry-protobuf-components-71d274`), which embeds a stale snapshot of the driver-backport branch (up to 8035f4bb / PR #935 era). Retargeting would have dragged those stale driver copies into `migrate-api-v2`. This branch cherry-picks only the #974 commits (original authorship preserved) with the telemetry hooks stripped out.

## What's included

- `fix(ws): Update Ws to pointer allowing polymorphism` — `extern wippersnapper *Ws` everywhere; fixes the two-globals problem (#830) where the base-class global shadowed the WiFi adapter, e.g. `getRSSI()` returning 0 (#829 family)
- `fix(base): move disconnect to superclasses, fix esp32 secrets ctor` — secrets ctor writes its own members, not through the (not-yet-set) global
- `refactor(ws): move ws ptr and ctor items to _init() and guard re-entry` — `Ws = this` happens in `provision()` once the vtable is in place
- `fix(platformio): remove duplicate lib defs`
- `fix(samd): Require library.json (via git) for Adafruit TinyUSB` — registry copy lacks library.json so the LDF misses TinyUSB's SPI dependency on SAMD
- `fix(gitignore): ignore repos folder`
- `fix(fs): keep fatfs ff.h ahead of diskio.h under clang-format` — the include sorter had reordered them, breaking rp2040/samd (diskio.h consumes BYTE/DWORD from ff.h); now guarded with clang-format off/on
- `feat(debug): print broker URL + port when not on io.adafruit.com` — v1 `printDeviceInfo()` parity for non-production brokers
- `fix(upload): reset ESPs via watchdog after flashing` — `board_upload.after_reset = watchdog-reset` under `[common:esp32]`; RTS hard-reset doesn't reliably reboot native-USB parts (S2/S3 sit dead until replugged)

Upstream's clang-format commit (b7a0c4a4) was skipped (it conflicted with post-pins-as-strings drift and resurrected telemetry files); a fresh clang-format v20 pass replaces it. Adds that only touched telemetry files were dropped — they stay with #946.

## Verification

- Local builds green: `featheresp32`, `raspberrypi_picow`, `adafruit_feather_esp32s2_reversetft`
- **Hardware-verified on a Feather ESP32-S2 Reverse TFT**: flashed 2.0.0-alpha.1, WIPPER boot file confirms version, connects to a staging broker (io.adafruit.us:8883 — exercising the new URL/port print), completes checkin, steady MQTT pings with live RSSI values (the pointer fix in action vs the old stub 0)
- `watchdog-reset` verified: board self-reboots after flashing with no replug (note: esptool reports a cosmetic port error after the watchdog fires, as the USB port vanishes before it can close cleanly)

## Relation to other PRs

- #946 (telemetry) shrinks to telemetry-only once this + #933 land and `migrate-api-v2` is merged back into its branch
- #933 (driver backport) is independent; both target `migrate-api-v2`


tyeth:
This incorporates the telemetry and polymorphism PRs and the latest protos, so migrate-api-v2 will be up to date.
Previously the PRs were approved, but stacked on the migrate-api-v2-backport-components branch+PR.

Self-rereviewed and diffed. Merging in after green CI, so in theory the firmware is up to date, and the v2 sensors PR (migrate-api-v2-backport-components) will reduce to just expected work.

PR #982, opened 2026-09-03
feat(fruitjam): port Adafruit Fruit Jam support to api-v2

tyeth:
Ports Adafruit Fruit Jam (RP2350 + onboard ESP32-C6 AirLift) support from `main` to the api-v2 codebase. Stacked on adafruit/Adafruit_Wippersnapper_Arduino#868 (uses its define-based adapter selection).

- `platformio.ini`: `adafruit_fruitjam` + `adafruit_fruitjam_debug` envs (8MB FS, WiFiNINA enabled via lib_ignore override)
- `ws_boards.h`: FruitJam board block (BOARD_ID `fruitjam`, TinyUSB, 5x NeoPixel status)
- `ws_platforms.h` + `src/Wippersnapper_demo.ino`: route `ARDUINO_ADAFRUIT_FRUITJAM_RP2350` to the AirLift WiFi adapter
- `airlift_wifi.h`: re-port of `fa9aefff` (Fruit Jam C6 guarding, dropped in the v2 rewrite) — FruitJam requires nina-fw 3.3.0+ (others keep 2.0.0-rc.0+adafruit), `ESP32_RESETN` reset pin, clearer version-mismatch messaging. Version check fails gracefully: an unreadable/incompatible version logs a warning and continues, and a null/empty version string no longer reaches `sscanf`
- Bootlog now reports the AirLift co-processor firmware version again (`_airlift_version`, captured in `getMacAddr()` during provisioning so it's valid before `wipper_boot_out.txt` is written; falls back to "unknown")

Build-verified locally: `adafruit_fruitjam` SUCCESS (RAM 5.6%, flash 6.1%). Note: `adafruit_pyportal_m4_titano` fails to link with FLASH overflowed by ~13.5KB **on the base branch already** (13452 bytes without this PR, 13612 with) — pre-existing on the 868+migrate-api-v2 merge, not introduced here.


tyeth-ai-assisted:
**Hardware verification on a real Fruit Jam** (RP2350B QFN80, flashed `b95b342c` via UF2):

`wipper_boot_out.txt`:
```
Adafruit.io WipperSnapper
Firmware Version: 2.0.0-alpha.1
Board ID: fruitjam
MAC Address: 98:88:E0:62:58:48
AirLift Coprocessor Firmware Version: 3.2.0
```

Serial log highlights:
```
ESP32 booted, version: 3.2.0
WARNING: Unable to read nina-fw version or incompatible version detected!
Required nina-fw version: 3.3.0
Connecting to <ssid>
Connected to WiFi!
...
Completed checkin process!
Running app loop...
```

Confirms:
- Board identifies as `fruitjam`, WIPPER drive + `secrets.json` provisioning works
- AirLift co-processor version is captured before the bootlog is written (real value, not "unknown")
- This board's C6 reports nina-fw 3.2.0, below the 3.3.0 minimum inherited from v1 — the check warns and gracefully continues as intended, and the device connects to IO, completes checkin, and runs the app loop. Min version left at 3.3.0 deliberately.

tyeth:
@brentru this is ready for review. When testing the wifi/offline pico PR ready to merge into v2 I noticed the lack of fruitjam definition. This brings v2 inline with v1 for fruitjam and airlift functionality.

src/platforms/airlift_wifi.h
```diff
@@ -168,6 +178,9 @@ class airlift_wifi : public wippersnapper {
   */
   bool firmwareCheck() {
     _fv = WiFi.firmwareVersion();
+    if (_fv == nullptr || _fv[0] == '\0')
+      return false; // co-processor unresponsive or returned no version
+    Ws->_airlift_version = _fv;
```
brentru, line 183: `_fv` is local to this class, but it isnt written to the boot output.

While "unknown" will be unconditionally written - you're fetching `_fv` here. You may want to store `_fv` prior to the check with early return.

brentru (changes requested):
@tyeth  Looks good - small question (maybe I'm wrong, dunno)

src/platforms/airlift_wifi.h
tyeth: Think it's alright.

> reasoning, per the comment: he believes that when firmwareCheck() early-returns, the fetched version never gets stored, so the bootlog would unconditionally print "unknown" even though _fv was actually fetched right there.
> 
> But that rests on a misread of where the bootlog gets its value. The flow is:
> 
> provision() → getMacAddr() — which also captures the version into Ws->_airlift_version (the second hunk of the same commit, in getMacAddr()), then
> provision() → FS init → bootlog written — with the value from step 1, then
> connect() → firmwareCheck() — whose store at line 183 is just a refresh, and irrelevant to the bootlog, which was already written.
> 
> So the bootlog never depends on firmwareCheck() at all — and this is hardware-proven: the Fruit Jam on your desk printed AirLift Coprocessor Firmware Version: 3.2.0 in wipper_boot_out.txt, a value that could only have come from the getMacAddr() capture since firmwareCheck() hadn't run yet.
> 
> And doing what he suggests would be a small regression: the early return only fires when _fv is null or empty, so "storing prior to the check" means clobbering either the "unknown" default or a previously captured good value with an empty string (blank bootlog line) — or storing a nullptr that bootFile.println()/debug prints would then dereference. Only-store-validated-values is deliberate; the two capture sites (early for the bootlog, refresh at connect) are the mechanism that makes the bootlog correct.
>

PR #990, opened 2026-09-10
Fix GH:Actions deprecations

lorennorman:
Bump all deprecated GH:Actions to their latest versions.

All were official `actions/*` actions with blue checks on the GH:A Marketplace, so I did not include the commit SHA for them.


lorennorman:
After the Actions run, all of the deprecations were cleared, however there are two arduino-related warnings:
image

@brentru or @tyeth can you take a look and make sure those are not important and/or fix them?
https://github.com/adafruit/Adafruit_Wippersnapper_Arduino/actions/runs/34501287452

tyeth (approved):
LGTM. Not sure we can quickly fix the arduino-library tag missing warning, but I'll put it on my list (it's the adabot weekly check script saying the same thing for last 3yrs, harmless but I forget the reason 🙃 )

PR #992, opened 2026-09-16
refactor(i2c): consolidate v1 read semantics into drvBase hooks + fix cadence bugs

tyeth:
Stacks on adafruit/Adafruit_Wippersnapper_Arduino#933 (`migrate-api-v2-backport-components`). Follow-up to the base/controller refactor already on that branch, from a fresh v1-vs-v2 review of sensor polling, warm-ups, retries and caching.

### Bugs fixed
- **Continuous mode published each I2C device once per boot.** `update()` bailed on `GetDidReadSend()` unconditionally, and the flag is only reset between sleep cycles. Now gated on `force` like the UART controller. The GPS controller had the identical bail and gets the same fix.
- **Stale sample republish / frozen sensor.** The old template returned the cached sample whenever the data-ready check failed, so a period shorter than the sensor interval produced duplicates and an unplugged device published its last value forever. A sample is now served only until the controller consumes it via `SampleDone()`; after that a fresh measurement is required.
- **`fastTick()` ran before the MUX channel was selected**, so an SGP behind a mux sampled whatever channel was current. Tick cadence now lives in the base, so the controller selects the mux only when a tick is actually due.
- **Forced updates (sleep mode) bypassed the backoff.** `ReadDue(now, force)` folds the period check and a retry timestamp that is honoured regardless of `force`, never faster than 1 Hz.
- SGP30/40/41 published algorithm output before the first tick (SGP41: zeros fed to the NOx algorithm during conditioning). Nothing publishes until the first successful measurement; SGP41 algorithms are not fed during conditioning, per the Sensirion reference usage.
- **Blocking reads removed from the read pass:** VL53L4CD (250 ms + up to 3x300 ms), VL53L4CX (250 ms), MLX90632 step mode (up to 6 s), DS2484/DS18B20 (750 ms), NAU7802 (up to 250 ms), LTR390 (2x110 ms mode switches) and MCP3421 (up to 500 ms) now measure in the background via `fastTick()`. No driver blocks the read pass for longer than a single short conversion.
- MLX90632 step-mode timeout truncated `millis()` to `int16_t`; HDC302X discard read now runs after auto-mode exit; D6T1A checks the `read()` status.

### Base API (`drvBase`)
Named by who calls them.

Driver hooks (driver implements, base calls), v1 names:
- `IsSensorReady()` - data-ready gate, default true.
- `ReadSensorData()` - the one I2C transaction into the driver's cache, default no-op.
- `fastTick()` - background step for devices that cannot produce a sample inside one read pass. 
Opt in with `_fast_tick_ms`. 
By default ticks run only in the `_tick_lead_ms` window before the next read is due (so a VL53 with a 5 minute period ranges for ~1 s per period); `_tick_lead_ms = TICK_ALWAYS` for algorithms that must run continuously (SGP gas index).
The tick files its result with `NewSample()`, which is the only sample source for a ticking driver. `_first_tick` is true on the first tick of a lead window so multi-step measurements restart rather than trust a result that completed while idle.

Controller calls (base implements):
- `ReadDue(now, force)` - period elapsed (or forced) and not in retry backoff.
- `AttemptRead()` - serves the unconsumed sample, else runs the two hooks; may or may not touch the bus. The getEvent accessors call it too and get the same-pass sample.
- `SampleDone(published)` - consumes the sample either way; success resets the failure counter, failure schedules the retry (3 quick retries ~1 s apart, then a full period) and returns the delay.
- `FastTickDue(now, force)` - tick cadence and lead-window check.

Warm-up: `_discard_samples`, set in a driver's constructor, makes `NewSample()` drop that many samples before the first is published, so known-garbage first conversions are handled in one place (DS18B20 1, BME680 1, HDC302X 1 replacing its blocking discard read, SGP30 15 for the IAQ init period, SGP40/41 45 for the gas-index blackout).

### Drivers converted
Onto `IsSensorReady()` / `ReadSensorData()`: SCD30, SCD4x, SEN5x, SEN6x (reference shape), STCC4, DPS310, HDC302X, D6T1A, BME680, BMP3xx, BMP5xx, PM25, ENS160, SHT4x, MS8607. One measurement per read pass instead of one per metric (BME680 was running five heater cycles per pass, MS8607 three conversions). Altitude is derived from the cached pressure with the libraries' own formula, since their `readAltitude()` helpers trigger another read.

Onto `fastTick()`: SGP30, SGP40, SGP41 (`TICK_ALWAYS`; SGP41 ported from v1 with serial, self-test, learning-window notes; SGP40 runs the Sensirion VOC algorithm on one raw read per tick instead of two heater cycles), VL53L4CD, VL53L4CX (50 ms polls in a 1 s lead window), MLX90632 (lead derived from the configured mode and refresh rate), DS2484/DS18B20 (start conversion, read scratchpad 750 ms later; every step retried next tick, scratchpad CRC-validated, and after 3 consecutive failures the bridge is reset and the bus re-searched so a removed, rewired or replaced sensor recovers; a missing sensor files no sample instead of NAN), NAU7802 (harvest at 10 SPS), LTR390 (ALS then UVS cycle with 110 ms settles), MCP3421 (one-shot start, read on `isReady()`).

Also adds an `adafruit_feather_esp32s3_debug` PlatformIO env.

### Datasheet accuracy sweep
Every I2C driver was reviewed against its datasheet for one rule: publish only data that is meaningfully accurate. Three commits apply the in-driver fixes (gas / T-RH-P / light-power groups): warm-up and blackout windows (SGP30 15 s, SGP40/41 46-sample blackout + conditioning cap, STCC4 20 s bypass, ENS160 validity flag, SEN5x/6x fan and gas-index windows plus device-status faults, PM25 30 s fan spin-up), sentinel and saturation rejection (SEN66 0xFFFF/0x7FFF unscaled by the library, BME680 gas 0, VEML7700/VCNL/LTR/TSL2591 full-scale, ToF range status, INA MATHOF, QMC5883P overflow), failed reads no longer published as NAN/0/garbage (HTU21D, MCP9808, Si7021, MPRLS, LC709203F, INA219, MPL115A2, MS8607, LPS2x, STEMMA soil, MAX1704x, VL53/VL6180X), modes that stop measuring (BMP280 forced/sleep, BMP5xx forced/standby, TMP117/119 one-shot/shutdown) re-triggered or suppressed, data-ready gates where the silicon has one (APDS9999, AS7331, BMP5xx, INA2xx/INA260, LPS28DFW, SPA06 temp, TMP117/119, VCNL4020), unit fixes (LTR390, VCNL4020/4200 published raw counts as lux), and out-of-spec defaults (DPS310 64 Hz x 64x). Items needing library changes, product decisions or bench checks are tracked in adafruit/Adafruit_Wippersnapper_Arduino#995.

### Verification
Not built locally (pioarduino `stable` needed a newer PlatformIO Core than the local penv had at the time). `build-migrate-api-v2.yml` dispatched manually on each commit since the PR base is not `migrate-api-v2`; full matrix green on every commit so far (two ESP32 jobs needed a rerun after a BSP download HTTP 500), plus the PR clang-format/doxygen check.


brentru:
@tyeth  - Hi, is this ready for review?

src/components/i2c/drivers/drvApds9999.h
```diff
@@ -90,10 +104,14 @@ class drvApds9999 : public drvBase {
...
   bool getEventLight(sensors_event_t *lightEvent) {
+    if (!AttemptRead() || !_ls_new)
+      return false;
     uint32_t r, g, b, ir;
     if (!_apds9999->getRGBIRData(&r, &g, &b, &ir))
       return false;
-
+    // Full-scale count for the configured resolution means saturation
+    if (g >= ((1UL << _ls_bits) - 1))
+      return false;
```
tyeth: debateably this could return NaN / INF instead of false, or another candidate for a status / data quality feed (saturated would be the current status at potential publish of this value)

src/components/i2c/drivers/drvAs7331.h
```diff
@@ -302,27 +302,35 @@ class drvAs7331 : public drvBase {
...
+                OUTCONVOF) are set: the gain is too high for the light level.
       @param    rawEvent
                 Pointer to an Adafruit_Sensor event.
       @returns  True if the sensor event was obtained successfully, False
                 otherwise.
   */
-  /*******************************************************************************/
   bool getEventRaw(sensors_event_t *rawEvent) {
+    if (!AttemptRead() || _as7331->hasOverflow())
+      return false;
```
tyeth: overflow should somehow make it back to user / error proto, or again status/data-quality feed or flag.

src/components/i2c/drivers/drvBme680.h
```diff
@@ -249,23 +255,27 @@ class drvBme680 : public drvBase {
...
   */
   bool getEventAltitude(sensors_event_t *altitudeEvent) {
-    if (!bmePerformReading())
+    if (!AttemptRead())
       return false;
-    altitudeEvent->altitude = (float)_bme->readAltitude(_seaLevelPressureHpa);
+    // Same formula as Adafruit_BME680::readAltitude(), without the re-read
+    float atmospheric = (float)_bme->pressure / 100.0F;
+    altitudeEvent->altitude =
+        44330.0F * (1.0F - pow(atmospheric / _seaLevelPressureHpa, 0.1903F));
```
tyeth: Ideally no magic numbers, use defines with appropriate names to imply algorithm or variable meaning

src/components/i2c/drivers/drvBmp3xx.h
```diff
@@ -272,23 +306,27 @@ class drvBmp3xx : public drvBase {
...
   */
   bool getEventAltitude(sensors_event_t *altitudeEvent) {
-    if (!_bmp3xx->performReading())
+    if (!AttemptRead())
       return false;
-    altitudeEvent->altitude = _bmp3xx->readAltitude(_seaLevelPressureHpa);
+    // Same formula as Adafruit_BMP3XX::readAltitude(), without the re-read
+    float atmospheric = _bmp3xx->pressure / 100.0F;
+    altitudeEvent->altitude =
+        44330.0F * (1.0F - pow(atmospheric / _seaLevelPressureHpa, 0.1903F));
```
tyeth: Same "magic numbers / algorithm in words" fix suitable here

src/components/i2c/drivers/drvDps310.h
```diff
@@ -64,8 +63,11 @@ class drvDps310 : public drvBase {
     }
 
     // init OK, perform sensor configuration
-    _dps310->configureTemperature(DPS310_64HZ, DPS310_64SAMPLES);
-    _dps310->configurePressure(DPS310_64HZ, DPS310_64SAMPLES);
+    // Datasheet 8.3: rate_T*t_T + rate_P*t_P must stay under 1 second; at
+    // 64x oversampling (104.4ms) that allows 4 Hz for each channel (835ms).
+    // The library's 64 Hz default is out of spec and its timing undefined.
```
tyeth: Confirm out of spec on HITL farm before altering default sample rate, this is totally unacceptable as ideally v1 == v2 default sensor settings.

src/components/i2c/drivers/drvDs2484.h
```diff
@@ -63,87 +78,178 @@ class drvDs2484 : public drvBase {
...
+      return false;
     _ds2484->OneWireSearchReset();
-    while (!found_device && _ds2484->OneWireSearch(_rom)) {
-      if (_rom[0] == DS18B20_FAMILY_CODE) {
-        found_device = true;
+    uint8_t rom[8];
+    while (_ds2484->OneWireSearch(rom)) {
+      if (rom[0] == DS18B20_FAMILY_CODE) {
+        memcpy(_rom, rom, sizeof(_rom));
+        return true;
```
tyeth: Could return error.proto if ds18b20 serial no longer matched and/or bus failure was detected.

src/components/i2c/drivers/drvEns160.h
```diff
@@ -107,13 +107,39 @@ class drvEns160 : public drvBase {
...
+    _i2c->write((uint8_t)ENS160_REG_DATA_STATUS);
+    if (_i2c->endTransmission(false) != 0)
+      return false;
+    if (_i2c->requestFrom((uint8_t)_address, (uint8_t)1) != 1)
+      return false;
+    uint8_t status = _i2c->read();
+    if (status & 0x40) // STATER: device error
+      return false;
+    if (((status >> 2) & 0x03) != 0) // VALIDITY: warm-up / start-up / invalid
+      return false;
```
tyeth: Ideally we'd feed back to error.proto that the sensor was in warm-up for the next hour etc.

src/components/i2c/drivers/drvHdc302x.h
```diff
@@ -150,10 +148,12 @@ class drvHdc302x : public drvBase {
...
                                                    TRIGGERMODE_LP0)) {
       WS_DEBUG_PRINTLN("Failed to read temperature and humidity.");
-      return _have_data;
+      return false;
     }
-    _last_read = millis();
-    _have_data = true;
+    // Raw 0x0000 words (cleared result registers) decode to exactly -45C /
+    // 0 %RH with a valid CRC: not a measurement
+    if (_temp <= -44.99 && _humidity <= 0.01)
```
tyeth: could probably be an or || as garbage data if humidity < 0.01% or temp < -45c

src/components/i2c/drivers/drvIna228.cpp
```diff
@@ -298,6 +298,19 @@ bool drvIna228::getEventVoltage(sensors_event_t *voltageEvent) {
  * otherwise.
  */
 bool drvIna228::getEventCurrent(sensors_event_t *currentEvent) {
+  // DIAG_ALRT MATHOF (bit 9): current/power arithmetic overflowed - the
+  // shunt calibration does not fit the measured current
+  if (_ina228->alertFunctionFlags() & (1u << 9))
```
tyeth: Definitely raise error.proto here as user could alter the settings / fix their setup

src/components/i2c/drivers/drvIna237.cpp
```diff
@@ -298,6 +298,19 @@ bool drvIna237::getEventVoltage(sensors_event_t *voltageEvent) {
  * otherwise.
  */
 bool drvIna237::getEventCurrent(sensors_event_t *currentEvent) {
+  // DIAG_ALRT MATHOF (bit 9): current/power arithmetic overflowed - the
+  // shunt calibration does not fit the measured current
+  if (_ina237->alertFunctionFlags() & (1u << 9))
+    return false;
```
tyeth: raise errors.proto here so user can resolve

src/components/i2c/drivers/drvIna238.cpp
```diff
@@ -298,6 +298,19 @@ bool drvIna238::getEventVoltage(sensors_event_t *voltageEvent) {
  * otherwise.
  */
 bool drvIna238::getEventCurrent(sensors_event_t *currentEvent) {
+  // DIAG_ALRT MATHOF (bit 9): current/power arithmetic overflowed - the
+  // shunt calibration does not fit the measured current
+  if (_ina238->alertFunctionFlags() & (1u << 9))
+    return false;
```
tyeth: raise errors.proto as user can resolve overflow in theory.

src/components/i2c/drivers/drvLps28dfw.h
```diff
@@ -222,27 +222,21 @@ class drvLps28dfw : public drvBase {
...
-        if (_temp == NULL) {
-          _temp = _lps28->getTemperatureSensor();
-        }
-        if (_pressure == NULL) {
-          _pressure = _lps28->getPressureSensor();
-        }
-        return true;
-      }
+      if ((_lps28->getStatus() & both) == both)
+        return _lps28->getEvent(&_pressure_ev, &_temp_ev);
```
tyeth: confirm still works if user only enables one subcomponent / metric out of temp + pressure

src/components/i2c/drivers/drvLtr329_Ltr303.h
```diff
@@ -207,8 +207,15 @@ class drvLtr329_Ltr303 : public drvBase {
...
-      _has_reading = true;
+      uint16_t ch0, ch1;
+      // The library returns the ALS_STATUS "data valid" bit; a full-scale
+      // channel (65535) is saturated
+      if (_LTR329->readBothChannels(ch0, ch1) && ch0 != 0xFFFF &&
+          ch1 != 0xFFFF) {
+        _visible_plus_ir = ch0;
+        _infrared = ch1;
+        _has_reading = true;
+      }
```
tyeth: if one channel is saturated, do we want to still use the other value, and do we wish to bubble up error.proto, so user or we could attempt gain / integration time adjustment.

src/components/i2c/drivers/drvLtr390.h
```diff
@@ -65,48 +74,107 @@ class drvLtr390 : public drvBase {
...
+      @brief    Background measurement step, called every LTR390_TICK_MS while
+                a read is pending. Cycles ALS then UVS: switch mode, wait
+                LTR390_MODE_SETTLE_MS, read once new data is flagged (or give
+                up on that mode after LTR390_DATA_TIMEOUT_MS), then move on.
+                When both modes have been visited the sample is filed with
+                NewSample(); a metric whose mode produced no data is reported
+                as unavailable by its accessor.
+  */
+  void fastTick() override {
+    ulong now = millis();
```
tyeth: retest with one metric disabled (and different operating modes / etc) to confirm newsample still fires

src/components/i2c/drivers/drvMax1704x.h
tyeth: Do we need to detect battery removal or failures to detect battery and therefore sensor misreporting / errors. Does it handle recovery okay, and flat batteries ok.
Probably fine / expected behaviour but worth validating.

src/components/i2c/drivers/drvMcp3421.h
```diff
@@ -85,34 +93,56 @@ class drvMcp3421 : public drvBase {
...
+  void fastTick() override {
+    ulong now = millis();
+    if (_first_tick)
+      _converting = false; // any conversion from a previous window is stale
+
+    if (!_converting || now - _convert_start > MCP3421_CONVERT_TIMEOUT_MS) {
+      if (_converting)
+        WS_DEBUG_PRINTLN("MCP3421: conversion timed out, restarting");
+      if (!_mcp3421->startOneShotConversion()) {
+        WS_DEBUG_PRINTLN("Failed to start one-shot conversion");
```
tyeth: failure to start conversion might want to be errors.proto

src/components/i2c/drivers/drvMs8607.h
```diff
@@ -70,6 +70,24 @@ class drvMs8607 : public drvBase {
...
+    // The library's getEvent() ignores read/CRC failures and returns true, so
+    // range-check against the datasheet spans (10-2000 mbar, -40..85C,
+    // 0..100 %RH) with the pressure span tightened to atmospheric
+    if (!_ms8607->getEvent(&_pressure, &_temp, &_humidity))
+      return false;
+    return _pressure.pressure >= 300.0F && _pressure.pressure <= 1200.0F &&
+           _temp.temperature >= -40.0F && _temp.temperature <= 85.0F &&
+           _humidity.relative_humidity >= 0.0F &&
+           _humidity.relative_humidity <= 100.0F;
+  }
```
tyeth: confirm discard sample not needed

src/components/i2c/drivers/drvPm25.h
```diff
@@ -18,6 +18,10 @@
 
 #include "drvBase.h"
 #include <Adafruit_PM25AQI.h>
+
+/// Plantower manual: "stable data should be got at least 30 seconds after the
+/// sensor wakeup ... because of the fan's performance"
+#define PM25_FAN_STARTUP_MS 30000
```
tyeth: This will be unexpected by users, who will expect data immediately. Maybe we have to errors.proto msg any warming-up sensors so users understand.

src/components/i2c/drivers/drvQmc5883p.cpp
```diff
@@ -246,20 +246,40 @@ bool drvQmc5883p::getEventRaw(sensors_event_t *rawEvent) {
...
 
-  // Check for overflow
-  if (_qmc->isOverflow()) {
+  // Datasheet 9.2.2: OVFL flags any axis outside +/-30000 LSB, but reading
+  // STATUS for the data-ready check above already cleared it, so apply the
+  // same limit to the raw data
+  if (abs(x) > 30000 || abs(y) > 30000 || abs(z) > 30000) {
     WS_DEBUG_PRINTLN("QMC5883P data overflow - skipping reading");
     return false;
   }
```
tyeth: report via errors.proto?

src/components/i2c/drivers/drvSen6x.h
```diff
@@ -96,37 +99,65 @@ class drvSen6x : public drvBase {
...
+    // returns them unscaled (3276.7, 6553.5 ...), so map them to NAN here.
+    _massConcentrationPm1p0 = pm1 == 0xFFFF ? NAN : pm1 / 10.0f;
+    _massConcentrationPm2p5 = pm25 == 0xFFFF ? NAN : pm25 / 10.0f;
+    _massConcentrationPm4p0 = pm4 == 0xFFFF ? NAN : pm4 / 10.0f;
+    _massConcentrationPm10p0 = pm10 == 0xFFFF ? NAN : pm10 / 10.0f;
+    _ambientHumidity = rh == 0x7FFF ? NAN : rh / 100.0f;
+    _ambientTemperature = t == 0x7FFF ? NAN : t / 200.0f;
+    _vocIndex = voc == 0x7FFF ? NAN : voc / 10.0f;
+    _noxIndex = nox == 0x7FFF ? NAN : nox / 10.0f;
+    _co2 = co2;
```
tyeth: Brent dislikes ternary operators, but this is probably ok / most compact.

src/components/i2c/drivers/drvSgp40.h
```diff
@@ -18,15 +18,22 @@
 
 #include "drvBase.h"
 #include <Adafruit_SGP40.h>
+#include <VOCGasIndexAlgorithm.h>
 #include <Wire.h>
 
 #define SGP40_FASTTICK_INTERVAL_MS 1000 ///< Enforce ~1 Hz sampling cadence
+/// The gas-index algorithm outputs 0 while uptime <= 45s, i.e. for the first
+/// 46 one-second samples (sensirion_gas_index_algorithm.c, INITIAL_BLACKOUT)
+#define SGP40_BLACKOUT_SAMPLES 46
```
tyeth: would be nice to warn users this is going on (errors.proto Warning: "Funhouse - SGP40" needs 45mins to warm up, expect data arrival after warm up period elapses")

Less important if < 60s, but maybe still wanted.

src/components/i2c/drivers/drvSgp41.h
tyeth: conditioning / warm up errors.proto maybe

src/components/i2c/drivers/drvStemmaSoil.h
```diff
@@ -81,13 +81,11 @@ class drvStemmaSoil : public drvBase {
...
     // https://github.com/adafruit/Adafruit_Seesaw/blob/master/Adafruit_seesaw.cpp
-    if (touchData == 65535) {
-      rawEvent->data[0] = NAN;
-    } else {
-      // TODO: Update this should we add a capacitive moisture type to
-      // adafruit_sensor
-      rawEvent->data[0] = (float)touchData;
-    }
+    if (touchData == 65535)
+      return false;
```
tyeth: potentially report touch > 65k read error via errors.proto (soil sensor confusion otherwise)

src/components/i2c/drivers/drvTmp117.h
```diff
@@ -169,10 +169,41 @@ class drvTmp117 : public drvBase {
...
+    if (_mode == TMP117_MODE_SHUTDOWN)
+      return false;
+    if (_mode == TMP117_MODE_ONE_SHOT && !_one_shot_pending) {
+      _one_shot_pending = _tmp117->setMeasurementMode(TMP117_MODE_ONE_SHOT);
+      return false;
     }
-    return _tmp117->setMeasurementMode(meas_mode);
+    tmp117_alerts_t alerts;
+    return _tmp117->getAlerts(&alerts) && alerts.data_ready;
+  }
```
tyeth: what happens in continuous mode if exposed / selected via settings

src/components/i2c/drivers/drvVeml7700.h
```diff
@@ -183,9 +183,12 @@ class drvVeml7700 : public drvBase {
                 otherwise.
   */
   bool getEventLight(sensors_event_t *lightEvent) {
+    // A full-scale ALS count (65535) is saturation: the lux the library
+    // derives from it (tens of thousands) is not a measurement
+    if (_veml->readALS(false) == 0xFFFF)
+      return false;
```
tyeth: We could report (errors.proto), or re-auto-gain and repeat measurement

src/components/i2c/drivers/drvVl53l4cd.h
```diff
@@ -98,61 +105,54 @@ class drvVl53l4cd : public drvBase {
...
+
+    VL53L4CD_Result_t results = {0};
+    uint8_t status = _VL53L4CD->VL53L4CD_GetResult(&results);
+    // (Mandatory) Clear HW interrupt to restart measurements
+    _VL53L4CD->VL53L4CD_ClearInterrupt();
+    // RangeStatus = 0 means valid data; otherwise keep waiting for a good one
+    if (status != VL53L4CD_ERROR_NONE || results.range_status != 0)
+      return;
+    // NOTE: results also carries sigma_mm (std deviation) should we want it.
+    _distance_mm = results.distance_mm;
```
tyeth: Consider the extra metric of sigma_mm (std deviation)

src/components/i2c/drivers/drvVl53l4cx.h
```diff
@@ -109,61 +144,25 @@ class drvVl53l4cx : public drvBase {
   }
 
   /*!
-      @brief    Gets the VL53L4CX's current proximity (first or second object).
+      @brief    Gets the VL53L4CX's current proximity (first or second object)
```
tyeth: Can transfer this into a second proximity metric very soon (like multiple light types)

src/components/i2c/drivers/drvVl6180x.h
```diff
@@ -89,10 +89,9 @@ class drvVl6180x : public drvBase {
       } else if (status == VL6180X_ERROR_RANGEOFLOW) {
         WS_DEBUG_PRINTLN("VL6180X: Range reading overflow");
       }
-      proximityEvent->data[0] = NAN;
-    } else {
-      proximityEvent->data[0] = range;
+      return false;
```
tyeth: do we want to feed the ranging errors back/

src/components/i2c/drivers/drvVncl4040.h
tyeth: Expected custom settings for integration time and led current and resolution etc

tyeth: See VCNL4200 for similar

platformio.ini
tyeth: Looks a bit crap, fix the M4 airlift into a debug target separate from normal target.

tyeth:
@brentru it is now ready for review. I will still make minor changes to platformio.ini, and potentially the vncl4040.
Any other changes would be after we decide how to handle warming up sensors.

src/components/gps/controller.cpp
```diff
@@ -207,8 +207,10 @@ void GPSController::update(bool force) {
     return; // bail-out!
 
   for (GPSHardware *drv : _gps_drivers) {
-    // (force only) - Was driver previously read and sent?
-    if (drv->GetDidReadSend())
+    // (force only, sleep mode) - Was driver previously read and sent? The
+    // flag is only reset between sleep cycles, so it must not gate periodic
+    // reads in continuous mode.
+    if (drv->GetDidReadSend() && force)
```
brentru: Can this be brought out into a separate PR? Unrelated to i2c.

src/components/i2c/drivers/drvBase.h
```diff
@@ -44,6 +44,10 @@ static_assert(sizeof(WsPinName::name) >= DRV_BASE_PIN_NAME_LEN,
     instead of waiting out a full period. Name kept in sync with the v1 driver
```
brentru: @tyeth  There's a number of new functions here which replace existing functions like `getEvent()`.

Could you please, in a comment, document what each of these new things do (i.e: fast tick, attemptread), why they are required for migration, etc so we can look at it together?

brentru (changes requested):

